PR #8, opened 2023-07-14
Create SQL file for user table (also handles organizations)

lemilonkh:


lgloege:
@lemilonkh , this looks nice. My only suggestions would be:
- adding a `last_updated` field
- maybe renaming `id` to `user_id`. My logic here is so the field name is consistent when we use it a foreign key.

PR #11, opened 2023-07-17
ON-565

cephaschapa:


lemilonkh (approved):
Looks good. I think these schemas will need to be referenced using foreign keys by other schemas, so let's implement that after merging #10 .

cephaschapa:
Thanks!! Will do

PR #14, opened 2023-07-27
WIP: Inventory CRUD API

lemilonkh:


app/models/index.js
```diff
@@ -0,0 +1,43 @@
...
+
+let sequelize;
+if (config.use_env_variable) {
+  sequelize = new Sequelize(process.env[config.use_env_variable], config);
+} else {
+  sequelize = new Sequelize(config.database, config.username, config.password, config);
+}
+
+fs
+  .readdirSync(__dirname)
```
evanp, line 20: So, you're reading all the models in the directory?

How many models are we expecting, such that this is easier than just listing the models directly?

app/next.config.js
```diff
@@ -1,4 +1,14 @@
 /** @type {import('next').NextConfig} */
-const nextConfig = {}
+const nextConfig = {
+  serverRuntimeConfig: {
```
evanp, line 3: Great. We want our config to come from the environment vars

app/src/util/api.ts
```diff
@@ -2,10 +2,18 @@ import createHttpError from "http-errors";
...
+
 export function apiHandler(handler: (req: Request) => Promise<NextResponse>) {
   return async (req: Request) => {
     try {
+      if (!db.initialized) {
+        await db.initialize();
+      }
+
       // TODO JWT authentication logic here
+      // await jwtMiddleware(req);
```
evanp, line 15: So, we'll need both session and JWT auth.

evanp (commented):
Overall pretty good. I'm concerned about the complicated file loading; it feels too clever. Is it Sequelize boilerplate?

app/models/index.js
lemilonkh: This is what sequelize generates when being setup, we can probably get rid of this and use our own initialization file (src/models/index.ts). I'm happy to add them manually in src/models/index.ts instead, where we would have the option to reference them and create the relations etc.

We'll just have to change the options for sequelize so it reads the models from there.

app/src/util/api.ts
lemilonkh: Yes, to support a proper (immediate) logout I would suggest using redis for storing the sessions.

lemilonkh:
@evanp yes, this is what `npx sequelize-cli init` generates... I'll get rid of it and load the models manually.

app/models/index.js
evanp: If it's the generated version, let's just use it!

PR #17, opened 2023-08-09
Auth pages, Onboarding, i18n, Cypress e2e tests

lemilonkh:
- Adds all required pages from the designs for authentication and onboarding
- Adds i18next for easy translation management
- Adds English and German translations
- Adds e2e tests for login and signup pages written using Cypress
  - execute them using `npm run cy:run` or watch them be executed visually using `npm run cy:open`


lemilonkh:
1. I tried to move all of the custom styling on buttons etc. that's shared into the Chakra UI theme. This way we can use the raw components from Chakra that immediately have the right styles applied.

Although we could potentially extract the form submit buttons, since they have a bit of custom attributes to prevent form validation from the browser.
2. Do you mean extracting the string with all the classes into a variable in another file? Could be done but I'd only do it for things that are repeated a lot.
3. I'll add comments where appropriate, sure.

evanp:
Oof. This PR is huge and has conflicts. Can you resolve these, and then I'll pull?

PR #18, opened 2023-08-09
health check API endpoint

mfonsecaOEF:


Global-API/health.py
```diff
@@ -0,0 +1,26 @@
...
+Base = declarative_base()
+
+@app.get("/health")
+def health_check():
+    try:
+        engine.connnect()
+        engine.close()
+        return {'status': 'ok'}
+    except:
+        return {'status': 'fail'}
```
lemilonkh, line 26: In this case it would be good to return an HTTP error code >= 400, so k8s automatically knows that the service is offline. Best would probably be 503 (service unavailable).

You can still return the JSON object in addition to that.

Global-API/health.py
```diff
@@ -0,0 +1,26 @@
+from fastapi import FastAPI
+from sqlalchemy import create_engine, exc
+from sqlalchemy.orm import sessionmaker
+from sqlalchemy.ext.declarative import declarative_base
+
+app = FastAPI()
+
+DATABASE_URL = "TBD"
```
lemilonkh, line 8: Let's load this from a .env file that is not added to the repository (since it will contain a password among other secrets, make sure to add it to a .gitignore file).

You can then have a local version of this file to connect to your own database for testing, and the server will have its own as well.

Global-API/health.py
```diff
@@ -0,0 +1,26 @@
...
+# Create a session maker for SQLAlchemy
+SessionLocal = sessionmaker(autocommit=False, autoflush=False, bind=engine)
+
+# Define a base class for SQLAlchemy models
+Base = declarative_base()
+
+@app.get("/health")
+def health_check():
+    try:
+        engine.connnect()
```
lemilonkh, line 22: Typo in method name.
Usually I think you would have a persistent connection to the database that you can just check if it still exists. This might be different for SQLalchemy/ Fast API though, so let's follow the best practices for that.

lemilonkh (commented):
Looks great, I think this can be merged with some minor changes 👍

Could you rename the folder to `global-api` to match the naming convention of the other folders?

lemilonkh (approved):

PR #19, opened 2023-08-14
ON-627: Login endpoint

evanp:
Login endpoint for CityCatalyst


app/src/app/api/auth/[...nextauth]/route.ts
```diff
@@ -0,0 +1,37 @@
...
+
+const handler = NextAuth({
+  providers: [
+    Credentials({
+      name: 'Credentials',
+      credentials: {
+        email: { label: "Email", type: "email", placeholder: "yourname@city.example" },
+        password: { label: "Password", type: "password" }
+      },
+      async authorize(credentials): Promise<any> {
```
Comment (lemilonkh) [inline app/src/app/api/auth/[...nextauth]/route.ts:14]: This is fine for now, but for using sessions and other types of authentication we probably need to set up the [sequelize adapter](https://authjs.dev/reference/adapter/sequelize) and merge our User model with theirs/ pass in our user model and define the necessary relations to their [models](https://authjs.dev/reference/adapters#models). Seems like a task for another PR though.

lemilonkh (approved):

PR #28, opened 2023-08-15
ON-682: Set up the Postgres database schema

lgloege:
This code creates the PostgreSQL database for the global API. Specifically, `create_database.py` creates an `Asset` table to hold raw un-processed ClimateTRACE data. The database parameters will be defined as GitHub Secrets


evanp:
Is it possible to use [Alembic](https://alembic.sqlalchemy.org/en/latest/) for the database management? It should make managing changes and updates to the schema easier.

PR #46, opened 2023-08-22
first code to the API endpoint

mfonsecaOEF:


global-api/routes/city_locode_endpoint.py
```diff
@@ -0,0 +1,135 @@
...
+    "AFOLU": ["V.3.1", "V.1.1","V.2.1"],
+}
+
+def sector_to_filename(inventoryPart):
+    gpc_reference_numbers = gpc_sector_to_reference_number.get(inventoryPart, [])
+    filenames = []
+
+    with SessionLocal() as session:
+        for reference_number in gpc_reference_numbers:
+            query_filenames = text("SELECT DISTINCT(filename) FROM ? WHERE reference_number = :reference_number")
```
lemilonkh, line 31: Do you need to enter the table name here instead of the `?`?

global-api/routes/city_locode_endpoint.py
```diff
@@ -0,0 +1,135 @@
...
+    return filenames
+
+def db_query(filenames, locode, year):
+    results = []
+
+    with SessionLocal() as session:
+        for filename in filenames:
+            query = text(f"SELECT * FROM {filename} "
+                        f"WHERE locode = :locode "
+                        f"AND year = :year")
```
lemilonkh, line 44: This query expects that we have the locode for all entries in the DB. If this ends up being problematic in the import step, I would suggest storing lat, lng coordinates as a PostGIS point so we could query it by the GeoJSON polygon of the city.

If this works out it could be much more efficient to query though 👍

global-api/routes/city_locode_endpoint.py
```diff
@@ -0,0 +1,135 @@
+from fastapi import FastAPI
+from sqlalchemy import create_engine, text 
+from sqlalchemy.orm import sessionmaker
+
+app = FastAPI()
```
lemilonkh, line 5: I think we should have one shared FastAPI app for all routes? I haven't checked the docs though, if this is recommended then go for it.

global-api/routes/city_locode_endpoint.py
```diff
@@ -0,0 +1,135 @@
...
+
+# Read the DATABASE_URL from the .env file
+DATABASE_URL = config("DATABASE_URL")
```
evanp, line 8: If we are going to use `decouple`, you have to add the dependency.

Also, why are we dependent on this .env file thing? Can't we read once from that file, and then just use the env values?

evanp:
The README says we have a skeleton directory with `routes`,`models`, `utils`, and `tests`.

Could we please use the documented structure?

global-api/routes/city_locode_endpoint.py
evanp: @lemilonkh we will have the locode in the `asset` table. I'll work with @mfonsecaOEF and @lgloege to make that happen.

global-api/routes/city_locode_endpoint.py
mfonsecaOEF: this was a suggestion from @lemilonkh

global-api/routes/city_locode_endpoint.py
```diff
@@ -0,0 +1,116 @@
...
+                "gpc_quality": gpc_quality_data
+            }
+        }
+    }
+    
+    # Build Points dictionary
+    Points = {
+        "points": [
+            {
+                "Ownership": {
```
lemilonkh, line 77: Lower case keys perhaps

global-api/routes/city_locode_endpoint.py
```diff
@@ -0,0 +1,116 @@
...
+SessionLocal = sessionmaker(autocommit=False, autoflush=False, bind=engine)
+
+# Extract the data by locode, year and sector/subsector
+def db_query(locode, year, reference_number):
+    
+    with SessionLocal() as session:
+        query = text("SELECT * FROM Asset " +
+                    "WHERE reference_number = :reference_number " +
+                    "AND locode = :locode " +
+                    "AND year = :year ")
```
lemilonkh, line 21: Think about using a query builder from SQLAlchemy here instead of a raw query

global-api/routes/city_locode_endpoint.py
```diff
@@ -0,0 +1,116 @@
...
+engine = create_engine(config("DATABASE_URL"))
+
+# Define SessionLocal for database interactions
+SessionLocal = sessionmaker(autocommit=False, autoflush=False, bind=engine)
+
+# Extract the data by locode, year and sector/subsector
+def db_query(locode, year, reference_number):
+    
+    with SessionLocal() as session:
+        query = text("SELECT * FROM Asset " +
```
lemilonkh, line 18: You might need to use quotes around the table name (Asset). Consider using lower case table names for ease of use (snake_case).

global-api/routes/main.py
```diff
@@ -0,0 +1,30 @@
+from fastapi import FastAPI
```
lemilonkh, line 1: Perhaps move this file into the global-api folder, so it's easy to find the program entry point.

lemilonkh (approved):

PR #47, opened 2023-08-22
ON-709: use Alembic to manage global API database

evanp:
This change adds Alembic as a dependency, and uses Alembic to create the asset table. It also includes some documentation, and removes the old database creation code.


lgloege:
This looks good. We may want to add a column for `emissions_quantity_units`

```python
sa.Column("emissions_quantity_units", sa.String(), nullable=True),
```

since, if I remember correctly, we want the units to be in kg instead of metric tonnes so it is an integer.

PR #48, opened 2023-08-23
feat(ON-425): add importer script for OSM geometries

lgloege:
This adds an importer for OSM geometries. Any feedback would be appreciated.  Works great when tested it out locally.

### DIFF
--- a/Global-API/importer/osm_importer.py
+++ b/Global-API/importer/osm_importer.py
@@ -1,0 +1,76 @@
+import argparse
+import logging
+import os
+import pandas as pd
+from pathlib import Path
+from sqlalchemy import create_engine, insert, MetaData, Table
+
+def record_generator(fl):
+    """returns a generator for the csv file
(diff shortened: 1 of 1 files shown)


Global-API/importer/osm_importer.py
```diff
@@ -0,0 +1,76 @@
...
+            generator = record_generator(file)
+
+            record_counter = 0
+            for record in generator:
+                record_counter += 1
+                logging.info(f"Processing row num: {record_counter} in file: {file} ")
+
+                record = {key: value for key, value in record.items() if value != ''}
+                table_data = {key: record.get(key) for key in record.keys() if key in fields}
+                ins = osm.insert().values(**table_data)
```
evanp, line 68: Can we check if the primary key (locode?) already exists, and do an "update" rather than "insert" if it does?

evanp (approved):

PR #49, opened 2023-08-24
feat(ON-692): extended chakraui theme to use cc tera ui, applied theme to dashboard

cephaschapa:
changes
1. Implemented city catalyst dashboard
2. Extended ChakraUI  theme to use CC TeraUI design specs
3. Applied theme to dashboard and footer
4. added new svg assets too


app/src/app/providers.tsx
```diff
@@ -9,17 +9,179 @@ const openSans = Open_Sans({ subsets: ['latin'] });
...
+      light: '#FFFFFF',
+      dark: '#00001F'
+    },
+
+    border: {
+      neutral: '#D7D8FA',
+      overlay: '#E6E7FF'
+    },
+
+    backround: {
```
lemilonkh, line 42: Typo, possibly also in the usages of this color group

lemilonkh (approved):
Awesome, great to have this!
I have named my colors differently, but will go through and batch rename all of them to fit your naming scheme for consistency.

Also, I have named component files as my-component.tsx instead of MyComponent. I don't have that many yet so I'm fine using PascalCase for the file names, but let's document these decisions in the style guide in the wiki.

Also, it would be good if you could run prettier on these files as a couple have mixed usage of ' and " etc.

We could introduce it as a commit hook if necessary for consistency.

cephaschapa:
Awesome thanks. Will work merging prettier in

PR #50, opened 2023-08-24
ON-681 improvement, application structured, routing fixes done, 

praveensinghrajput23:
This PR is the improved version ON-681 branch.
- Application structured establised.
- API Routing implemented.
- Settings file created to fetch data only once from .env file

After this i will work on get data from database using SqlAlchemy instead of raw queries.
- Should I also merge the importer directory in this application itself??
- Should I dockerize the project?


praveensinghrajput23:
please review the pr. i have followed the structure mentioned so that it will be good to structure the code

PR #51, opened 2023-08-25
feat(ON-684): added global api deployement, service, db and actions

cephaschapa:


.github/workflows/global-api-actions.yml
```diff
@@ -0,0 +1,92 @@
...
+          createuser -w -h localhost -p 5432 -U postgres citycatalyst
+          createdb -w -h localhost -p 5432 -U postgres citycatalyst -O citycatalyst
+          cd app
+          cp env.example .env
+          npm run db:migrate
+
+      - name: Run NextJS build
+        run: |
+          cd app
+          npm run build
```
lemilonkh, line 39: This file needs to be adjusted to the global-api project - use python instead of NodeJS, run python code validation (e.g. `black --check` for code style checking, later on tests, build the dockerfile for it at the end

k8s/cc-global-api-db.yml
```diff
@@ -0,0 +1,38 @@
...
+      app: db
+  template:
+    metadata:
+      labels:
+        app: db
+    spec:
+      volumes:
+        - name: postgres
+          persistentVolumeClaim:
+            claimName: cc-db-pvc
```
lemilonkh, line 20: This might need to be different, so the global-api DB doesn't use the volume for the CityCatalyst DB

k8s/cc-global-api-db.yml
```diff
@@ -0,0 +1,38 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: cc-db-deploy
```
lemilonkh, line 4: Name needs to be adjusted

lemilonkh (changes requested):

PR #52, opened 2023-08-25
Manual Input Wizard

lemilonkh:
- Adds /data and /data/[step] pages with lots of forms and UI
- Links to /data and /data/[step] from the dashboard
- Creates custom components RadioButton and TagSelect, as well as page specific ones in the [step] folder


cephaschapa (approved):
Great work!
Please let me know if there are some components on the dash that you created that match the ones that are currently there. And we can reuse or change.

PR #63, opened 2023-08-31
city boundary geometry using UN LOCODE

mfonsecaOEF:


global-api/routes/city_boundaries_endpoint.py
```diff
@@ -0,0 +1,30 @@
...
+            + "WHERE locode = :locode "
+        )
+        result = session.execute(
+            query,
+            {"locode": locode},
+        ).fetchone()
+
+    return result[0] if result else None
+
+@app.get("/api/v0/cityboundary/city/{locode}")
```
lemilonkh, line 21: Let's merge this, but we should afterwards switch to a router-based setup where the main file imports these routes definitions that themselves define a router.

See this PR for an example: https://github.com/Open-Earth-Foundation/CityCatalyst/pull/50/files#diff-8db69d50a2770566ee1a73a0452ceedac873221aa95eb5bc5d35accf25cddd03R41

lemilonkh: This way we can launch the server as one program and have all routes available from it, instead of starting each route as an individual python process, which isn't practical on a server.

lemilonkh (approved):

global-api/routes/city_boundaries_endpoint.py
mfonsecaOEF: Ok, got it!

PR #64, opened 2023-09-01
ON-731 - Import emissions factors US

lgloege:
This contains a script to process emissions factors for the US from EFDB and CSV files to import the data into the database. Feedback is greatly appreciated, especially on how to handle UUIDs. Right now I am setting the UUID in the script, which probably isn't the best solution because these IDs will change if you re-run the script.


lemilonkh (approved):

emissions_factors/scripts/utils.py
```diff
@@ -0,0 +1,64 @@
+import csv
+from pathlib import Path
+import uuid
+
+
+def uuid_generate_v4():
+    """generate a version 4 UUID"""
+    return str(uuid.uuid4())
```
lemilonkh, line 8: Can you implement a seeded version of this (seed the random generator once when the script starts running and then use it to generate the UUID)?

See here: https://stackoverflow.com/a/50743795
You could use either [`faker`](https://github.com/joke2k/faker) or `hashlib` to achieve this as mentioned there. Hashing is a great way to turn string IDs that are consistent into UUIDv4's. Just note that UUIDv4 has [6 bits that are meant to be hardcoded](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_.28random.29) and not randomly selected, so using faker might be better as that library handles it for you.

emissions_factors/scripts/utils.py
```diff
@@ -0,0 +1,64 @@
...
+        the path to the directory you want to create
+
+    Returns
+    ---------
+    None:
+
+    Example
+    --------
+    make_dir('/path/to/new/directory')
+    """
```
lemilonkh, line 58: Quite a bit of documentation for a method that just creates a folder, but nice to have I suppose.

lemilonkh (approved):

lemilonkh (approved):

PR #74, opened 2023-09-06
feat(ON-774): add new migration to CC EmissionsFactor

lgloege:
This adds the following columns to the `EmissionsFactor` table:
-  `gpc_refno`: GPC reference number
- `gas`: chemical formula of the gas
- `region`: region code the EF is applicable for (eg. CA) [optional]
- `time_frame`: time frame the EF is applicable for (eg 2010)  [optional]
- `reference`: reference for the EF [optional] 

This should be enough to uniquely define the emissions factor. Please let me know if any changes are necessary.

### DIFF
--- a/app/migrations/20230906193347-modify-EmissionsFactor.cjs
+++ b/app/migrations/20230906193347-modify-EmissionsFactor.cjs
@@ -1,0 +1,52 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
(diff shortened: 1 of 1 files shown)


lgloege:
some sources include a range of emission factors. Do we want to capture the range or just state that we use the median if a range is given. In the CSVs I have been capturing the range as well as the median, but not sure if we will ever use that range.

lemilonkh (approved):

PR #79, opened 2023-09-08
ON-776: create import script

lgloege:
This is an an import script to load CSVs into the citycatalyst database. This is based on the [script used for openclimate](https://github.com/Open-Earth-Foundation/OpenClimate/blob/develop/import/import_openclimate_data.py). 

ChatGPT helped write these functions to sort tables in dependency order:

```python
    def topological_sort(graph, node, visited, stack):
        visited[node] = True
        for neighbor in graph[node]:
            if not visited[neighbor]:
                topological_sort(graph, neighbor, visited, stack)
        stack.append(node)

    def get_topological_order(graph):
        visited = {node: False for node in graph}
        stack = []
        for node in graph:
            if not visited[node]:
                topological_sort(graph, node, visited, stack)
        return stack
```

to run the importer locally, first set the following environment variables

```bash
export CC_DATABASE=citycatalyst
export CC_USER=citycatalyst
export CC_HOST=localhost
export CC_PASSWORD=''
```

then run the importer:

```bash
python import_citycatalyst_data.py path/to/directory/with/csvs
```

### DIFF
--- a/emissions_factors/data_processed/EFDB_US/Publisher.csv
+++ b/emissions_factors/data_processed/EFDB_US/Publisher.csv
@@ -1,2 +1,2 @@
-publisher_id,publisher_name,URL
+publisher_id,name,URL
 5be81e5d-1d8a-45b5-9ad8-389f699afb04,IPCC,https://www.ipcc.ch/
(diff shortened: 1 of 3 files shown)


lemilonkh:
We might want to import the data instead using [sequelize seeders](https://sequelize.org/docs/v6/other-topics/migrations/#creating-the-first-seed). Let's discuss the best approach here. It's more custom code that way, but we have full control over how the data is imported.

lgloege:
sounds good. I mainly built this to test out importing data locally.

evanp:
I think this is a useful enough utility that we should include it.

PR #94, opened 2023-09-14
Dashboard API

lemilonkh:
- Adds progress endpoint for inventory
- Adds test
- Adds column `source` to `SubSectorValue` (and migration for it)


app/tests/api/inventory.test.ts
```diff
@@ -145,4 +167,102 @@ describe("Inventory API", () => {
...
+            subsectorId,
+            sectorValueId,
+            datasourceId: sources[i]?.datasourceId,
+            inventoryId: existingInventory!.inventoryId,
+          });
+        }
+      }
+    }
+
+    const url = `http://localhost:3000/api/v0/city/${locode}/inventory/${inventory.year}/progress`;
```
cephaschapa, line 221: Should this be just /api/v0/.....?

cephaschapa (approved):

app/tests/api/inventory.test.ts
lemilonkh: So this needs to be a fully qualified URL in order to successfully pass the URL parser.
The thing is it's not actually used anywhere, so we might as well pass a placeholder URL in for all routes.
I just did this mainly for documentation in the tests, so it's clear which route is used.

lemilonkh: The URL never actually gets fetched as I call the endpoint handler directly to make the tests faster.

PR #102, opened 2023-09-18
feat(ON-809): add EDGAR tables to ccglobal API database

lgloege:
this adds three tables to the ccglobal API Database. I named them to be specifically used for [EDGAR data](https://edgar.jrc.ec.europa.eu/dataset_ghg70#p3), but they may generalizable. This PR adds three tables:
- `edgar_grid`: this contains relevant grid information (centroid, polygon in WKT, and area)
- `edgar_data`: this contains the emissions data for each gas. The corresponding GPC reference number is included and there is a FK to `edgar_grid`
- `edgar_city_data`: this contains the fraction of each grid cell that is within each locode. Only grid cells where the fraction is >0 will be included.

`edgar_data` contains all the raw data, so this could be used for custom boundaries. While `edgar_city_data` contains the fractions of grid cells within the locode geometries we got from OSM.

### DIFF
--- a/global-api/migrations/versions/64eb87bae0a3_add_edgar_tables.py
+++ b/global-api/migrations/versions/64eb87bae0a3_add_edgar_tables.py
@@ -1,0 +1,57 @@
+"""add-EDGAR-tables
+
+Revision ID: 64eb87bae0a3
+Revises: bd6c5bbd6eb4
+Create Date: 2023-09-18 12:25:22.687198
+
+"""
+from typing import Sequence, Union
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.sql import text
+
+# revision identifiers, used by Alembic.
+revision: str = '64eb87bae0a3'
+down_revision: Union[str, None] = 'bd6c5bbd6eb4'
+branch_labels: Union[str, Sequence[str], None] = None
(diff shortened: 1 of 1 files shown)


global-api/migrations/versions/64eb87bae0a3_add_edgar_tables.py
```diff
@@ -0,0 +1,57 @@
...
+        sa.Column('area', sa.Float, nullable=False),
+        sa.Column("created_date", sa.DateTime(), nullable=False),
+        sa.Column("modified_date", sa.DateTime(), server_default=text('CURRENT_TIMESTAMP'))
+    )
+
+    op.create_table('GridCellEmissionsEdgar',
+        sa.Column('id', sa.dialects.postgresql.UUID(as_uuid=True), primary_key=True),
+        sa.Column('year', sa.Integer, nullable=False),
+        sa.Column('reference_number', sa.String, nullable=False),
+        sa.Column("gas", sa.String(), nullable=False),
```
lemilonkh, line 36: Would be good to make the use of ' and " consistent. You can probably do it automatically with the black formatter.

lemilonkh (approved):

evanp:
LGTM. Thanks for the work, @lgloege

PR #106, opened 2023-09-21
Feature/connect onboarding

cephaschapa:
Changes
- Adds openclimate citySlice
- Adds openclimate cityReducer to store
- Adds concatenate openclimate api to store middleware
- Adds openclimateApi service for text search queries
- Adds get city mutation to api service for adding new city
- Adds dropdown with select options
- Adds recent searches component
- Adds hooks and functions to handle selected dropdown values 
- Adds dispatch and select redux hooks to persist reusable city data during on boarding
- Adds check mark functionality (we can use this to check if city exists)


app/src/components/recent-searches.tsx
```diff
@@ -0,0 +1,122 @@
...
+                <Text
+                  className="group-hover:text-white"
+                  color="content.secondary"
+                  fontSize="body.lg"
+                  fontFamily="body"
+                  fontWeight="normal"
+                  lineHeight="24"
+                  letterSpacing="wide"
+                >
+                  Rio de Janeiro
```
lemilonkh, line 34: Why is this hardcoded? If you don't have the data for this yet, you can have an array at the top of the component and render the entries here. Otherwise let's save the recent entries into local storage.
One way to do that would be to use redux-persist:
https://blog.logrocket.com/persist-state-redux-persist-redux-toolkit-react/

app/src/components/recent-searches.tsx
```diff
@@ -0,0 +1,122 @@
...
+                </Text>
+                <Text
+                  className="group-hover:text-[#E8EAFB]"
```
lemilonkh, line 37: You can add custom colors to the Tailwind config.
In this case better would be the following Chakra attribute, which allows you to use theme colors:
`_hover={{ color: "color.name" }}`

app/src/features/city/openclimateCitySlice.ts
```diff
@@ -0,0 +1,31 @@
...
+interface CityState {
+  city?: OCCityArributes;
+}
+
+const initialState = {
+  city: undefined,
+} as CityState;
+
+export const openclimateCitySlice = createSlice({
+  name: "openclimatecity",
```
lemilonkh, line 14: Should be either `openClimateCity` or just `city` and we remove citySlice in favor of this.

app/src/services/api.ts
```diff
@@ -26,5 +29,28 @@ export const api = createApi({
       transformResponse: (response: { data: InventoryProgressResponse }) =>
         response.data,
     }),
+    addCity: builder.mutation<CityAttributes, {}>({
```
lemilonkh, line 32: Would be good to add tags to the city query and using `invalidatesTags` here to clear the cached city.

app/src/services/api.ts
```diff
@@ -26,5 +29,28 @@ export const api = createApi({
...
+        body: data,
+      }),
+    }),
+  }),
+});
+
+export const openclimateAPI = createApi({
+  reducerPath: "openclimateapi",
+  baseQuery: fetchBaseQuery({
+    baseUrl: "https://openclimate.openearth.dev",
```
lemilonkh, line 45: This URL should probably depend on the environment we're using, for prod we should use the prod OpenClimate API. You can extract it to an env var or decide which one to use based on `process.env.NODE_ENV === 'production'`

app/src/services/api.ts
```diff
@@ -26,5 +29,28 @@ export const api = createApi({
...
+  }),
+  endpoints: (builder) => ({
+    getOCCity: builder.query<any, string>({
```
lemilonkh, line 48: You should use `OCCityAttributes` here instead of any to get type safety

PR #113, opened 2023-09-25
ON-791: Add i18n to CityCatalyst dashboard

cephaschapa:
- Adds dashboard.json to for en and es
- Replaces text with translation


app/src/app/[lng]/page.tsx
```diff
@@ -186,7 +190,7 @@ export default function Home({ params: { lng } }: { params: { lng: string } }) {
                   lineHeight="32"
                   fontWeight="semibold"
                 >
-                  Welcome Back,
+                  <Trans t={t}>welcome-back</Trans>,
```
Comment (lemilonkh) [inline app/src/app/[lng]/page.tsx:193]: This is fine, but just FYI you don't necessarily have to use the Trans component in all places, this works too:
`{t("welcome-back")}`

app/src/app/[lng]/page.tsx
```diff
@@ -403,7 +401,10 @@ export default function Home({ params: { lng } }: { params: { lng: string } }) {
                 fontWeight="semibold"
                 lineHeight="32"
               >
-                GPC Basic Emission Inventory Calculation - Year {year}
+                <Trans t={t}>
+                  gpc-basic-emissions-inventory-calculations-year
+                </Trans>{" "}
+                {year}
```
Comment (lemilonkh) [inline app/src/app/[lng]/page.tsx:407]: You can have placeholders for data in translation strings: `{{year}}`, and then pass this data into the `t` function or the `Trans` component. Here this works but there are more complex examples of having variables in the middle of text etc. where it becomes relevant.

app/src/app/[lng]/page.tsx
```diff
@@ -413,18 +414,16 @@ export default function Home({ params: { lng } }: { params: { lng: string } }) {
               color="interactive.control"
               letterSpacing="wide"
             >
-              The data you have submitted is now officially incorporated into
-              your city&apos;s {year} GHG Emissions Inventory, compiled
-              according to the GPC Basic methodology.{" "}
+              {t("gpc-data-description")} {year} {t("add-data-description-ctn")}
```
Comment (lemilonkh) [inline app/src/app/[lng]/page.tsx:417]: Here using a placeholder in the translation string would be better.  
Relevant docs:  
https://www.i18next.com/translation-function/interpolation  
https://react.i18next.com/latest/trans-component#using-with-react-components

lemilonkh (approved):

PR #123, opened 2023-10-02
ON-840: importer for EDGAR API

lgloege:
this PR creates import scripts for the EDGAR API endpoint. 

- `gridcelledgar_importer.py`: this imports the grid cell information and **needs to run before first**
- `gridcellemissionsedgar_importer.py`: this imports the emissions at each grid cell
- `citycelloverlapedgar_importer.py`: this imports the fraction of each cell that is within each city boundary.

I tested these out locally and they all work.

The [EDGAR tables PR](https://github.com/Open-Earth-Foundation/CityCatalyst/pull/102) needs to be merged before this one


global-api/importer/edgar/citycelloverlapedgar_importer.py
```diff
@@ -0,0 +1,97 @@
...
+    df_grid = (
+        pd.DataFrame(list_grid_coords)
+        .rename(columns={"id": "cell_id"})
+        .astype({"cell_id": str})
+    )
+
+    table = Table("CityCellOverlapEdgar", metadata_obj, autoload_with=engine)
+
+    lon_res = 0.1
+    lat_res = 0.1
```
lemilonkh, line 47: These should maybe be defined as constants at the top. So they are easy to adjust if we need to for e.g. another data source that uses a similar structure.

global-api/importer/edgar/gridcelledgar_importer.py
```diff
@@ -0,0 +1,64 @@
...
+    engine = create_engine(args.database_uri)
+    metadata_obj = MetaData()
+
+    table = Table("GridCellEdgar", metadata_obj, autoload_with=engine)
+
+    lats = ds.lat.values
+    lons = ds.lon.values
+
+    lat_res = 0.1  # degrees
+    lon_res = 0.1  # degrees
```
lemilonkh, line 42: Could be imported from a common file or at least defined at the top

lemilonkh (approved):

global-api/importer/edgar/citycelloverlapedgar_importer.py
lgloege: Thanks, I'll define them at the top, they are constants for the EDGAR, but this will not be the case for telescoping grids. 

Another approach is to not hard code the values, but use the discrete difference along the lat and lon vectors. We could use [decimal](https://docs.python.org/3/library/decimal.html) to avoid floating point arithmetic error. This would allow us to write a general function that works with most grids. I decided against this for now because the resolution is constant for EDGAR so hard-coding was the simpler and quicker approach.

lemilonkh: Agreed, let's keep the constants but just make sure they are easy to find and reused between the scripts for EDGAR.

PR #127, opened 2023-10-03
Crosswalk data

mfonsecaOEF:


global-api/importer/crosswalk/crosswalk_grid_importer.py
```diff
@@ -0,0 +1,70 @@
...
+    ds = get_crosswalk('US', 'airport', 'hi')
+
+    engine = create_engine(args.database_uri)
+    metadata_obj = MetaData()
+
+    table = Table("crosswalk_GridCell", metadata_obj, autoload_with=engine)
+
+    lats = ds.lat.values
+    lons = ds.lon.values
+
```
lgloege, line 46: This can be made more efficient. One solution is do something like this:

```python
lats = ds.lat.values.flatten()
lons = ds.lon.values.flatten()

assert len(lats) == len(lons), 'len of lats and lons do not match'

for lat, lon in zip(lats, lons):
    # do something with lat and lon
```

I'm "flattening" the arrays so they're 1D and, as a sanity check, making sure the lengths are the same.  With this solution we're only looping through the coordinates once.

global-api/importer/crosswalk/utils.py
```diff
@@ -0,0 +1,532 @@
...
+    x_dim="lon",
+    y_dim="lat",
+    *args,
+    **kwargs,
+):
+    """previously called get_intersection"""
+    return ds.rio.set_spatial_dims(x_dim=x_dim, y_dim=y_dim).rio.clip(
+        geometries, crs=crs, all_touched=all_touched
+    )
+
```
lgloege, line 393: This works. However, since the data is hosted on a THREDDS data server (TDS) we could use unidata's [siphon package](https://github.com/Unidata/siphon). Below is some code showing how to access crosswalk data using siphon

```python
from siphon.catalog import TDSCatalog
import xarray as xr

def tds_catalog(url):
    """create a catalog client"""
    return TDSCatalog(url)


def tds_generator(catalog):
    """generator for files in a THREDDS catalog"""
    yield from (dataset for dataset in catalog.datasets)
        

def get_dataset_url(catalog, dataset):
    """retrieve the URL to the dataset"""
    return catalog.datasets[dataset].access_urls["OPENDAP"]


# ===================================
# example
# ===================================

# create TDS catalog
url = "https://thredds.daac.ornl.gov/thredds/catalog/ornldaac/1741/catalog.xml"
catalog = tds_catalog(url)

# lists all the datasets
gen = tds_generator(catalog)
datasets = [fl for fl in gen]

# filter the datasets by keywords
keywords = ['US', 'airport']
datasets_filtered = [dataset for dataset in datasets if dataset.endswith('_hi.nc4') and all(keyword in dataset for keyword in keywords)]

# read dataset_url
dataset_url = get_dataset_url(catalog, datasets_filtered[0])
ds = xr.open_dataset(dataset_url)
```

PR #143, opened 2023-10-09
crosswalk endpoint

mfonsecaOEF:


global-api/routes/city_locode_endpoint_crosswalk.py
```diff
@@ -0,0 +1,70 @@
...
+        "totals": {
+            "emissions": {
+                "co2_mass": str(series.get("CO2", 0)),
+                "co2_co2eq": str(series.get("CO2", 0)),
+                "gpc_quality": str(gpc_quality_data),
+            }
+        }
+    }
+
+    return {"totals": totals}
```
lgloege, line 70: Change `return {"totals": totals}` to ` return {totals}` so "totals" isn't duplicated in the response

lgloege:
Make sure to update `global-api/main.py` so the endpoint is included in the app.

in `main.py`, import the router

```python
from routes.city_locode_endpoint_crosswalk import api_router as crosswalk_city_locode_route
```

then include the router in the app

```python
app.include_router(
    crosswalk_city_locode_route,
    tags=["Crosswalk Labs"],
)
```

global-api/routes/city_locode_endpoint_crosswalk.py
lemilonkh: Or actually just `return totals` in this case, since totals is a dict already.

global-api/routes/city_locode_endpoint_crosswalk.py
```diff
@@ -0,0 +1,68 @@
...
+        .astype({"emissions_total": int})
+        .loc[:, ["emissions_total"]]
+        .squeeze()
+    )
+
+    return {
+        "totals": {
+            "emissions": {
+                "co2_mass": str(series.get("CO2", 0)),
+                "co2_co2eq": str(series.get("CO2", 0)),
```
lemilonkh, line 64: This should be integers

lemilonkh: @mfonsecaOEF can you remove the str() calls?

mfonsecaOEF: Ready!

PR #146, opened 2023-10-12
feat: download endpoint for CSV; closes ON-856

evanp:
This is the download endpoint for the CSV file. I still need to add the code in the XLS endpoint, but there's a function for it in there.

I added tests to check that the CSV and XLS endpoints exist, and a test to fetch the CSV endpoint, and check its results.


app/src/app/api/v0/city/[city]/inventory/[year]/route.ts
```diff
@@ -18,9 +20,76 @@ export const GET = apiHandler(async (_req: NextRequest, { params }) => {
   }
   inventory.city = city;
 
-  return NextResponse.json({ data: inventory });
+  let body:Buffer|null = null;
+  let headers:Record<string, string>|null = null;
+
+  switch (_req.nextUrl.searchParams.get('format')?.toLowerCase()) {
```
Comment (lemilonkh) [inline app/src/app/api/v0/city/[city]/inventory/[year]/route.ts:26]: Since you're now using the `_req` argument, let's rename it to `req` (_ indicates it is unused and prevents a linter warning about unused arguments).

app/src/app/api/v0/city/[city]/inventory/[year]/route.ts
```diff
@@ -18,9 +20,76 @@ export const GET = apiHandler(async (_req: NextRequest, { params }) => {
   }
   inventory.city = city;
 
-  return NextResponse.json({ data: inventory });
+  let body:Buffer|null = null;
+  let headers:Record<string, string>|null = null;
```
Comment (lemilonkh) [inline app/src/app/api/v0/city/[city]/inventory/[year]/route.ts:24]: You can run prettier (Shift + Option + F on VS Code on Mac to do it for the currently open file, or run `prettier` in the project directory (`app`) to do it for all) to format the code to a common style.

app/src/app/api/v0/city/[city]/inventory/[year]/route.ts
```diff
@@ -18,9 +20,76 @@ export const GET = apiHandler(async (_req: NextRequest, { params }) => {
...
+    return [
+      value.subcategoryId,
+      value.totalEmissions,
+      value.activityUnits,
+      value.activityValue,
+      value.emissionFactorValue,
+      value.datasourceId
+    ].join(",");
+  });
+  return Buffer.from([headers, ...subSectorLines, ...subCategoryLines].join("\n"));
```
Comment (lemilonkh) [inline app/src/app/api/v0/city/[city]/inventory/[year]/route.ts:86]: Should this maybe be structured as 1 subsector line, then all of its subcategories, then the next subsector? We will add the GPC reference number (see [this PR](https://github.com/Open-Earth-Foundation/CityCatalyst/pull/145)) to the SubSector/ SubCategory data model, so it could be used instead of the values for the first column here, which will just be UUIDs that won't mean much to an end user (unless it needs to be imported into CC again for some reason...)

app/tests/api/inventory.test.ts
```diff
@@ -110,6 +175,43 @@ describe("Inventory API", () => {
...
+    const lines = csv.split("\n");
+    assert.ok(lines.length > 0);
+    const headers = lines[0].split(",");
+    assert.equal(headers.length, 6);
+    assert.equal(headers[0], "Inventory Reference");
+    assert.equal(headers[1], "Total Emissions");
+    assert.equal(headers[2], "Activity Units");
+    assert.equal(headers[3], "Activity Value");
+    assert.equal(headers[4], "Emission Factor Value");
+    assert.equal(headers[5], "Datasource ID");
```
lemilonkh, line 197: You can test all of this in one assertion with `assert.deepEqual(headers, ["Inventory Reference", ...]);`, which will also have more helpful error messages.

app/tests/api/inventory.test.ts
```diff
@@ -110,6 +175,43 @@ describe("Inventory API", () => {
...
+    assert.ok(lines.length > 1);
+    assert.strictEqual(lines.length, 3);
+    assert.ok(lines.slice(1).every(line => line.split(",").length == 6));
+  });
+
+  it("should find an inventory with xls format", async () => {
+    const url = `http://localhost:3000/api/v0/city/${locode}/inventory/${inventory.year}?format=xls`;
```
lemilonkh, line 204: Since this URL actually doesn't serve a function in the tests, I've resulted to make a new function `mockRequest` that doesn't take a URL as the argument. Otherwise this would be hard to keep in sync since it's not checked/ enforced by the test or route for any test case.

lemilonkh (approved):
Just some minor style/ detail feedback, LGTM otherwise! :sparkles:

app/src/app/api/v0/city/[city]/inventory/[year]/route.ts
Comment (evanp) [reply app/src/app/api/v0/city/[city]/inventory/[year]/route.ts]: Yeah, I think the sorting there can be handled by the downloader.

PR #170, opened 2023-10-23
Feature/settings and profile

cephaschapa:
Chages
- [x] Adds User API routes
- [x] Adds settings page
- [x] Adds my profile, my files and my inventories
- [x] Adds nested tabs for my profile, my files and my inventories 
- [x] Adds users table with filters
- [x] Adds input forms for modifying users
- [x]  Adds table to manage cities
- [x] Adds table to manage files
- [x] Adds table to manage inventories
- [x] Adds modals for all actions
-
- [ ] integrate inventory api
- [ ] integrate file api
- [ ] integrate users api


app/src/app/api/v0/city/[city]/user/[user]/route.ts
```diff
@@ -0,0 +1,58 @@
+import { db } from "@/models";
+import { apiHandler } from "@/util/api";
+import { createCityRequest, createUserRequest } from "@/util/validation";
+import createHttpError from "http-errors";
+import { NextRequest, NextResponse } from "next/server";
+
+export const GET = apiHandler(async (_req: NextRequest, { params }) => {
```
Comment (lemilonkh) [inline app/src/app/api/v0/city/[city]/user/[user]/route.ts:7]: Are we sure the user routes need to be nested under the `city/[city]` folder? It could be `api/v0/user/[user]` instead, which is shorter and it seems like the city isn't really necessary here...

app/src/app/api/v0/city/[city]/user/[user]/route.ts
```diff
@@ -0,0 +1,58 @@
...
+  const locode = params.city;
+  const city = await db.models.City.findOne({ where: { locode } });
+  if (!city) {
+    throw new createHttpError.NotFound("City not found");
+  }
```
Comment (lemilonkh) [inline app/src/app/api/v0/city/[city]/user/[user]/route.ts:12]: You're not using this so it can probably be removed.

app/src/app/api/v0/city/[city]/user/route.ts
```diff
@@ -0,0 +1,24 @@
...
+import createHttpError from "http-errors";
+import { NextRequest, NextResponse } from "next/server";
+import { randomUUID } from "node:crypto";
+
+export const POST = apiHandler(async (_req: NextRequest, { params }) => {
+  const body = createUserRequest.parse(await _req.json());
+  const user = await db.models.User.create({
+    userId: randomUUID(),
+    ...body,
+  });
```
Comment (lemilonkh) [inline app/src/app/api/v0/city/[city]/user/route.ts:13]: If we create a user this way, it won't have any login information attached. So it's maybe required to allow for the user to sign up afterwards still (which will probably fail if a record with their email exists already). So if we use this we need to change the signup endpoint as well (and perhaps set an `isInvited` boolean field in the user record here to show that it's not a fully registered user yet?).

app/src/components/Modals/add-user-modal.tsx
```diff
@@ -0,0 +1,106 @@
...
+            padding="24px"
+            borderBottom={"1px solid #fafafa"}
+          >
+            Add User
+          </ModalHeader>
+          <ModalCloseButton marginTop="10px" />
+          <ModalBody>
+            <form
+              onSubmit={handleSubmit(() => {
+                alert("hello");
```
lemilonkh, line 57: Can be removed?

app/src/components/Modals/add-user-modal.tsx
```diff
@@ -0,0 +1,106 @@
...
+                />
+
+                <FormSelectInput
+                  label="Role"
+                  value="admin"
+                  register={register}
+                  error={errors.role}
+                  id="role"
+                  onInputChange={() => {}}
+                />
```
lemilonkh, line 81: This should be `disabled` if the current user is not an admin? Or maybe we don't allow editing any users unless the user is an admin...

This also needs to be reflected in the access control of the user routes.

app/src/components/Modals/delete-inventory-modal.tsx
```diff
@@ -0,0 +1,176 @@
...
+                  textAlign="center"
+                  fontFamily="heading"
+                  w="408px"
+                  fontSize="body.large"
+                  letterSpacing="wide"
+                  fontStyle="normal"
+                >
+                  Are you sure you want to{" "}
+                  <span style={{ fontWeight: "bold" }}>permanently delete</span>{" "}
+                  this city&pos;s from CityCatalyst?
```
lemilonkh, line 102: i18n
Also the &pos; is visible on the dialog (I think it's `&apos;` instead)

app/src/components/Modals/delete-inventory-modal.tsx
```diff
@@ -0,0 +1,176 @@
...
+                        <Text
+                          fontSize="body.md"
+                          fontStyle="normal"
+                          lineHeight="20px"
+                          fontFamily="heading"
+                          color="content.tertiary"
+                          letterSpacing="wide"
+                        >
+                          Enter your password to confirm the deletion of the
+                          cities information.
```
lemilonkh, line 144:
```suggestion
                          city's information.
```

app/src/components/form-select-input.tsx
```diff
@@ -0,0 +1,54 @@
...
+        background={isDisabled ? "background.neutral" : "background.default"}
+        color={isDisabled ? "content.tertiary" : "content.secondary"}
+        value={value || ""}
+        {...register(id, {
+          required: `${id} is required`,
+        })}
+        onChange={(e) => onInputChange(e)}
+      >
+        <option value="admin">Admin</option>
+        <option value="contributer">Contributer</option>
```
lemilonkh, line 48:
```suggestion
        <option value="contributor">Contributor</option>
```

app/src/components/navigation-bar.tsx
```diff
@@ -147,8 +152,40 @@ export function NavigationBar({
...
+                display="flex"
+                flexDirection="column"
+                justifyContent="space-around"
+                height="128px"
+                zIndex={2000}
+              >
+                <NextLink href="/settings">
+                  <MenuItem
+                    paddingTop="12px"
+                    paddingBottom="12px"
```
lemilonkh, line 169: You can use py={3} instead of these two lines as a shorthand

app/src/components/navigation-bar.tsx
```diff
@@ -147,8 +152,40 @@ export function NavigationBar({
...
+                    paddingLeft="16px"
+                    paddingRight="16px"
```
lemilonkh, line 171: px={4}

app/src/components/navigation-bar.tsx
```diff
@@ -147,8 +152,40 @@ export function NavigationBar({
...
+                      mr={4}
+                    />
+                    {t("settings")}
+                  </MenuItem>
+                </NextLink>
+                <MenuItem
+                  paddingTop="12px"
+                  paddingBottom="12px"
+                  paddingLeft="16px"
+                  paddingRight="16px"
```
lemilonkh, line 186: Can be shortened same as above

app/src/components/navigation-bar.tsx
```diff
@@ -147,8 +152,40 @@ export function NavigationBar({
...
+                  onClick={() => signOut()}
```
lemilonkh, line 187:
```suggestion
                  onClick={signOut}
```

app/src/models/CityUser.ts
```diff
@@ -51,6 +51,7 @@ export class CityUser
           type: DataTypes.UUID,
           allowNull: false,
           primaryKey: true,
+          defaultValue: Sequelize.UUIDV4,
```
lemilonkh, line 54: We should do this for most models probably. Makes it easier to create new entries without having to manually create a `randomUUID` each time 👍

app/src/util/validation.ts
```diff
@@ -82,3 +82,13 @@ export const createSubCategory = z.object({
 });
 
 export type CreateSubCategory = z.infer<typeof createSubCategory>;
+
+export const createUserRequest = z.object({
```
lemilonkh, line 86: Could be moved to before the route definition to make it easier to see what data is available/ how the validation works (at least if it's only used in one place)

app/src/app/api/v0/city/[city]/user/[user]/route.ts
Comment (cephaschapa) [reply app/src/app/api/v0/city/[city]/user/[user]/route.ts:7]: right right, i was actually debating. Ill resolve it

app/src/components/navigation-bar.tsx
cephaschapa: wait that actually works ... awesome

PR #189, opened 2023-11-02
Fix asset table migration

lgloege:
This PR makes two changes to the `asset` table:
- changes `emissions_quantity` from type `integer` to `bigint`, otherwise can not store data with units kg
- changes `id` column from auto-incrementing integer to UUID primary key.


global-api/migrations/versions/e4a7d83a0339_alter_column_asset_id_to_uuid.py
```diff
@@ -0,0 +1,34 @@
...
+revision: str = "e4a7d83a0339"
+down_revision: Union[str, None] = "66d4e6787ba6"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.drop_column("asset", "id")
+    op.add_column(
+        "asset", sa.Column("id", sa.dialects.postgresql.UUID, primary_key=True)
```
lemilonkh, line 24: Can you specify a default value of UUIDv4() basically, that sets a random UUID?

lemilonkh (approved):

evanp:
I just copy-and-pasted to main. Looks like it's working. Thanks @lgloege !

PR #200, opened 2023-11-06
Feature/city outline

cephaschapa:
changes
- adds global api to boundary route
- adds `wellknown` to convert global api boundary data to geoJson multipolygon leaflet format
- adds adds CityMap component to onboarding
- adds api to fetch contextual data (area, region, coundty) from openclimate
- Adds contextual data to onboarding and dashboard


app/src/services/api.ts
```diff
@@ -87,6 +94,9 @@ export const openclimateAPI = createApi({
     getOCCity: builder.query<any, string>({
       query: (q) => `/api/v1/search/actor?q=${q}`,
     }),
+    getOCCityData: builder.query<any, string>({
```
lemilonkh, line 97: Instead of any here you could use {locode: string} (same for the others).
This means you have type safety when using this hook in a component.

lemilonkh:
Please resolve the merge conflicts

app/src/services/api.ts
lemilonkh: Nvm, the response is marked as any. Would be nice to add the type definition for that though, so the usage of API data can be properly type checked.

PR #201, opened 2023-11-06
Fix/qa bug fixes

cephaschapa:
changes
- QA bug fixes for layout
- QA bug fixes for styling
- QA bug fixes for missing links


app/src/app/[lng]/data/[step]/SourceDrawer.tsx
```diff
@@ -95,16 +95,15 @@ export function SourceDrawer({
               </Stack>
             </DrawerBody>
           )}
+          <Stack
+            w="full"
+            className="drop-shadow-top border-t-2 absolute left-0 flex justify-center items-center"
```
Comment (lemilonkh) [inline app/src/app/[lng]/data/[step]/SourceDrawer.tsx:100]: Keeping the padding might be good in case the width of the source drawer is lower than the hardcoded button width.

app/src/app/[lng]/data/[step]/SourceDrawer.tsx
```diff
@@ -95,16 +95,15 @@ export function SourceDrawer({
...
+          >
+            <Button onClick={onConnectClick} w="543px" h={16} my={6}>
```
Comment (lemilonkh) [inline app/src/app/[lng]/data/[step]/SourceDrawer.tsx:102]: Do we need to hardcode the width or can we just add horizontal padding and make it depend on the text?

Helpful once we translate this page.

app/src/i18n/locales/en/data.json
```diff
@@ -34,6 +34,7 @@
   "add-data-details": "Manually add your calculated values for required sub-sectors to the inventory.",
   "select-subsector": "Select sub-sector",
   "residential-buildings": "Residential buildings",
+  "residential-buildings-description": "(needs translation)",
```
lemilonkh, line 37: Let's get these values from either the data or product teams. I guess we'll need one for every sector, subsector and subcategory (there is a TODO text on those in the manual data entry form).

lemilonkh (approved):
Great work, this helps elevate the overall quality of the app a lot! ✨

cephaschapa:
Great thanks for the feedback 👍

PR #205, opened 2023-11-08
Make EDGAR importer more memory efficient

lgloege:
This PR improves how much memory the `citycelloverlapedgar_importer.py` script uses by removing the need to use dataframes and using generators instead of storing an entire query.


global-api/importer/edgar/utils.py
```diff
@@ -445,6 +445,14 @@ def all_locodes_and_geometries(session):
     return session.execute(query).fetchall()
 
 
+def all_locodes_and_geometries_generator(session):
+    """Generate locode and geometry pairs from the database."""
+    query = text("""SELECT locode, geometry FROM osm;""")
+    result = session.execute(query)
```
lemilonkh, line 451: I think this variable would still contain all of that data in the DB, right? Might be worth looking into cursors or paginating this query (with limit and offset), so you only load a part of the data at once and discard it (set variable null/ overwrite) before loading the next part.

Only if you are having memory issues here though, as you're impacting the overall speed otherwise.

lemilonkh (approved):

global-api/importer/edgar/utils.py
lgloege: I was using  `sys.getsizeof()`  to check the size of variables. The size of this generator is very small so I presume its not storing the entire query in memory. Having said that, I don't completely understand how generators "yield" data.

PR #223, opened 2023-11-15
Catalogue Sync Script

lemilonkh:
Run it with `npm run sync-catalogue`

Contains:

- Data Catalogue sync script
- k8s config for the data catalogue sync cronjob as well as running it before deployment currently
- Run seeders on deployment as a k8s job
- Fix GPC seeder to be idempotent by writing a custom bulkUpsert method


app/scripts/catalogue-sync.ts
```diff
@@ -0,0 +1,214 @@
...
+  if (!lastUpdate?.last_update) {
+    throw new Error(
+      "Failed to query last catalogue update with error " +
+        lastUpdateResponse.status +
+        " " +
+        lastUpdateResponse.statusText,
+    );
+  }
+
+  console.log(
```
cephaschapa, line 83: Are we using this for logging?

cephaschapa (approved):
Looks solid! left one comment too just out of curiosity.

app/scripts/catalogue-sync.ts
lemilonkh: I think it's helpful to know the timestamps for debugging purposes in case the catalogue doesn't update properly on the server.

PR #225, opened 2023-11-15
Update locodes in climatetrace asset table

lgloege:
This PR uses nominatim and openclimate to get the locode for points in the asset table. This creates a csv file with `id` and `locode` that is used to update records in the table.  I prioritized BASIC GPC sectors.


evanp:
As mentioned on Slack, we have a data file mapping locodes to OSM IDs which will probably give a better hit rate than using OpenClimate's search.

lgloege:
@evanp and @lemilonkh , was able to match a little 703 points using the osmid as a lookup. This is more than using openClimate, but not as great as I was expecting. This is ready for review. There are notes in the README how to update the table.

PR #226, opened 2023-11-17
Fix/qa fixes

cephaschapa:
changes
- Adds changes for QA feedback
- Adds changes to persist selected language when routing to a different page
- Adds fallback view for cities without boundaries
- Changes Zoom value to show multi-polygon cities on the map
- Replaced global api url variable with a conditional bases on the environment
- Added translated (en, es, de ) descriptions for subsectors (product has to verify)


app/src/app/[lng]/onboarding/done/[cityId]/[year]/page.tsx
```diff
@@ -9,16 +9,14 @@ import NextLink from "next/link";
...
-  params: { lng: string };
+  params: { lng: string; year: number };
 }) {
   const data = useAppSelector((state) => state.openClimateCity.city);
   const { t } = useTranslation(lng, "onboarding");
 
-  // TODO load these from the API
-  const cityName = "Ciudad Autónoma de Buenos Aires";
-  const year = "2023";
+  console.log(data);
```
Comment (lemilonkh) [inline app/src/app/[lng]/onboarding/done/[cityId]/[year]/page.tsx:19]: Can be removed

app/src/app/[lng]/onboarding/setup/page.tsx
```diff
@@ -388,7 +388,9 @@ export default function OnboardingSetup({
     goToNext();
   };
 
-  const { data: cityData } = useGetOCCityDataQuery(data.locode);
+  const { data: cityData } = useGetOCCityDataQuery(data.locode, {
+    skip: data.locode.length ? false : true,
```
Comment (lemilonkh) [inline app/src/app/[lng]/onboarding/setup/page.tsx:392]: No need to change this but fun fact - an empty string is false-y in JS. So just data.locode ? ... achieves the same thing.

app/src/app/[lng]/onboarding/setup/page.tsx
```diff
@@ -495,7 +497,7 @@ export default function OnboardingSetup({
             />
           )}
         </div>
-        <div className="bg-white w-full fixed bottom-0 left-0 border-t-4 border-brand flex flex-row py-8 px-8 drop-shadow-2xl hover:drop-shadow-4xl transition-all">
+        <div className="bg-white w-full fixed z-[9999] bottom-0 left-0 border-t-4 border-brand flex flex-row py-8 px-8 drop-shadow-2xl hover:drop-shadow-4xl transition-all">
```
Comment (lemilonkh) [inline app/src/app/[lng]/onboarding/setup/page.tsx:500]: Not sure if the z-index needs to be this high but good fix for now. I think the map is around z-index 500 or so, but it has multiple layers.

app/src/app/providers.tsx
```diff
@@ -27,12 +27,12 @@ const poppins = localFont({
     },
   ],
 });
-const openSans = localFont({
+const poppins = localFont({
```
lemilonkh, line 30: That makes a lot of sense, thanks for fixing the mixup :sweat_smile: I was wondering why it was always looking so different from the designs but didn't put 1 and 1 together.

app/src/components/CityMap.tsx
```diff
@@ -5,9 +5,16 @@ import { geoJSONBoundingBox } from "@/util/geojson";
 import { Box, Center, Spinner } from "@chakra-ui/react";
 import type { GeoJsonObject } from "geojson";
 import { LatLngBoundsLiteral, LatLngExpression } from "leaflet";
+import L from "leaflet";
```
lemilonkh, line 8:
```suggestion
import type L from "leaflet";
```

This means it's only used for the TS compiler but not included in the final page code.

app/src/components/navigation-bar.tsx
```diff
@@ -38,7 +39,12 @@ export function NavigationBar({
 }) {
   const { t } = useTranslation(lng, "navigation");
   const onChangeLanguage = (language: string) => {
-    i18next.changeLanguage(language);
+    Cookies.set("i18next", language);
+    const cookieLanguage = Cookies.get("i18next");
```
lemilonkh, line 43: Why are you reading the cookie immediately after writing it? This means it will always be `language`, so no need to read the cookie here imo.

app/src/i18n/locales/de/data.json
```diff
@@ -0,0 +1,104 @@
+{
```
lemilonkh, line 1: This file (I think it's from the german language PR made by the outside contributor?) should be updated to reflect the translation feedback I left on that PR.

lemilonkh: It's mostly correct though so it's a good start.

app/src/i18n/locales/de/data.json
```diff
@@ -0,0 +1,104 @@
...
+  "emission-factors-values": "Werte der Emissionsfaktoren",
+  "co2-emission-factor": "CO2-Emissionsfaktor",
+  "n2o-emission-factor": "N2O-Emissionsfaktor",
+  "ch4-emission-factor": "CH4-Emissionsfaktor",
+  "optional": "Optional",
+  "emissions-factor-details": "Mindestens ein Emissionsfaktor ist erforderlich. Leere Werte werden als 0 berechnet und müssen begründet werden.",
+  "source-reference": "Quellenreferenz",
+  "source-reference-required": "Fügen Sie eine Beschreibung der Referenzquelle und begründete Emissionsfaktoren, falls zutreffend, hinzu",
+  "source-reference-placeholder": "Geben Sie ausführlich an, aus welcher Quelle der eingegebene Wert stammt",
+  "calculations-details": "Alle Berechnungen berücksichtigen einen GWP-Wert von <1>{{co2Value}} für CO2 und {{ch4Value}} für CH4</1>.",
```
lemilonkh, line 93: This is outdated, there is now a link here instead of two separate numbers that are being passed in (see value in English data translation file).

lemilonkh (approved):

app/src/i18n/locales/de/data.json
lemilonkh: Might also need to be updated on the spanish version of the translation.

PR #227, opened 2023-11-17
Upsert records into datasource 

lgloege:
This PR add `third_party` to the `source_type` field and now upserts records.

### DIFF
--- a/global-api/importer/datasource_seeder/utils.py
+++ b/global-api/importer/datasource_seeder/utils.py
@@ -4,6 +4,28 @@ from datetime import datetime
 from pathlib import Path
 import uuid
 
+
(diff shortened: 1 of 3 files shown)


global-api/importer/datasource_seeder/utils.py
```diff
@@ -4,6 +4,28 @@
 from pathlib import Path
 import uuid
 
+
+def upsert_record(engine, table, pkey, record):
```
lemilonkh, line 8: This might be an easier and more performant option for upsert (especially if you're upserting large amounts of data, since it's a single query):

https://stackoverflow.com/a/44395983

lemilonkh (approved):

PR #243, opened 2023-12-06
Fix saving on add data screen

lemilonkh:
- Adds unique constraint (and index) on SubSectorValue table for `(inventory_id, subsector_id)` to prevent multiple values being saved for a single inventory
- Fixes race condition for API requests during add data form saving by saving the SubCategoryValue instances one after another (for now, there can be a batch endpoint later)
- This deletes all current SubSectorValue entries to prevent issues during the migration. Do I need to delete the SubCategoryValues as well so we don't have left-over data in the DB? Or is it fine since these tables will be removed once we refactor our DB structure?
- Removes manual progress increment when connecting data sources, since inventory progress is now automatically reloaded

### DIFF
--- a/app/migrations/20231206143948-SubSectorValue-unique.cjs
+++ b/app/migrations/20231206143948-SubSectorValue-unique.cjs
@@ -1,0 +1,30 @@
+"use strict";
+
+/** @type {import('sequelize-cli').Migration} */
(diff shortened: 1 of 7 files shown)


cephaschapa (approved):
great work

PR #244, opened 2023-12-08
Pre-check data sources

lemilonkh:
Adds data source pre-checking, so seeing if the global API returns values for them. Currently has no caching, so needs to be re-done after a page reload (but is still reasonably fast to do).

Adds a reload button for the data sources.

Also contains a bunch of UI fixes like the data source card not filling the whole width if there is only one or the grey background color not going to the bottom of the page.

## Screenshots:  
image

image


app/scripts/catalogue-sync.ts
```diff
@@ -77,9 +80,7 @@ async function syncDataCatalogue() {
   // convert to unix timestamp in ms
   const lastUpdate = lastUpdateData.last_update * 1000;
 
-  console.log(
-    `Last update: DB - ${previousUpdate}, API - ${lastUpdate}`,
-  );
+  console.log(`Last update: DB - ${previousUpdate}, API - ${lastUpdate}`);
   if (lastUpdate <= previousUpdate) {
     console.warn("Already on the newest data catalogue version, exiting.");
```
cephaschapa, line 85: Are we using this for logging also?

app/src/app/[lng]/data/[step]/page.tsx
```diff
@@ -54,6 +55,49 @@ function getMailURI(locode?: string, sector?: string, year?: number): string {
   return `mailto://info@openearth.org,greta@openearth.org?subject=Missing third party data sources&body=City: ${locode}%0ASector: ${sector}%0AYear: ${year}`;
```
Comment (cephaschapa) [inline app/src/app/[lng]/data/[step]/page.tsx:55]: Could be also nice to have these emails in our env variables in future.

app/src/app/[lng]/data/[step]/page.tsx
```diff
@@ -54,6 +55,49 @@ function getMailURI(locode?: string, sector?: string, year?: number): string {
...
+        loadingText={t("searching")}
+        onClick={onSearchClicked}
+        mb={2}
+        px={6}
+        h={16}
+        py={4}
+      >
+        {t("search-available-datasets")}
+      </Button>
+      <Text color="content.tertiary" align="center" size="sm">
```
Comment (cephaschapa) [inline app/src/app/[lng]/data/[step]/page.tsx:94]: I think this may also require, letterSpacing, lineHeight and fontWeight

app/src/app/[lng]/data/[step]/page.tsx
```diff
@@ -458,32 +512,54 @@ export default function AddDataSteps({
...
+          align="center"
+          verticalAlign="center"
+          justify="space-between"
+          mb={12}
+        >
+          <Stack>
+            <Heading size="lg" mb={2}>
+              {t("check-data-heading")}
+            </Heading>
+            <Text color="content.tertiary">{t("check-data-details")}</Text>
```
Comment (cephaschapa) [inline app/src/app/[lng]/data/[step]/page.tsx:525]: This one as well.

cephaschapa (commented):
Looks neat, just left a few comments mostly styling.

app/scripts/catalogue-sync.ts
lemilonkh: Yes, otherwise the program would just terminate without letting you know why. So I think it's fine to keep this.

app/src/app/[lng]/data/[step]/page.tsx
Comment (lemilonkh) [reply app/src/app/[lng]/data/[step]/page.tsx:55]: That's a good point :+1:
I'll add a variable for that.

Comment (lemilonkh) [reply app/src/app/[lng]/data/[step]/page.tsx:55]: There is now `process.env.SUPPORT_EMAILS`, but I added the current emails as a default value.

PR #250, opened 2023-12-11
Feature/dashboard contextual info

cephaschapa:
- Adds Population routes (GET, POST)
- Adds two endpoints to RTK `getCityPopulation` and `addCityPopulation`
- Adds and await method to post population data after City creation
- Adds a util function to shorted population values to B, M, Ks (prevents overflowing but could be revisited)
- State behavior


app/src/app/api/v0/city/[city]/route.ts
```diff
@@ -19,9 +19,10 @@ export const GET = apiHandler(
         {
           model: db.models.User,
           as: "users",
-          // where: {
-          //   userId: session.user.id,
-          // },
+        },
+        {
+          model: db.models.Population,
+          as: "populations",
```
Comment (lemilonkh) [inline app/src/app/api/v0/city/[city]/route.ts:25]: Perhaps keep the comment here.
It was commented out to prevent errors due to access control during demos etc., but we should look into adding the access control back in once there is a way to invite others to your inventory (or that multiple people can have an inventory for the same city).

app/src/app/[lng]/page.tsx
```diff
@@ -330,16 +366,28 @@ export default function Home({ params: { lng } }: { params: { lng: string } }) {
...
+                        ) : (
+                          <Text
+                            fontFamily="heading"
+                            color="base.light"
+                            fontSize="headline.sm"
+                            fontWeight="semibold"
+                            lineHeight="32"
+                          >
+                            {city?.area}
+                            <span className="text-[16px]">km2</span>
```
Comment (lemilonkh) [inline app/src/app/[lng]/page.tsx:388]:
```suggestion
                            <span className="text-[16px]">km<sup>2</sup></span>
```

Could be nicer to have the unit properly formatted here.

app/src/app/[lng]/page.tsx
```diff
@@ -280,7 +289,14 @@ export default function Home({ params: { lng } }: { params: { lng: string } }) {
                           fontWeight="semibold"
                           lineHeight="32"
                         >
-                          700<span className="text-[16px]">Mtco2e</span>
+                          {inventory?.totalEmissions ? (
```
Comment (lemilonkh) [inline app/src/app/[lng]/page.tsx:292]: Not sure if we should display "In progress" while the inventory progress is not yet 100%? Otherwise the number is sort of incomplete...

Maybe fine for now.

app/src/app/[lng]/page.tsx
```diff
@@ -280,7 +289,14 @@ export default function Home({ params: { lng } }: { params: { lng: string } }) {
...
+                            <>
+                              {inventory.totalEmissions}{" "}
+                              <span className="text-[16px]">Mtco2e</span>
+                            </>
+                          ) : (
+                            "In progress"
```
Comment (lemilonkh) [inline app/src/app/[lng]/page.tsx:298]: Please add a translation string for this and use `t("in-progress")` here

app/src/app/[lng]/onboarding/setup/page.tsx
```diff
@@ -434,7 +470,16 @@ export default function OnboardingSetup({
         area: ocCityData?.area!,
         region: ocCityData?.region!,
         country: ocCityData?.country!,
-      }).unwrap();
+      })
+        .unwrap()
+        .then(async (res: any) => {
+          await addCityPopulation({
+            cityId: res.data.cityId,
+            locode: res.data.locode!,
```
Comment (lemilonkh) [inline app/src/app/[lng]/onboarding/setup/page.tsx:478]: You could just use `data.locode` here, since it didn't change since then, so you don't have to "unwrap" it with the ! operator.

app/src/app/[lng]/onboarding/setup/page.tsx
```diff
@@ -408,6 +434,16 @@ export default function OnboardingSetup({
...
   useEffect(() => {
     if (cityData) {
+      const population = cityData?.data.population.filter(
+        (item: any) => item.year === data.year,
+      );
+      const populationObject = {
+        year: population[0]?.year,
+        population: population[0]?.population,
+        datasourceId: population[0]?.datasource_id,
+      };
```
Comment (lemilonkh) [inline app/src/app/[lng]/onboarding/setup/page.tsx:444]: Here you could use `.find` instead of `.filter`, so you don't have to index into the array afterwards and it stops after it has found the first matching entry.

app/src/app/[lng]/onboarding/setup/page.tsx
```diff
@@ -309,7 +315,14 @@ function ConfirmStep({
               <Icon as={MdOutlinePeopleAlt} boxSize={6} mt={1} mr={2} />
               <Box>
                 <Text fontSize="xl">
-                  3,978.9M
+                  {population ? (
+                    <>
+                      {shortenNumber(population)}
+                      {getShortenNumberUnit(population)}
```
Comment (lemilonkh) [inline app/src/app/[lng]/onboarding/setup/page.tsx:321]: Nit: I would prefer `getShortenedNumberUnit` for the method name, or making it one method (that the `shortenNumber` method returns a formatted string instead of just the number part).

Comment (lemilonkh) [reply app/src/app/[lng]/onboarding/setup/page.tsx:321]: I see that you're using both methods separately further down, so never mind that second suggestion :sweat_smile:

app/src/app/api/v0/city/[city]/population/route.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export const POST = apiHandler(
+  async (
+    req: Request,
+    context: { session?: Session; params: Record<string, string> },
+  ) => {
+    const body = createPopulationRequest.parse(await req.json());
+    if (!context.session) {
+      throw new createHttpError.Unauthorized("Unauthorized");
+    }
```
Comment (lemilonkh) [inline app/src/app/api/v0/city/[city]/population/route.ts:17]: Later on we will have to do access control here as well, so a user can't modify the Population entries of a city that he doesn't "own"/ have access to in our database.

app/src/util/helpers.ts
```diff
@@ -22,3 +22,27 @@ export function getCurrentVersion(): string {
...
+  } else if (number >= 1e6) {
+    return (number / 1e6).toFixed(1); // Million
+  } else if (number >= 1e3) {
+    return (number / 1e3).toFixed(1); // Thousand
+  } else {
+    return number.toString();
+  }
+}
+
+export function getShortenNumberUnit(number: number): string {
```
lemilonkh, line 38: Nit: I would prefer `getShortenedNumberUnit` here, since it feels off to have a verb in the middle here.

app/src/util/helpers.ts
```diff
@@ -22,3 +22,27 @@ export function getCurrentVersion(): string {
...
+  if (number >= 1e9) {
+    return "B"; // Billion
+  } else if (number >= 1e6) {
+    return "M"; // Million
+  } else if (number >= 1e3) {
+    return "K"; // Thousand
+  } else {
+    return number.toString();
```
lemilonkh, line 46:
```suggestion
    return "";
```

I think in this case you will print the number twice otherwise.

app/src/util/validation.ts
```diff
@@ -111,3 +115,12 @@ export const createUserRequest = z.object({
 });
 
 export type CreateUserRequest = z.infer<typeof createUserRequest>;
+
+export const createPopulationRequest = z.object({
+  cityId: z.string().uuid(),
+  population: z.number(),
+  year: z.number(),
+  datasourceId: z.string().optional(),
```
lemilonkh, line 123: Not sure if you should pass in a datasourceId here. Maybe rather another identifier that is used to look up the datasourceId in our database? Like "open_climate_api" or something, that is then checked against our DB.

lemilonkh (changes requested):

PR #255, opened 2023-12-14
Allow saving single scopes, fix scope loading, confirm dialog

lemilonkh:
- Allow saving the SubsectorDrawer form if not all scopes have been filled
- Fixes scope loading so they don't all share the same values
- Adds confirmation dialog if you have unsaved data and try to close the SubsectorDrawer:

image


cephaschapa (approved):
Looks neat :)

PR #263, opened 2023-12-18
Feature/connect settings and profile

cephaschapa:
Changes
- Connects account details form to update mutation route
- Connects users tab to users mutation and query routes (CRUD)
- Connects cities tab to cities mutation and query routes (CRUD)
- Adds bulk deletion of users when selected
- Adds toasts notification
- Adds translation strings to i18n files
- Adds request password confirmation route
- Adds modification to users and city routes
- Code refactoring and cleanup


app/src/app/[lng]/settings/page.tsx
```diff
@@ -51,6 +48,26 @@ export default function Settings({
...
+        userId: userInfo?.userId!,
+      },
+      { skip: !userInfo },
+    );
+
+  const { data: cityUsers } = api.useGetCityUsersQuery(
+    { locode: userInfo?.defaultCityLocode! },
+    {
+      skip: !userInfo,
+      pollingInterval: 3000,
```
Comment (lemilonkh) [inline app/src/app/[lng]/settings/page.tsx:67]: Should we really poll this API route, or is it sufficient to use `invalidateTags` in the mutations to force a reload of the data?

app/src/app/api/v0/city/[city]/user/route.ts
```diff
@@ -11,14 +11,32 @@ export const POST = apiHandler(async (_req: NextRequest, { params }) => {
...
   }
 
-  user.addCity(cityId);
+  user.addCity(city.cityId);
 
   return NextResponse.json({ data: user });
 });
+
+export const GET = apiHandler(async (_req: NextRequest, { params }) => {
+  const users = await db.models.User.findAll({
```
Comment (lemilonkh) [inline app/src/app/api/v0/city/[city]/user/route.ts:26]: This API route doesn't use params.user, so why is it in this file instead of the file one level above? Also, both GET functions seem to do the same thing?

app/src/app/[lng]/settings/page.tsx
```diff
@@ -51,6 +48,26 @@ export default function Settings({
...
+    api.useGetUserInfoQuery();
+
+  const { data: currentUserData, isLoading: isUserDataLoading } =
+    api.useGetUserQuery(
+      {
+        locode: userInfo?.defaultCityLocode!,
+        userId: userInfo?.userId!,
+      },
+      { skip: !userInfo },
+    );
```
Comment (lemilonkh) [inline app/src/app/[lng]/settings/page.tsx:61]: What is this query for? Can't you get the same data from the other two requests?

We should limit the amount of API requests to the absolute minimum for a snappy user experience.

app/src/app/api/v0/auth/requestpassword/route.ts
```diff
@@ -0,0 +1,36 @@
...
+import { NextRequest, NextResponse } from "next/server";
+import { z } from "zod";
+
+const requestPassword = z.object({
+  password: z.string().min(4).regex(passwordRegex),
+});
+
+export const POST = apiHandler(
+  async (
+    _req: NextRequest,
```
lemilonkh, line 17: If you are using a function argument, name it without the _
```suggestion
    req: NextRequest,
```

Initial _ is reserved to mark function arguments as unused.

app/src/app/api/v0/auth/requestpassword/route.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    const userId = context.session?.user.id;
+    const user = await db.models.User.findOne({ where: { userId } });
+
+    if (!user) {
+      throw createHttpError.NotFound("User not found!");
+    }
+
+    const password = body.password;
+    const comparePassword = await bcrypt.compare(password, user.passwordHash!);
```
lemilonkh, line 30: I'm not sure if it's a good idea to expose this via an API endpoint. This could very easily lead to brute forcing attacks. For which functionality is this route necessary?

app/src/app/api/v0/city/[city]/route.ts
```diff
@@ -44,15 +44,6 @@ export const DELETE = apiHandler(
...
       where: { locode: params.city },
-      include: [
-        {
-          model: db.models.User,
-          as: "users",
-          // where: {
-          //   userId: session.user.id,
-          // },
-        },
-      ],
```
Comment (lemilonkh) [inline app/src/app/api/v0/city/[city]/route.ts:55]: Why is the access control etc. removed here? Aren't the users necessary?

app/src/app/api/v0/city/[city]/user/[user]/route.ts
```diff
@@ -33,7 +30,7 @@ export const GET = apiHandler(async (_req: NextRequest, { params }) => {
 });
 
 export const PATCH = apiHandler(async (_req: NextRequest, { params }) => {
-  const body = createUserRequest.parse(await _req.json());
+  const body = await _req.json();
```
Comment (lemilonkh) [inline app/src/app/api/v0/city/[city]/user/[user]/route.ts:33]: Why remove the validation?

app/src/app/api/v0/city/[city]/user/[user]/route.ts
```diff
@@ -18,9 +18,6 @@ export const GET = apiHandler(async (_req: NextRequest, { params }) => {
       {
         model: db.models.City,
         as: "cities",
-        where: {
-          locode: locode,
-        },
```
Comment (lemilonkh) [inline app/src/app/api/v0/city/[city]/user/[user]/route.ts:23]: Why remove this?

app/src/components/Modals/delete-city-modal.tsx
```diff
@@ -44,6 +49,27 @@ const DeleteCityModal: FC<DeleteCityModalProps> = ({
     setValue,
   } = useForm<{ password: string }>();
   const { t } = useTranslation(lng, "dashboard");
+  const [requestPasswordConfirm] = api.useRequestConfirmPasswordMutation();
```
lemilonkh, line 52: This is not the best approach to implement a password confirmation.
Usually you would use the same logic as a login, and require a recently created JWT token here (otherwise show a login form (which can be just a password prompt as you have here) or redirect the user).
Let's talk about this some more in a meeting.

app/src/components/Modals/delete-file-modal.tsx
```diff
@@ -100,7 +101,7 @@ const DeleteFileModal: FC<DeleteFileModalProps> = ({
                 fontWeight="semibold"
                 fontSize="button.md"
                 type="button"
-                onClick={() => alert(userData.id)}
+                onClick={() => alert(userData.userId)}
```
lemilonkh, line 104: Let's remove this alert and either do a log statement (using `logger`) that it's unimplemented or implement the account deletion.

app/src/components/Modals/delete-inventory-modal.tsx
```diff
@@ -157,7 +158,7 @@ const DeleteInventoryModal: FC<DeleteInventoryModalProps> = ({
                       fontWeight="semibold"
                       fontSize="button.md"
                       type="button"
-                      onClick={() => alert(userData.id)}
+                      onClick={() => alert(userData.userId)}
```
lemilonkh, line 161: Remove alert and either add logger call or actual logic for deletion.

app/src/components/Modals/update-user-modal.tsx
```diff
@@ -42,10 +48,64 @@ const UpdateUserModal: FC<UpdateUserModalProps> = ({
...
+  const [setUserData] = api.useSetUserDataMutation();
+
+  const toast = useToast();
+
+  const [inputValue, setInputValue] = useState<string>("");
+
+  const onSubmit: SubmitHandler<ProfileInputs> = async (data) => {
+    // TODO
+    // Submit data via the api
+    await setUserData({
```
lemilonkh, line 60: I think you can remove this TODO comment as you're submitting the data now?

app/src/components/Tabs/my-profile-tab.tsx
```diff
@@ -64,33 +67,92 @@ import DeleteUserModal from "@/components/Modals/delete-user-modal";
...
+  }, [setValue, session, status, userInfo]);
+
+  const cities: any = userInfo && userInfo.cities;
 
+  const [setCurrentUserData] = useSetCurrentUserDataMutation();
+  const toast = useToast();
   const onSubmit: SubmitHandler<ProfileInputs> = async (data) => {
     // TODO
     // Submit data via the api
+    await setCurrentUserData({
```
lemilonkh, line 114: Remove TODO comment

app/src/components/Tabs/my-profile-tab.tsx
```diff
@@ -111,38 +173,31 @@ const MyProfileTab: FC<MyProfileTabProps> = ({ session, status, t, lng }) => {
...
-    setFilteredUsers(result);
-  }, [role, searchTerm]);
+      setFilteredUsers(result);
+    }
+  }, [role, searchTerm, cityUsers]);
 
   useEffect(() => {
-    const selectedUserByRole = filteredUsers.filter((users) =>
-      users.role.toLocaleLowerCase().includes(role.toLocaleLowerCase()),
+    const selectedUserByRole = filteredUsers.filter(
```
lemilonkh, line 198: You're returning multiple users here.
```suggestion
    const selectedUsersByRole = filteredUsers.filter(
```

lemilonkh (changes requested):

app/src/app/api/v0/city/[city]/user/[user]/route.ts
Comment (cephaschapa) [reply app/src/app/api/v0/city/[city]/user/[user]/route.ts:23]: didn't require this either after switching to the other suggested GET route

app/src/app/api/v0/auth/verify/route.ts
```diff
@@ -0,0 +1,72 @@
...
+    context: { session?: Session; params: Record<string, string> },
+  ) => {
+    const email = context.session?.user.email;
+    const user = await db.models.User.findOne({ where: { email } });
+
+    if (!user) {
+      throw createHttpError.NotFound("User not found!");
+    }
+
+    if (!process.env.RESET_TOKEN_SECRET) {
```
lemilonkh, line 28: Add this secret (with placeholder value) to:
- env.example
- github action (tests)

app/src/services/api.ts
```diff
@@ -267,21 +266,46 @@ export const api = createApi({
...
+      { password: string; token: string }
+    >({
+      query: ({ password, token }) => ({
+        url: `/auth/verify`,
         method: "POST",
-        body: { password },
+        body: { password, token },
       }),
     }),
+    getCities: builder.query<CityAttributes, null>({
```
lemilonkh, line 288: void instead of null

app/src/app/[lng]/settings/page.tsx
```diff
@@ -51,20 +48,17 @@ export default function Settings({
...
-  const { data: currentUserData, isLoading: isUserDataLoading } =
-    api.useGetUserQuery(
-      {
-        locode: userInfo?.defaultCityLocode!,
-        userId: userInfo?.userId!,
-      },
-      { skip: !userInfo },
-    );
+  const { data: cities, isLoading: isCitiesLoading } = api.useGetCitiesQuery(
+    null,
```
Comment (lemilonkh) [inline app/src/app/[lng]/settings/page.tsx:52]: void

app/src/app/[lng]/settings/page.tsx
```diff
@@ -48,8 +48,10 @@ export default function Settings({
   const { data: userInfo, isLoading: isUserInfoLoading } =
     api.useGetUserInfoQuery();
 
+  let isVoid: void;
+
   const { data: cities, isLoading: isCitiesLoading } = api.useGetCitiesQuery(
-    null,
+    isVoid,
```
Comment (lemilonkh) [inline app/src/app/[lng]/settings/page.tsx:54]: You can just remove this line and the definition above, void in the redux route type make it so you don't have to pass anything here.

app/src/app/api/v0/auth/verify/route.ts
```diff
@@ -25,14 +25,14 @@ export const GET = apiHandler(
       throw createHttpError.NotFound("User not found!");
     }
 
-    if (!process.env.RESET_TOKEN_SECRET) {
+    if (!process.env.VEFIFICATION_TOKEN_SECRET) {
```
lemilonkh, line 28: Typo in all of these
```suggestion
    if (!process.env.VERIFICATION_TOKEN_SECRET) {
```

app/src/components/Modals/delete-city-modal.tsx
```diff
@@ -50,9 +50,10 @@ const DeleteCityModal: FC<DeleteCityModalProps> = ({
     formState: { errors, isSubmitting },
     setValue,
   } = useForm<{ password: string }>();
+  let isVoid: void;
   const { t } = useTranslation(lng, "dashboard");
   const [requestPasswordConfirm] = api.useRequestVerificationMutation();
-  const { data: token } = api.useGetVerifcationTokenQuery(null, {
+  const { data: token } = api.useGetVerifcationTokenQuery(isVoid, {
```
lemilonkh, line 56: Just remove this instead of passing isVoid in here (and remove that definition).

PR #277, opened 2024-01-11
Flat inventory refactoring

lemilonkh:
- Replace SectorValue, SubSectorValue, SubCategoryValue tree with single table - InventoryValue
- Add new table GasValue to store either direct emissions and emissions factor reference or just the emissions factor reference, to be used with activityAmount in InventoryValue
- Add migration "FlatInventory"
- Allow storing user emissions factors in the EmissionsFactor table marked with the new `inventoryId` property
- Rebuild progress route and update InventoryValue route
- Update tests
- Switch to ES2020 target in tsconfig.json, so BigInt literals (`1n`) are supported, making calculations easier to read
- Contains helper code from `https://gist.github.com/wchargin/93062064e1ec0e3383b91b921da86048` to allow multiplying a BigInt with a float without truncating it. This code was merged in [this PR](https://github.com/sourcecred/sourcecred/pull/1715#issuecomment-603354146) into a MIT/Apache licensed codebase, but doesn't directly have a license attached to it. @evanp is this an issue? If so, we would likely need to include the [bignumber.js package](https://www.npmjs.com/package/bignumber.js) from npm.

**TODO**:
- [x] Test InventoryValue PATCH route
- [x] Adjust add data page to use new API routes and data schemas
- [x] Restore previous gas data and assign SubSector and Sector to InventoryValue


app/src/app/api/v0/inventory/[inventory]/value/[subcategory]/route.ts
Comment (cephaschapa) [inline app/src/app/api/v0/inventory/[inventory]/value/[subcategory]/route.ts:1]: Are we doing access control for any inventory requests?

Comment (lemilonkh) [reply app/src/app/api/v0/inventory/[inventory]/value/[subcategory]/route.ts:1]: Currently not. To introduce access control again I think we need to do:
- make inventories unique by their ID instead of only allowing one inventory per city and year (and adjust the API routes for this so they take the inventory UUID instead of the locode)
- create an invite system so you can invite other users to your inventory

Also we probably need to have a many to many relationship from Inventory to User to reflect this access control model properly.

Comment (cephaschapa) [reply app/src/app/api/v0/inventory/[inventory]/value/[subcategory]/route.ts:1]: That's a lot :). Okay we can do it later gradually then.

cephaschapa (approved):
Had only one comment regarding access control which as you explained requires a lot changes, we could set up jira ticks (as tech debt), other than that this looks great!!!

app/migrations/20240104144355-FlatInventory.cjs
```diff
@@ -0,0 +1,344 @@
...
+          co2eq_years: {
+            type: Sequelize.INTEGER,
+            allowNull: true,
+          },
+        },
+        { transaction },
+      );
+
+      /// GasValue ///
+      await queryInterface.createTable(
```
evanp, line 180: I think I'd really like to see existing data copied over.

app/migrations/20240104144355-FlatInventory.cjs
```diff
@@ -0,0 +1,344 @@
...
+          field: "id",
+        },
+        onDelete: "SET NULL",
+        onUpdate: "SET NULL",
+        transaction,
+      });
+    });
+  },
+
+  async down(queryInterface) {
```
evanp, line 235: Thanks for doing this. I think we're losing data, but that's probably OK.

app/migrations/20240104144355-FlatInventory.cjs
```diff
@@ -0,0 +1,344 @@
...
+      );
+      await queryInterface.dropTable("SubSectorValue", { transaction });
+      await queryInterface.dropTable("SectorValue", { transaction });
+
+      /// EmissionsFactor ///
+      await renameColumns(
+        "EmissionsFactor",
+        {
+          emissions_factor_id: "id",
+          emissions_factor: "emissions_per_activity",
```
evanp, line 133: Hmm. OK. I think "emissions factor" is the term of art here. Are you just changing it to keep it from conflicting with the name of the table?

app/seeders/util/util.cjs
```diff
@@ -26,4 +29,19 @@ async function bulkUpsert(
   }
 }
 
-module.exports = { bulkUpsert };
+async function parseFile(folder, filename) {
+  const records = [];
+  const parser = fs
+    .createReadStream(
+      `${__dirname}/../../seed-data/${folder}/${filename}.csv`,
```
evanp, line 36: This feels fragile!

app/src/models/DataSource.ts
```diff
@@ -704,69 +703,37 @@ export class DataSource
   getPublisher!: Sequelize.BelongsToGetAssociationMixin<Publisher>;
   setPublisher!: Sequelize.BelongsToSetAssociationMixin<Publisher, PublisherId>;
   createPublisher!: Sequelize.BelongsToCreateAssociationMixin<Publisher>;
-  // DataSource hasMany SubSectorValue via datasourceId
```
evanp, line 707: Wow, that cleans up a lot of code!

evanp (approved):
Overrall the data looks good. I'm a little concerned about the migration dropping data; if we can fix that before demo, great. Otherwise, @lemilonkh please merge this!

app/migrations/20240104144355-FlatInventory.cjs
lemilonkh: I changed it since that's what you decided to call it in the diagram we made. I can change it back of course in case emissions_factor is better.

app/migrations/20240104144355-FlatInventory.cjs
lemilonkh: This is only meant to be used in a dev setup really. If we ever need to call it manually on the server we can adjust the down method to preserve the data beforehand.

app/seeders/util/util.cjs
lemilonkh: `__dirname` is always the directory name of the current file. So unless util.cjs is moved, this works fine (at which point it can be adjusted). I don't really want to handle absolute file paths here, since they will be different on every machine. So this seems like the best approach to me.

PR #281, opened 2024-01-17
ON-1148 mendoza import

mfonsecaOEF:
This PR includes:
- extraction script 
- transformation script
for stationary energy for Mendoza cities


global-api/migrations/versions/88e4463fef06_locode_data_table.py
```diff
@@ -0,0 +1,50 @@
...
+# revision identifiers, used by Alembic.
+revision: str = '88e4463fef06'
+down_revision: Union[str, None] = 'cc8d00d4372e'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "city_locode",
```
evanp, line 24: Not a great table name! How about "citywide_emissions"?

global-api/routes/city_locode_endpoint_mendoza_stationary_energy.py
```diff
@@ -0,0 +1,50 @@
+from fastapi import APIRouter, HTTPException
+from sqlalchemy import text
+import pandas as pd
```
evanp, line 3: Booo! No Pandas in API endpoints!

global-api/routes/city_locode_endpoint_mendoza_stationary_energy.py
```diff
@@ -0,0 +1,50 @@
...
+
+@api_router.get("/source/{source_name}/city/{locode}/{year}/{GPC_refno}")
+def get_emissions_by_locode_and_year(source_name: str, locode: str, year: str, GPC_refno: str):
+
+    records = db_query(source_name, locode, year, GPC_refno)
+
+    if not records:
+        raise HTTPException(status_code=404, detail="No data available")
+
+    df = pd.DataFrame(records)
```
evanp, line 36: Come now! You get an array, convert it to a data frame, and then convert it back to an array with iterrows?

I think we can do this a little simpler.

evanp (changes requested):
I gave some feedback. I think these need to be fixed before we merge.

global-api/migrations/versions/88e4463fef06_locode_data_table.py
mfonsecaOEF: it sounds better. I'm going to change it.

global-api/routes/city_locode_endpoint_mendoza_stationary_energy.py
mfonsecaOEF: changed!

PR #282, opened 2024-01-19
Upload own data backend

cephaschapa:
changes
- Adds migration to create userFile relation
- Adds userFile model class
- Adds file to user association in init-models
- Adds POST route to create a user file
- Adds GET route to fetch all user files
- Adds GET route to fetch a user file
- Adds DELETE route to remove a user file
- Adds file download route to enable downloading files of varying file types
- Adds unit test
- Adds helper functions to accommodate file creation to be used for unit tests
- Minor comments


app/src/app/api/v0/user/[user]/file/route.ts
```diff
@@ -0,0 +1,107 @@
...
+import { createUserFileRequset } from "@/util/validation";
+import { randomUUID } from "crypto";
+import createHttpError from "http-errors";
+import { Session } from "next-auth";
+import { NextRequest, NextResponse } from "next/server";
+import { z } from "zod";
+
+// TODO: use these variables to configure file size and format
+const MAX_FILE_SIZE = 5000000;
+const ACCEPTED_FILE_FORMATS = []; // file formats types to be parsed and refined later
```
Comment (lemilonkh) [inline app/src/app/api/v0/user/[user]/file/route.ts:12]: Could be .xls, .csv, .json

app/src/app/api/v0/user/[user]/file/route.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    const user = await db.models.User.findOne({
+      attributes: ["userId"],
+      where: {
+        userId: userId,
+      },
+    });
+    if (!user) {
+      throw new createHttpError.NotFound("User not found");
+    }
```
Comment (lemilonkh) [inline app/src/app/api/v0/user/[user]/file/route.ts:32]: Same here, remove the query and move the access check into the UserFile query

app/src/app/api/v0/user/[user]/file/route.ts
```diff
@@ -0,0 +1,107 @@
...
+    const user = await db.models.User.findOne({
+      attributes: ["userId"],
+      where: {
+        userId: userId,
+      },
+    });
+
+    if (!user) {
+      throw new createHttpError.NotFound("User not allowed");
+    }
```
Comment (lemilonkh) [inline app/src/app/api/v0/user/[user]/file/route.ts:68]: Check in UserFile query

app/src/app/api/v0/user/[user]/file/route.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    const formData = await req.formData();
+    const file = formData.get("data") as unknown as File;
+    if (!file) throw new createHttpError.BadRequest("File not found");
+
+    const filename = file.name;
+
+    const fileType = file.name.slice(
+      ((filename.lastIndexOf(".") - 1) >>> 0) + 2,
```
Comment (lemilonkh) [inline app/src/app/api/v0/user/[user]/file/route.ts:77]: What is the unsigned right shift by 0 for?
I don't think it does anything in this case (but it might overflow into very large positive numbers when lastIndexOf returns -1 or 0.

app/src/app/api/v0/user/[user]/file/route.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    const fileData = {
+      userId: formData.get("userId"),
+      file_reference: formData.get("file_reference"),
+      url: formData.get("url"),
+      data: buffer,
+      file_type: fileType,
+      sector: formData.get("sector"),
+      status: formData.get("status"),
+      gpc_ref_no: formData.get("gpc_ref_no"),
```
Comment (lemilonkh) [inline app/src/app/api/v0/user/[user]/file/route.ts:91]: I think these would need to be camelCase instead of snake_case so that they are parsed correctly by sequelize? Since that's how they are defined in the UserFile model.

Comment (lemilonkh) [reply app/src/app/api/v0/user/[user]/file/route.ts:91]: Ok nvm, you defined most of them as snake_case in the model. I would strongly favor consistency here, since all of our models use camelCase for the properties, where they have an `field` property in the sequelize column definitions in the bottom that contains the snake_case column name (see the other DB models for reference).

app/src/models/UserFile.ts
```diff
@@ -0,0 +1,123 @@
...
+  userId?: string;
+  file_reference?: string;
+  data?: Buffer;
+  file_type?: string;
+  sector?: string;
+  status?: string;
+  url?: string;
+  gpc_ref_no?: string;
+  created?: Date;
+  lastUpdated?: Date;
```
lemilonkh, line 16: Change to use camelCase

app/src/models/UserFile.ts
```diff
@@ -0,0 +1,123 @@
...
+          allowNull: true,
+        },
+        status: {
+          type: DataTypes.STRING(255),
+          allowNull: true,
+        },
+        gpc_ref_no: {
+          type: DataTypes.STRING(255),
+          allowNull: true,
+        },
```
lemilonkh, line 104: Add `field: "file_reference"` and rename to fileReference (and similar for the others).

app/src/models/init-models.ts
```diff
@@ -649,6 +652,8 @@ export function initModels(sequelize: Sequelize) {
   });
   User.belongsTo(User, { as: "organization", foreignKey: "organizationId" });
   User.hasMany(User, { as: "users", foreignKey: "organizationId" });
+  User.hasOne(UserFile, { foreignKey: "userId", as: "user" });
```
lemilonkh, line 655: The user should be able to upload multiple files, so this should be `hasMany` instead of `hasOne`.

app/src/util/validation.ts
```diff
@@ -124,3 +124,18 @@ export const createPopulationRequest = z.object({
...
+// user file schema validation
+export const createUserFileRequset = z.object({
+  userId: z.string().uuid().optional(),
+  file_reference: z.string().optional(),
+  data: z.any(),
+  file_type: z.string().optional(),
+  sector: z.string().optional(),
+  url: z.string().url().optional(),
+  status: z.string().optional(),
+  gpc_ref_no: z.string().optional(),
```
lemilonkh, line 137: Use camelCase for these for consistency with the other API endpoints.

app/src/models/UserFile.ts
```diff
@@ -72,17 +72,17 @@ export class UserFile
             model: "User",
             key: "user_id",
           },
-          field: "user_id",
+          field: "userId",
```
lemilonkh, line 75: Field should be user_id

app/src/models/UserFile.ts
```diff
@@ -98,7 +98,7 @@ export class UserFile
           type: DataTypes.STRING(255),
           allowNull: true,
         },
-        gpc_ref_no: {
+        gpcRefNo: {
```
lemilonkh, line 101: `field: "gpc_ref_no",`

app/src/app/api/v0/user/[user]/file/route.ts
```diff
@@ -56,20 +46,11 @@ export const POST = apiHandler(
...
-      where: {
-        userId: userId,
-      },
-    });
-
-    if (!user) {
-      throw new createHttpError.NotFound("User not allowed");
-    }
-
     const formData = await req.formData();
```
Comment (lemilonkh) [inline app/src/app/api/v0/user/[user]/file/route.ts:49]: Should we validate formData using zod as well? You're not doing null checks below for some of the properties, so it might make sense and give better error messages than the internal server error that is the result when the user passes the wrong data types in some fields here.

app/src/app/api/v0/user/[user]/file/[file]/download-file/route.ts
```diff
@@ -34,13 +23,16 @@ export const GET = apiHandler(
...
+      throw new createHttpError.NotFound("file does not belong to this user");
+
     let body: Buffer | undefined;
     let headers: Record<string, string> | null = null;
 
     body = userFile.data;
     headers = {
-      "Content-Type": "application/vnd.ms-excel",
-      "Content-Disposition": `attachment; filename="${userFile.id}.${userFile.file_type}"`,
+      "Content-Type": `application/${userFile.fileType}`,
```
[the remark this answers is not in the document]
Comment (lemilonkh) [reply app/src/app/api/v0/user/[user]/file/[file]/download-file/route.ts]: You can use `fileEndingToMIMEType[userFile.fileType || "default"]`

```js
const fileEndingToMIMEType = {
  "csv": "text/csv",
  "json": "application/json",
  "xls": "application/vnd.ms-excel",
  "xlsx": "application/vnd.ms-excel",
  "default": "application/x-binary",
};
```

lemilonkh (approved):

PR #321, opened 2024-02-12
Fix/settings profile bug fixes

cephaschapa:
Changes
- Adds modal footer to all modals
- Moves the form modal button to modal footer
- Adds translation
- Adds tags to api service
- modifies email input

### DIFF
--- a/app/src/app/[lng]/auth/signup/page.tsx
+++ b/app/src/app/[lng]/auth/signup/page.tsx
@@ -116,12 +116,7 @@ export default function Signup({
             {errors.name && errors.name.message}
           </FormErrorMessage>
         </FormControl>
-        <EmailInput
-          disabled={false}
-          register={register}
-          error={errors.email}
-          t={t}
-        />
+        <EmailInput register={register} error={errors.email} t={t} />
         <PasswordInput register={register} error={errors.password} t={t}>
           <FormHelperText>
             <InfoOutlineIcon
--- a/app/src/app/[lng]/auth/signup/page.tsx
+++ b/app/src/app/[lng]/auth/signup/page.tsx
@@ -116,7 +116,12 @@ export default function Signup({
             {errors.name && errors.name.message}
           </FormErrorMessage>
         </FormControl>
-        <EmailInput register={register} error={errors.email} t={t} />
+        <EmailInput
+          disabled={false}
+          register={register}
+          error={errors.email}
+          t={t}
+        />
         <PasswordInput register={register} error={errors.password} t={t}>
           <FormHelperText>
             <InfoOutlineIcon
--- a/app/src/components/Modals/add-user-modal.tsx
+++ b/app/src/components/Modals/add-user-modal.tsx
@@ -26,14 +26,21 @@ import { MdCheckCircleOutline } from "react-icons/md";
 import { api } from "@/services/api";
 import { UserAttributes } from "@/models/User";
 import FormSelectOrganization from "../form-select-organization";
+import { TFunction } from "i18next";
 
 interface AddUserModalProps {
   isOpen: boolean;
   onClose: any;
   userInfo: UserAttributes;
+  t: TFunction;
 }
 
-const AddUserModal: FC<AddUserModalProps> = ({ isOpen, onClose, userInfo }) => {
+const AddUserModal: FC<AddUserModalProps> = ({
+  isOpen,
+  onClose,
+  userInfo,
+  t,
+}) => {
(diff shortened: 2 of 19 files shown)


app/src/app/[lng]/auth/signup/page.tsx
```diff
@@ -116,7 +116,12 @@ export default function Signup({
             {errors.name && errors.name.message}
           </FormErrorMessage>
         </FormControl>
-        <EmailInput register={register} error={errors.email} t={t} />
+        <EmailInput
+          disabled={false}
```
Comment (lemilonkh) [inline app/src/app/[lng]/auth/signup/page.tsx:120]: Why the disabled={false} attribute here? That should be the default?

app/src/components/Modals/add-user-modal.tsx
```diff
@@ -140,62 +148,79 @@ const AddUserModal: FC<AddUserModalProps> = ({ isOpen, onClose, userInfo }) => {
...
+            borderBottomWidth="1px"
+            borderStyle="solid"
+            borderColor="border.neutral"
           >
-            Add User
+            {t("add-user")}
           </ModalHeader>
           <ModalCloseButton marginTop="10px" />
-          <ModalBody>
+          <ModalBody p="24px" px="48px">
```
lemilonkh, line 161: I usually prefer using Chakra's unit system (which means numbers are 1/4 of the pixel value usually, similar to Tailwind).
So `p={6} px={12}` in this case. It's fine if you'd rather use pixels though.

app/src/components/Modals/add-user-modal.tsx
```diff
@@ -140,62 +148,79 @@ const AddUserModal: FC<AddUserModalProps> = ({ isOpen, onClose, userInfo }) => {
...
+            borderColor="border.neutral"
+            w="full"
+            display="flex"
+            alignItems="center"
+            p="24px"
+            justifyContent="center"
+          >
+            <Button
+              h="56px"
+              w="472px"
```
lemilonkh, line 208: Do these need to be hardcoded? That usually conflicts with people setting custom font sizes in their browser/ OS, which then breaks the layout instead of just scaling the button.

app/src/components/Modals/add-user-modal.tsx
```diff
@@ -140,62 +148,79 @@ const AddUserModal: FC<AddUserModalProps> = ({ isOpen, onClose, userInfo }) => {
...
+              paddingBottom="16px"
+              px="24px"
+              letterSpacing="widest"
+              textTransform="uppercase"
+              fontWeight="semibold"
+              fontSize="button.md"
+              type="submit"
+              onClick={handleSubmit(onSubmit)}
+              p={0}
+              m={0}
```
lemilonkh, line 219: You could set these instead of the h/w attributes, so the total button size depends on the user's font size.

app/src/i18n/locales/de/settings.json
```diff
@@ -49,5 +49,15 @@
     "delete-file": "Datei löschen",
     "status": "Status",
     "download-csv": "Als CSV herunterladen",
-    "delete-inventory": "Inventar löschen"
+    "delete-inventory": "Inventar löschen",
+    "delete-inventory-prompt": "Sind Sie sicher, dass Sie das Inventar dieser Stadt <2>dauerhaft löschen</2> möchten?",
+    "enter-password-info": "Geben Sie Ihr Passwort ein, um das Löschen der Städteinformationen zu bestätigen.",
+    "user-details-updated": "Benutzerdaten aktualisiert",
+    "something-went-wrong": "Etwas ist schief gelaufen",
+    "is-organization": "Ist eine Organisation",
```
lemilonkh, line 57: This will be removed from the user model in my upcoming PR btw.
So we can get rid of the checkbox for it (either here or in the future).

app/src/app/[lng]/auth/signup/page.tsx
Comment (lemilonkh) [reply app/src/app/[lng]/auth/signup/page.tsx:120]: I see, it's a custom attribute. You can make it an optional property in EmailInput and set the default value to false.

app/src/components/email-input.tsx
```diff
@@ -14,21 +14,25 @@ export default function EmailInput({
   t,
   name = t("email-address"),
   id = "email",
+  disabled,
```
lemilonkh, line 17: Set a default value of `false` here, so you don't have to specify it if the input shouldn't be disabled.
```suggestion
  disabled = false,
```

lemilonkh (approved):

PR #322, opened 2024-02-15
Implement access control in backend

lemilonkh:


app/src/app/[lng]/page.tsx
```diff
@@ -187,7 +185,7 @@ export default function Home({ params: { lng } }: { params: { lng: string } }) {
       2000,
       "semantic.info",
     );
-    fetch(`/api/v0/city/${locode}/inventory/${year}?format=csv`)
+    fetch(`/api/v0/inventory/${defaultInventoryId}?format=csv`)
```
Comment (cephaschapa) [inline app/src/app/[lng]/page.tsx:188]: should we consider moving this to RTK in future?

app/src/app/api/v0/city/[city]/user/route.ts
```diff
@@ -1,34 +1,36 @@
...
+import { NextResponse } from "next/server";
 import { randomUUID } from "node:crypto";
 
-export const POST = apiHandler(async (_req: NextRequest, { params }) => {
+export const POST = apiHandler(async (_req, { params, session }) => {
   const body = createUserRequest.parse(await _req.json());
+
+  const city = await UserService.findUserCity(params.city, session);
+
+  // TODO shouldn't the users sign up themselves? This will probably prevent signup
```
Comment (cephaschapa) [inline app/src/app/api/v0/city/[city]/user/route.ts:14]: In addition new users are only added on signup with invite code.

app/src/app/api/v0/inventory/[inventory]/progress/route.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  const romanTable: Record<string, number> = {
+    I: 1,
+    II: 2,
+    III: 3,
+    IV: 4,
+    V: 5,
+    VI: 6,
+    VII: 7,
+    "": 1337,
```
Comment (cephaschapa) [inline app/src/app/api/v0/inventory/[inventory]/progress/route.ts:22]: will this be applicable?
MCCCXXXVII: 1337

cephaschapa (approved):

app/src/app/api/v0/inventory/[inventory]/progress/route.ts
Comment (lemilonkh) [reply app/src/app/api/v0/inventory/[inventory]/progress/route.ts:22]: This is just a default case to prevent errors for empty strings (that can occur if there is no reference number for some reason). So I just have this case for an empty string so there is still a number and they can be sorted.

PR #323, opened 2024-02-15
Implement code coverage metrics

cephaschapa:
Changes
- Quick start to show test coverage in tabular format format
- adds gh workflow
- adds ci script


.github/workflows/coverage-report.yml
```diff
@@ -0,0 +1,14 @@
...
+on:
+  pull_request:
+    branches:
+      - develop
+jobs:
+  coverage:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: artiomtr/jest-coverage-report-action@v1.3
```
lemilonkh, line 11: I think we need to run our tests beforehand? So install dev dependencies, run the tests with npm run ci:test, then show the coverage report. Also not sure if we shouldn't rather use one of the lcov actions instead, since we're not using jest.

lemilonkh (approved):

PR #324, opened 2024-02-15
Carbon monitor

lgloege:
This PR adds processings and import script for Carbon Monitor and makes a new alembic migration


global-api/importer/carbon_monitor/process.py
```diff
@@ -0,0 +1,150 @@
...
+            "Korea": "Korea, the Republic of",
+            "Turkey": "Türkiye",
+            "Russia": "Russian Federation",
+        }
+    }
+
+    df = (
+        pd.read_csv(INPUT_FILE, parse_dates=["date"])
+        # 0. filter out rows with 0 emissions
+        .loc[lambda x: x['value (KtCO2 per day)']>0]
```
lemilonkh, line 53: Why are you filtering out the 0 values? Wouldn't these still be valid and good to have present, so the user doesn't have to research them themselves?

Or are these invalid

lemilonkh (approved):
Looks great 👍
I think it would be helpful to add an extraction script that downloads the file from the given URL, so we can integrate this data source into Airflow in the future and process it automatically.

This can be handled in a future PR though.

evanp:
@mfonsecaOEF @lgloege I don't think this `CarbonMonitor` table is necessary. We have a `citywide_emissions` table. Can you use that instead? Mau, can you help Luke with this?

evanp:
So, it looks like there are outstanding issues here. @lgloege can you resolve?

global-api/importer/carbon_monitor/process.py
evanp: I discussed with @lgloege and our consensus is that it probably makes sense to be conservative in what we import. Zeroes may be measured values, but they may also signal a lack of information.

evanp (commented):
Seems to be well-reviewed.

PR #335, opened 2024-02-20
Feature/invite system backend

cephaschapa:
Changes
- Adds user invitation route
- Adds email invitation template
- Adds user invite migration
- Adds user invite model
- integrates invite route into rtk and ui
- Adds req body data validation
- Modifies user route


app/src/app/api/v0/city/invite/[invite]/route.ts
```diff
@@ -5,61 +5,47 @@ import { Session } from "next-auth";
...
+    throw new createHttpError.NotFound("Not found");
+  }
 
-    const invite = await db.models.CityInvite.findOne({
-      where: {
-        id: params.invite,
-      },
-    });
+  const token = _req.nextUrl.searchParams.get("token");
+  const email = _req.nextUrl.searchParams.get("email");
```
Comment (lemilonkh) [inline app/src/app/api/v0/city/invite/[invite]/route.ts:20]: Email needs to be part of the token, otherwise any user obtaining the token could use the invite. If it's part of the signed JWT body, you can be sure that it's only valid for a single user. Could be their email or their user ID.

Comment (lemilonkh) [reply app/src/app/api/v0/city/invite/[invite]/route.ts:20]: You can still accept it as the search parameter like here, just make sure you verify it.

Comment (lemilonkh) [reply app/src/app/api/v0/city/invite/[invite]/route.ts:20]: I'm seeing you pass it into the token, but you don't compare the contents of the decrypted token to the passed email address.

app/src/app/api/v0/city/invite/route.ts
```diff
@@ -9,70 +9,58 @@ import jwt from "jsonwebtoken";
...
+  }
 
-    const city = await db.models.City.findOne({
-      where: { locode: body.locode },
-    });
-    if (!city) {
-      throw new createHttpError.NotFound("City not found");
-    }
+  const invitationCode = jwt.sign(
+    { email: session?.user.email },
```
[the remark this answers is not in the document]
lemilonkh: The email in the token should not be the current user's email but rather the email of the user that is being invited.

app/src/app/api/v0/city/invite/route.ts
```diff
@@ -9,70 +9,58 @@ import jwt from "jsonwebtoken";
...
+  const host = process.env.HOST ?? "http://localhost:3000";
+  const sendInvite = await sendEmail({
+    to: body.email!,
+    subject: "Test Invite",
+    html: render(
+      InviteUserTemplate({
+        url: `${host}/api/v0/city/invite/${invite.id}?token=${invitationCode}&email=${body.email}`,
+        user: { email: body.email, name: body.name },
+        city,
+        invitee: { name: session?.user.name!, email: session?.user.email! },
```
lemilonkh, line 56: Why is the invitee the current user? Shouldn't it be the user that's being invited?
Otherwise rename this variable to `inviter` or `invitingUser` (both here and in the email template)

app/src/components/Modals/add-user-modal.tsx
```diff
@@ -27,36 +27,37 @@ import { TFunction } from "i18next";
 interface AddUserModalProps {
   isOpen: boolean;
   onClose: () => void;
-  cityId: string | undefined;
   t: TFunction;
+  userInfo: UserAttributes;
+  defaultCityId: string | undefined;
```
lemilonkh, line 32:
```suggestion
  defaultCityId?: string;
```
Works too and is more readable as an optional parameter imo

app/src/components/Modals/add-user-modal.tsx
```diff
@@ -27,36 +27,37 @@ import { TFunction } from "i18next";
 interface AddUserModalProps {
   isOpen: boolean;
   onClose: () => void;
-  cityId: string | undefined;
   t: TFunction;
+  userInfo: UserAttributes;
```
lemilonkh, line 31: I think this needs to be marked with a ? (or `| null`) as well, since it'll be `null` while the data is loading from the API.

app/src/components/Modals/add-user-modal.tsx
```diff
@@ -94,8 +95,9 @@ const AddUserModal: FC<AddUserModalProps> = ({
         });
       } else {
         await inviteUser({
-          userId: res && res.data.data.userId!,
-          locode: userInfo.defaultCityLocode!,
+          name: res.data.data ? res.data.data.name! : data.name,
```
lemilonkh, line 98: Why res.data.data ? You can unwrap the API response in the Redux api.ts file like we do for the other API calls.

app/src/components/Modals/add-user-modal.tsx
```diff
@@ -140,7 +141,7 @@ const AddUserModal: FC<AddUserModalProps> = ({
     <>
       <Modal blockScrollOnMount={false} isOpen={isOpen} onClose={onClose}>
         <ModalOverlay />
-        <ModalContent minH="600px" minW="568px" marginTop="10%">
+        <ModalContent minH="300px" minW="568px" marginTop="10%">
```
lemilonkh, line 144: Why is this a percentage? This means that the margin will be different depending on the page/ screen size?

app/src/components/Modals/add-user-modal.tsx
```diff
@@ -171,23 +172,6 @@ const AddUserModal: FC<AddUserModalProps> = ({
...
                 />
-
-                <FormSelectInput
-                  label={t("role")}
-                  value={inputValue}
-                  register={register}
-                  error={errors.role}
-                  id="role"
-                  onInputChange={onInputChange}
-                />
```
lemilonkh, line 174: This should only be shown for admins imo? Fine to comment out for now but we could also keep it and only conditionally render it when the current user is an admin?

app/src/components/Tabs/my-profile-tab.tsx
```diff
@@ -934,9 +936,7 @@ const MyProfileTab: FC<MyProfileTabProps> = ({
                                             background: "content.link",
                                             color: "white",
                                           }}
-                                          onClick={() => {
-                                            alert(city.cityId);
-                                          }}
+                                          onClick={() => {}}
```
lemilonkh, line 939: Remove this if it's supposed to do nothing. Or add the logic for it.

app/src/models/CityInvite.ts
```diff
@@ -1,13 +1,11 @@
...
 import { DataTypes, Model, Optional } from "sequelize";
-import { User, UserAttributes, UserId } from "./User";
 import { City, CityId } from "./City";
 import { UserFileCreationAttributes } from "./UserFile";
 
 export interface CityInviteAttributes {
   id: string;
-  userId?: string;
-  locode?: string;
+  cityId?: string;
```
lemilonkh, line 8: Add email here? Otherwise we can't show it in a list which users have been invited.

lemilonkh (approved):

PR #355, opened 2024-03-04
Feature/disconnect third party data

cephaschapa:
Changes
- Adds route to delete inventoryValue
- Adds api mutation in api service
- Adds disconnect functionality to third part data ui
- Adds access control for inventoryValue


.github/workflows/web-actions.yml
```diff
@@ -126,8 +126,9 @@ jobs:
           kubectl set env deployment/cc-web-deploy SMTP_PASSWORD=${{secrets.SMTP_PASSWORD}}
           kubectl set env deployment/cc-web-deploy NEXTAUTH_SECRET=${{secrets.NEXTAUTH_SECRET}}
           kubectl set env deployment/cc-web-deploy RESET_TOKEN_SECRET=${{secrets.RESET_TOKEN_SECRET}}
+          kubectl set env deployment/cc-web-deploy VERIFICATION_TOKEN_SECRET=${{secrets.VERIFICATION_TOKEN_SECRET}}
+          kubectl rollout restart deployment cc-web-deploy -n default
```
lemilonkh, line 130: Remove this line, it should only be called at the very end imo

app/src/app/[lng]/data/[step]/page.tsx
```diff
@@ -346,6 +346,8 @@ export default function AddDataSteps({
         dataSourceIds: [source.datasourceId],
       }).unwrap();
 
+      console.log(response);
```
Comment (lemilonkh) [inline app/src/app/[lng]/data/[step]/page.tsx:349]: Remove logging

app/src/app/api/v0/inventory/[inventory]/inventoryvalue/[inventoryvalue]/route.ts
```diff
@@ -0,0 +1,27 @@
...
+  const inventoryValue = await UserService.findUserInventoryValue(
+    params.inventoryvalue,
+    params.inventory,
+    session,
+  );
+
+  await inventoryValue.destroy();
+
+  return NextResponse.json({ data: inventoryValue, deleted: true });
+});
```
Comment (lemilonkh) [inline app/src/app/api/v0/inventory/[inventory]/inventoryvalue/[inventoryvalue]/route.ts:27]: I think we have the same routes in `/api/v0/inventory/[inventory]/value/[subcategory]/route.ts`. They are just being accessed by the SubCategory ID as the identifier instead of the InventoryValue ID because when creating one (upsert) it doesn't have a set ID yet (and it's easier to search for a specific SubCategory's inventory value from the add data form etc.).

So I would prefer merging the routes or maybe just keeping the existing ones. Since will most likely also have the subCategoryId available in the frontend when calling them.

app/src/backend/UserService.ts
```diff
@@ -6,6 +6,7 @@ import type { City } from "@/models/City";
 import type { Inventory } from "@/models/Inventory";
 import type { User } from "@/models/User";
 import type { Includeable } from "sequelize";
+import { InventoryValue } from "@/models/InventoryValue";
```
lemilonkh, line 9: You can do `import type` here as well, so the class isn't added to the final JS output since it's only used during type checking.

app/src/i18n/locales/en/data.json
```diff
@@ -159,5 +159,6 @@
   "download-template": "Download template to upload data",
   "file-upload-steps": "Follow the steps in our template and upload the data to calculate your inventory without waiting for CityCatlyst's data review.",
   "add-custom": "Add custom",
-  "save-missing-scope-info": "Save or fill in missing information before proceeding."
+  "save-missing-scope-info": "Save or fill in missing information before proceeding.",
+  "disconnect-data": "Disconnect Data"
```
lemilonkh, line 163: Can you get the translations for our other supported languages? I can give you the German ones if that helps.

```
"save-missing-scope-info": "Speichern Sie oder geben Sie die fehlenden Informationen ein before Sie fortfahren.",
  "disconnect-data": "Datenquelle trennen"
  ```

File: app/src/services/api.ts [hunk 6]
```diff
@@ -354,6 +355,15 @@ export const api = createApi({
       transformResponse: (response: { data: EmissionsFactorResponse }) =>
         response.data,
     }),
+    disconnectThirdPartyData: builder.mutation({
+      query: ({ inventoryId, inventoryValueId }) => ({
+        method: "DELETE",
+        url: `/inventory/${inventoryId}/inventoryvalue/${inventoryValueId}`,
```
lemilonkh, line 361: Could use the existing API endpoint that uses `value` instead of `inventoryvalue` and then passing in the SubCategory ID instead of the InventoryValue ID

lemilonkh:
Can you also write a test for the new functionality in `UserService`? Might make sense to add a new test suite for it. Maybe we'll do it once the Jest PR is merged though?

app/src/backend/UserService.ts
cephaschapa: route no longer exists deletion

PR #368, opened 2024-03-11
Create script for making default admin account and add route to change user role

lemilonkh:
- Adds script `scripts/create-admin.ts`.
  - You need to have the `DEFAULT_ADMIN_EMAIL` and `DEFAULT_ADMIN_PROCESS` env vars defined for it.
- Adds route `/api/v0/auth/role` to change a user's role. Can only be called by admin accounts.
  - Adds tests for it
  - Docs are here: https://github.com/Open-Earth-Foundation/CityCatalyst/wiki/CityCatalyst-Backend-API#post-apiv0authrole


app/package.json
```diff
@@ -26,7 +26,8 @@
     "cy:run": "cypress run",
     "cy:test": "start-server-and-test start http://localhost:3000/en/auth/login cy:run",
     "prettier": "npx prettier . --write",
-    "email": "email dev --dir src/lib/emails"
+    "email": "email dev --dir src/lib/emails",
+    "create-admin": "tsx scripts/create-admin.ts"
```
evanp, line 30: Does this work? I don't think we have tsx in our Docker image. I've run into a similar problem with the catalog sync.

app/scripts/create-admin.ts
```diff
@@ -0,0 +1,43 @@
...
+      "create-admin.ts: Missing default admin credentials DEFAULT_ADMIN_EMAIL and DEFAULT_ADMIN_PASSWORD in env!",
+    );
+    return;
+  }
+
+  const passwordHash = await bcrypt.hash(
+    process.env.DEFAULT_ADMIN_PASSWORD,
+    12,
+  );
+  const user = await db.models.User.create({
```
evanp, line 27: I don't think this is idempotent. If this runs every time we deploy, it's going to cause an error, isn't it?

lemilonkh: It would fail the second time it's run, yes. I was going to have it as a one-off task instead of running it on each deploy, but I can make it check if the account already exists and exit gracefully in case that's important.

app/package.json
lemilonkh: You're right. tsx is currently added as a dev dependency. Not sure if we have to transpile it at build time with a custom task or if adding tsx to the main dependencies is fine?

lemilonkh: Added tsx to normal dependencies for now

cephaschapa (approved):

PR #371, opened 2024-03-12
Feature/allow citywide file access

cephaschapa:
Changes
- adds migration to include city_id in userFile table
- updates userFile GET route to filter cities by city_id
- updates userFile POST route to include city_id during record creation
- moves userFile routes to city dir
- Updates ui api service routes
- UI updates


app/src/util/validation.ts
```diff
@@ -116,6 +116,7 @@ export type CreatePopulationRequest = z.infer<typeof createPopulationRequest>;
 // user file schema validation
 export const createUserFileRequset = z.object({
   userId: z.string().uuid().optional(),
+  cityId: z.string().uuid().optional(),
```
lemilonkh, line 119: Tbh I think userId and cityId are mandatory, because without them we can't really reference the files anymore. So would be fine to remove `.optional()` here.

lemilonkh (approved):

lemilonkh (approved):

PR #375, opened 2024-03-14
Fix Codecov failing on Dependabot PRs and clean up code coverage reporting

lemilonkh:


.github/workflows/web-actions.yml
```diff
@@ -53,21 +53,25 @@ jobs:
...
-      - name: Report code coverage
-        uses: zgosalvez/github-actions-report-lcov@v3
-        with:
-          coverage-files: ./app/lcov*.info
-          minimum-coverage: 40
-          artifact-name: code-coverage-report
-          github-token: ${{ secrets.GITHUB_TOKEN }}
-          working-directory: ./app
-          update-comment: true
+      # - name: Setup LCOV
```
evanp, line 56: Don't comment out code. We use a version control system. We can get it back when we need to. You can just delete this.

evanp (changes requested):
LGTM, except for the commented code.

PR #386, opened 2024-03-18
Allow multiple inventory years, add region and country population during onboarding

lemilonkh:


cephaschapa (approved):

app/src/app/api/v0/city/[city]/population/route.ts
```diff
@@ -2,25 +2,69 @@ import UserService from "@/backend/UserService";
...
+  } else {
+    countryPopulation = await db.models.Population.create({
+      cityId,
+      countryPopulation: body.countryPopulation,
+      year: body.countryPopulationYear,
     });
   }
 
-  return NextResponse.json({ data: population });
+  return NextResponse.json({ success: true });
```
Comment (cephaschapa) [inline app/src/app/api/v0/city/[city]/population/route.ts:69]: Are we not returning population data?

cephaschapa (approved):

app/src/app/api/v0/city/[city]/population/route.ts
Comment (lemilonkh) [reply app/src/app/api/v0/city/[city]/population/route.ts:69]: We would have to return 3 different entries and it's the same data that has been saved. But can do.

Comment (lemilonkh) [reply app/src/app/api/v0/city/[city]/population/route.ts]: Done :+1: 
It's now returning `{ data: {cityPopulation, regionPopulation, countryPopulation} }`

PR #387, opened 2024-03-20
Show error if population is missing for downscaled data sources on SourceDrawer

lemilonkh:
image


app/src/app/[lng]/data/[step]/SourceDrawer.tsx
```diff
@@ -171,6 +174,12 @@ export function SourceDrawer({
                 </Text>
               </HStack>
 
+              {sourceData?.issue && (
+                <Text color="semantic.danger" size="sm" mt={-4}>
```
Comment (cephaschapa) [inline app/src/app/[lng]/data/[step]/SourceDrawer.tsx:178]: could also use the app-theme text here for consistency

cephaschapa (approved):

PR #395, opened 2024-03-25
basic global-api testing

amanda-eames:
basic unit tests for global api end points and automatic comment for test report


lemilonkh (approved):
This is great, nice that you have set up codecov uploads already :+1: 
It's set tot a different name, so it probably computes a separate coverage percentage from the CityCatalyst backend. Otherwise we'll have to adjust the tags.

PR #398, opened 2024-03-29
Fix/invite collaborators to invite flow

cephaschapa:
Changes
- Adds inventory id from the req body in the signup route that allows new users to be associated to the city during the sign up process
- Adds dynamic inventory path to dashboard and subsequent pages
- Adds a logic to check if the inventory exists in the path otherwise it falls back to the default inventory
- Adds a custom 404 page: this will ensure that current CC users have a way to navigate back to the dashboard as the new changes will cause the router to navigate to a 404 page since the path url no longer exists
- Adds styling additions to app theme
- Adds inventory id to redirect path
- removed the fallback redirect if user is not found
- adds inventoryId to sign up validation


app/src/app/[...not_found]/page.tsx
```diff
@@ -0,0 +1,69 @@
...
+        <Button
+          onClick={() => router.push(`/${userInfo?.defaultInventoryId}`)}
+          gap="8px"
+          h="48px"
+          px="24px"
+          isLoading={isUserInfoLoading}
+        >
+          <Text fontSize="body.md">go to dashboard</Text>
+          <ArrowForwardIcon />
+        </Button>
```
Comment (lemilonkh) [inline app/src/app/[...not_found]/page.tsx:63]: Instead of manually applying the formatting and gap, you can use the `rightIcon` property of ChakraUI's `Button` component: https://chakra-ui.com/docs/components/button/usage#button-with-icon

app/src/app/[...not_found]/page.tsx
```diff
@@ -0,0 +1,69 @@
...
+        <Text
+          mb="48px"
+          w="0px"
+          width="400px"
+          textAlign="center"
+          lineHeight="32px"
+          letterSpacing="wide"
+        >
+          Seems like we&apos;ve wandered into the emission mist. Please go back
+          to the dashboard
```
Comment (lemilonkh) [inline app/src/app/[...not_found]/page.tsx:52]: I18n?

app/src/app/[...not_found]/page.tsx
```diff
@@ -0,0 +1,69 @@
...
+        </Text>
+        <Button
+          onClick={() => router.push(`/${userInfo?.defaultInventoryId}`)}
+          gap="8px"
+          h="48px"
+          px="24px"
+          isLoading={isUserInfoLoading}
+        >
+          <Text fontSize="body.md">go to dashboard</Text>
```
Comment (lemilonkh) [inline app/src/app/[...not_found]/page.tsx:61]: I18n?

app/src/app/[lng]/[inventory]/page.tsx
```diff
@@ -76,31 +76,35 @@ export default function Home({ params: { lng } }: { params: { lng: string } }) {
...
   if (!isUserInfoLoading && userInfo) {
     defaultInventoryId = userInfo.defaultInventoryId;
 
     // TODO also add this to login logic or after email verification to prevent extra redirect?
     // if the user doesn't have a default inventory, redirect to onboarding page
-    if (!defaultInventoryId) {
+    if (!inventoryId) {
       // fixes warning "Cannot update a component (`Router`) while rendering a different component (`Home`)"
-      setTimeout(() => router.push("/onboarding"), 0);
+      setTimeout(() => router.push(`/${defaultInventoryId}`), 0);
```
Comment (lemilonkh) [inline app/src/app/[lng]/[inventory]/page.tsx:97]: Update the comment in case you intend to always redirect here. Also the user might not have a `defaultInventoryId` set at this stage, so check if it is null and redirect to onboarding in that case (like before).

app/src/app/[lng]/[inventory]/settings/page.tsx
```diff
@@ -67,6 +70,8 @@ export default function Settings({
     skip: !userInfo,
   });
 
+  console.log(inventory?.year);
+
```
Comment (lemilonkh) [inline app/src/app/[lng]/[inventory]/settings/page.tsx:74]: Remove log

app/src/app/[lng]/[inventory]/settings/page.tsx
```diff
@@ -52,11 +53,13 @@ export default function Settings({
     skip: !userInfo,
   });
 
+  const { inventory: cityParam } = useParams();
+  const inventoryId = cityParam as string;
+
```
Comment (lemilonkh) [inline app/src/app/[lng]/[inventory]/settings/page.tsx:58]: Why are you calling it cityParam if it contains an inventory ID?

app/src/app/[lng]/auth/signup/page.tsx
```diff
@@ -42,11 +43,20 @@ export default function Signup({
...
     register,
     setError: setFormError,
+    setValue,
     formState: { errors, isSubmitting },
   } = useForm<Inputs>();
 
   const [error, setError] = useState("");
 
+  const searchParams = useSearchParams();
+  const params = searchParams.get("callbackUrl");
```
Comment (lemilonkh) [inline app/src/app/[lng]/auth/signup/page.tsx:53]: Rename this to `callbackUrl` so it's easier to understand. You have an URL string here, not a list of parameters.

app/src/app/[lng]/auth/signup/page.tsx
```diff
@@ -42,11 +43,20 @@ export default function Signup({
...
+  const inventory = params?.split("/").pop();
```
Comment (lemilonkh) [inline app/src/app/[lng]/auth/signup/page.tsx:54]: This could potentially break if there are other arguments/ ? or # at the end of the URL. You can use the built-in `URL` class to parse an URL and extract the `pathname`, which you can then call split and pop on.

Details here: https://developer.mozilla.org/en-US/docs/Web/API/URL/pathname

app/src/app/api/v0/city/invite/[invite]/route.ts
```diff
@@ -35,15 +36,12 @@ export const GET = apiHandler(async (req, { params, session }) => {
...
-  }
 
   const city = await db.models.City.findOne({
     where: { cityId: invite.cityId },
   });
 
   await user?.addCity(city?.cityId);
 
-  return NextResponse.redirect(host);
+  return NextResponse.redirect(`${host}/en/${inventory}`);
```
Comment (lemilonkh) [inline app/src/app/api/v0/city/invite/[invite]/route.ts:46]: Don't hard-code `en` here. You can leave it out and the middleware will automatically redirect to the correct language for the user while loading the page (that's how it works for all our other links).

app/src/app/api/v0/city/invite/route.ts
```diff
@@ -49,7 +49,7 @@ export const POST = apiHandler(async (req, { params, session }) => {
     subject: "City Catalyst - City Invitation",
     html: render(
       InviteUserTemplate({
-        url: `${host}/api/v0/city/invite/${invite.id}?token=${invitationCode}&email=${body.email}`,
+        url: `${host}/api/v0/city/invite/${invite.id}?inventoryId=${body.inventoryId}&&token=${invitationCode}&email=${body.email}`,
```
lemilonkh, line 52: Remove second & here (in &&)

app/src/components/InventorySelect.tsx
```diff
@@ -33,7 +33,9 @@ export const InventorySelect = ({
     await setUserInfo({
       cityId: inventory.cityId!,
       defaultInventoryId: inventory.inventoryId,
-    }).unwrap();
+    }).then(() => {
+      router.push(`/${inventory.inventoryId}`);
+    });
```
lemilonkh, line 38: Don't remove the unwrap as that means there is no promise AFAIK.
Also since this is an async function, you don't need to use .then but you can just put the router.push line after the await.

app/src/app/[...not_found]/page.tsx
Comment (cephaschapa) [reply app/src/app/[...not_found]/page.tsx:63]: yeah was wondering what it was called

app/src/app/api/v0/city/invite/[invite]/route.ts
Comment (cephaschapa) [reply app/src/app/api/v0/city/invite/[invite]/route.ts:46]: ahh missed this one

app/src/util/validation.ts
```diff
@@ -44,6 +44,7 @@ export const signupRequest = z
     confirmPassword: z.string().min(4),
     inviteCode: z.string().min(6).max(6),
     acceptTerms: z.literal<boolean>(true),
+    inventory: z.string().uuid(),
```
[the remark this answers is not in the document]
cephaschapa: good that you caught that, what do you think? I can leave it optional

PR #399, opened 2024-03-29
fix(ui): not redirecting to login page after pressing log out button

lemilonkh:


app/src/components/navigation-bar.tsx
```diff
@@ -47,7 +51,7 @@ export function NavigationBar({
 
     // change language in URL without reloading page
     const newPath = location.pathname.replace(/^\/[A-Za-z]+/, `/${language}`);
-    history.replaceState("", "", newPath);
+    history.replaceState(null, "", newPath);
```
cephaschapa, line 54: I think we should also consider making changes to all our parts using `history.replaceState` in the future too

cephaschapa (approved):

PR #400, opened 2024-03-29
global-api testing updated .env name

amanda-eames:
had wrong name .env file

### DIFF
--- a/.github/workflows/global-api-develop.yml
+++ b/.github/workflows/global-api-develop.yml
@@ -38,7 +38,7 @@ jobs:
           sleep 10
           createuser -w -h localhost -p 5432 -U postgres citycatalyst
           createdb -w -h localhost -p 5432 -U postgres citycatalyst -O citycatalyst
-          cp env.sample .env
+          cp sample.env .env
 
       - name: Build coverage file
         run: |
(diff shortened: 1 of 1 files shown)


evanp (changes requested):
Need to fix this filename before going further

.github/workflows/global-api-develop.yml
```diff
@@ -38,7 +38,7 @@ jobs:
           sleep 10
           createuser -w -h localhost -p 5432 -U postgres citycatalyst
           createdb -w -h localhost -p 5432 -U postgres citycatalyst -O citycatalyst
-          cp env.example .env
+          cp env.sample .env
```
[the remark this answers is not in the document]
amanda-eames: you are right!

.github/workflows/global-api-develop.yml
```diff
@@ -11,6 +11,14 @@ on:
...
       - .github/workflows/global-api-develop.yml
     branches: ["develop"]
+  pull_request:
+    paths:
+      - global-api/**
+      - k8s/cc-global-api-deploy.yml
+      - k8s/cc-global-api.yml
+      - k8s/cc-global-api-migrate.yml
+      - .github/workflows/global-api-develop.yml
+    branches: ["develop"]
```
lemilonkh, line 21: Don't need to use `develop` here, as it applies for pull requests into any branch

lemilonkh (approved):

.github/workflows/global-api-develop.yml
lemilonkh: You can just remove the branches line here

PR #403, opened 2024-03-29
Custom polygons

amanda-eames:
- added custom polygon ingestion for mendoza file 
- renamed the route with generic non-source name 
- added custom polygon to the run everything script


global-api/main.py
evanp, line 1: So, I think this change is kind of unrelated to the custom outlines one. Could you make it a separate PR? I think it's a good one, but let's keep them distinct.

Otherwise, this PR is good to go.

evanp (changes requested):
Looks great! I think the country_code_endpoint is a bit out of place, though. Can we make that change with a separate PR?

PR #411, opened 2024-04-02
fix(ui): wrong scale label being displayed for data sources on SourceDrawer

lemilonkh:
Was displaying "country" for all of them instead of "city" for the non-downscaled sources


evanp (approved):
This looks right!

PR #416, opened 2024-04-03
fix:adding_missing_columns

mfonsecaOEF:
this PR contains:
- new migration file that includes emission_factor_value and emission_factor_units columns in `country_code` table

### DIFF
--- a/global-api/migrations/versions/949c5b9cc18d_ef_in_country_code_table.py
+++ b/global-api/migrations/versions/949c5b9cc18d_ef_in_country_code_table.py
@@ -1,0 +1,28 @@
+"""EF_in_country_code_table
+
+Revision ID: 949c5b9cc18d
+Revises: 583858ff1aa8
+Create Date: 2024-04-03 10:24:31.131245
+
+"""
+from typing import Sequence, Union
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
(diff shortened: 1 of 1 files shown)


lemilonkh (approved):

evanp:
The migration failed when upgrading to test today. When I checked, I could see that this code just wouldn't work; it uses `alter_column` for columns it's trying to add.

I don't think non-working code should get through PRs. That's kind of why we have PRs.

@mfonsecaOEF before making a pull request, you have to at least run the code. With migrations, I'd like to know that you ran the upgrade _and_ the downgrade. Can you schedule a pairs programming session with me so we can practice writing migrations? It would be nice to include @amanda-eames so we can make sure we're doing things right.

@lemilonkh should the reviewer run migrations code forward and backwards at least once? Or should we run migrations in the GitHub actions, to automate the testing?

lemilonkh:
I thought the reason it's using alter column is to make the columns nullable, and the PR title was just wrong? @mfonsecaOEF 

I'll revert the merge.

amanda-eames:
If we are reverting this change then I won't approve #448 the loading scipts require the additional columns.

lemilonkh:
I've reverted the merge of this PR as it was unfinished. @mfonsecaOEF please create the new columns properly and create another PR.

evanp:
I fixed the problem in `develop`. The code is working now.

PR #419, opened 2024-04-04
Fix/user file error

cephaschapa:


app/src/app/[lng]/[inventory]/settings/page.tsx
```diff
@@ -60,14 +60,16 @@ export default function Settings({
...
   const cityId = inventory?.city.cityId;
+
   const { data: cityUsers } = api.useGetCityUsersQuery(
     { cityId: cityId! },
     { skip: !cityId },
   );
 
   const { data: userFiles } = api.useGetUserFilesQuery(cityId!, {
-    skip: !userInfo,
+    skip: !inventory,
```
Comment (lemilonkh) [inline app/src/app/[lng]/[inventory]/settings/page.tsx:72]:
```suggestion
    skip: !cityId,
```

You should have the actual variable that's being sent as a parameter here, not the value it's derived from. in case something went wrong in the middle, you don't induce an invalid request to the server that way.

lemilonkh (approved):

PR #439, opened 2024-04-15
Fix/show population values

cephaschapa:
Changes
- Updated the population route to return population values closest to the inventory year
- Added a helper function that checks whether the nearest year has a value of not proceeds to find the next year closest year with a value

### DIFF
--- a/app/src/app/api/v0/city/[city]/population/[year]/route.ts
+++ b/app/src/app/api/v0/city/[city]/population/[year]/route.ts
@@ -9,13 +9,13 @@ export const GET = apiHandler(async (_req: Request, { session, params }) => {
   const population = await db.models.Population.findAll({
     where: { cityId: city?.cityId },
   });
-  const populationEntry: PopulationEntry[] = population.map(({population, year})=>{
+  const populations: PopulationEntry[] = population.map(({population, year})=>{
     return {
       year,
       population: population!
     }
   })
  
-  const allPopulationData = findClosestYearToInventory(populationEntry, parseInt(params.year))
-  return NextResponse.json({ data: allPopulationData });
+  const closestYearPopulationData = findClosestYearToInventory(populations, parseInt(params.year))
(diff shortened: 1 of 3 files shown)


app/src/util/helpers.ts
```diff
@@ -156,3 +156,38 @@ export function findClosestYear(
     null as PopulationEntry | null,
   );
 }
+
+export function findClosestYearToInventory(
```
lemilonkh, line 160: This should already by in develop, right?
Or are we backporting changes from main here?

app/src/app/api/v0/city/[city]/population/[year]/route.ts
```diff
@@ -1,12 +1,21 @@
...
-});
+  const populationEntry: PopulationEntry[] = population.map(({population, year})=>{
+    return {
+      year,
+      population: population!
+    }
+  })
+ 
+  const allPopulationData = findClosestYearToInventory(populationEntry, parseInt(params.year))
+  return NextResponse.json({ data: allPopulationData });
```
Comment (lemilonkh) [inline app/src/app/api/v0/city/[city]/population/[year]/route.ts:20]: Variable could be named closestYearPopulationData or something like that

app/src/app/api/v0/city/[city]/population/[year]/route.ts
```diff
@@ -1,12 +1,21 @@
...
 export const GET = apiHandler(async (_req: Request, { session, params }) => {
   const city = await UserService.findUserCity(params.city, session);
-  const population = await db.models.Population.findOne({
-    where: { cityId: city?.cityId, year: params.year },
+  const population = await db.models.Population.findAll({
+    where: { cityId: city?.cityId },
   });
-  return NextResponse.json({ data: population });
-});
+  const populationEntry: PopulationEntry[] = population.map(({population, year})=>{
```
Comment (lemilonkh) [inline app/src/app/api/v0/city/[city]/population/[year]/route.ts:12]: Variable should be renamed to `populations` since it's an array

PR #441, opened 2024-04-15
feat(api): admin access control

lemilonkh:
- allows admin accounts to find, create, delete user files for cities that they don't belong to
- allows admin accounts to edit/ delete inventory values for other accounts

### DIFF
--- a/app/src/app/api/v0/city/[city]/file/route.ts
+++ b/app/src/app/api/v0/city/[city]/file/route.ts
@@ -1,4 +1,5 @@
 import NotificationService from "@/backend/NotificationService";
+import UserService from "@/backend/UserService";
 import AdminNotificationTemplate from "@/lib/emails/AdminNotificationTemplate";
 import { db } from "@/models";
 import { apiHandler } from "@/util/api";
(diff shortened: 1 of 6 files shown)


app/src/app/api/v0/city/[city]/file/route.ts
```diff
@@ -60,130 +59,119 @@ export const GET = apiHandler(async (_req: Request, context) => {
...
-      cityId,
-    },
-  });
-
-  if (!city) {
-    throw new createHttpError.Unauthorized("Unauthorized");
-  }
+export const POST = apiHandler(
+  async (req: NextRequest, { params, session }) => {
+    const service = new NotificationService(); // TODO cache this/ make it a singleton
```
Comment (cephaschapa) [inline app/src/app/api/v0/city/[city]/file/route.ts:64]: We can create a ticket for this

cephaschapa (approved):

app/src/app/api/v0/city/[city]/file/route.ts
Comment (lemilonkh) [reply app/src/app/api/v0/city/[city]/file/route.ts:64]: I'll make it part of the PR feedback ticket we've added to refactor this function

PR #444, opened 2024-04-15
feat: function for estimating value from a series

evanp:
Added an estimate for series of annual values. Will either return the exact value, an extrapolation forward or backwards, or an interpolation.

If the extrapolation is too far (greater than the length of the series), it returns null.

Includes unit tests.

### DIFF
--- a/app/src/util/series.ts
+++ b/app/src/util/series.ts
@@ -1,0 +1,46 @@
+export function estimate(series: { year: number, value: number }[], year: number): number | null {
+
+  // if it's in the series, just return the value
+
+  const exact = series.find(e => e.year === year);
+  if (exact) {
+    return exact.value;
+  }
+
+  // we need at least two points to interpolate or extrapolate
+
+  if (series.length < 2) {
+    return null;
+  }
+
+  if (year < series[0].year) { // extrapolate backwards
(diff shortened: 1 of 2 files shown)


evanp:
Note that this PR doesn't change the UI for getting population. That needs to be done separately!

app/src/util/series.ts
```diff
@@ -0,0 +1,46 @@
...
+    return null;
+  }
+
+  if (year < series[0].year) { // extrapolate backwards
+    let first = series[0]
+    let next = series.find(e => e.year - first.year >= first.year - year)
+    if (!next) {
+      return null
+    }
+    let rate = (1.0 * (next.value - first.value)) / (next.year - first.year);
```
lemilonkh, line 22: Why are you using 1.0 as a factor here? JS only deals with the Number type that is a float anyways.
Same for the two other occurrences below.

lemilonkh (approved):

app/src/util/series.ts
evanp: Fixed! Thanks for the idea; it's a bad habit. I took out the Math.round(), too, since there may be cases we want decimals here.

PR #448, opened 2024-04-16
Argentinian datasets

mfonsecaOEF:
this PR includes all the scripts and files from the transformation phase, which means:
- README file
- extraction script
- raw data from the source
- transformation script
- processed csv file with the current global-api schema
- load script
for the following sources:
- BEN
- ENARGAS
- SESCO


amanda-eames (commented):
Transformation scripts look good, need to add the following:
- region route for global-api
- add data source to seeder file
- update the run.sh file to include the new dataset

lemilonkh:
@amanda-eames why was the PR closed? I think it's still relevant, just needs to be updated?

PR #463, opened 2024-04-26
fix(ui): redirecting unauthenticated users to login page

cephaschapa:
Changes
- Enables checks in the useSession() hook to redirect unauthenticated users to login for both the root page and dashboard


app/src/app/[lng]/[inventory]/page.tsx
```diff
@@ -384,24 +392,25 @@ export default function Home({ params: { lng } }: { params: { lng: string } }) {
...
-                           fontFamily="heading"
-                           color="border.neutral"
-                           fontSize="headline.sm"
-                           fontWeight="semibold"
-                           lineHeight="32"
-                         >
-                           N/A
-                         </Text>
+                        {inventory?.city.area! == 0 ||
+                        inventory?.city.area === null ? (
```
Comment (lemilonkh) [inline app/src/app/[lng]/[inventory]/page.tsx:396]: It would make sense to do the null check first here, and use === for comparison with 0.

lemilonkh (approved):

PR #476, opened 2024-04-30
Added toast error for failed validation on adding a new city 

saviobosco:
Added toast error for failed validation on adding a new city and updated the command to set up the UI App

### DIFF
--- a/app/src/app/[lng]/onboarding/setup/page.tsx
+++ b/app/src/app/[lng]/onboarding/setup/page.tsx
@@ -800,6 +800,7 @@ export default function OnboardingSetup({
 
     if (!newData.city || !ocCityData?.actor_id || year < 0 || !data.locode) {
       // TODO show user toast? These should normally be caught by validation logic
+      makeErrorToast("Missing data, can't go to next step!");
       console.error("Missing data, can't go to next step!");
       return;
     }
--- a/app/src/app/[lng]/onboarding/setup/page.tsx
+++ b/app/src/app/[lng]/onboarding/setup/page.tsx
@@ -294,7 +294,7 @@ function SetupStep({
                     required: t("select-city-required"),
                   })}
                   autoComplete="off"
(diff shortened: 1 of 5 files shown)


app/src/app/[lng]/onboarding/setup/page.tsx
```diff
@@ -294,7 +294,7 @@ function SetupStep({
                     required: t("select-city-required"),
                   })}
                   autoComplete="off"
-                  onFocus={() => setOnInputClicked(true)}
+                  onKeyUp={() => setOnInputClicked(true)}
```
Comment (lemilonkh) [inline app/src/app/[lng]/onboarding/setup/page.tsx:297]: Why did you make this change? If anything I think both listeners would need to be connected, so perhaps keep the original line and just add this?

lemilonkh (changes requested):

lemilonkh:
This PR does a lot that's not covered by the title/ description. But should be mergeable if the onFocus listener is added back in on the onboarding page.

saviobosco:
Hello @lemilonkh  I'm happy to get your feedback.  This PR is three separate pull requests, I think GitHub just made it appear as one PR.

app/src/app/[lng]/onboarding/setup/page.tsx
Comment (saviobosco) [reply app/src/app/[lng]/onboarding/setup/page.tsx:297]: I changed the event listener from Onfocus to onKeyUp because the input Text field wasn't that reactive with the OnFocus Listener. The onFocus Listener was only fired when the input text field was focused.  In my experience, I had to type a city name, click outside the input field and, click back before it fetched the list of cities.  

Nice suggestion, Both listeners could be connected. I will add the onFocus too.

saviobosco:
Hi @lemilonkh kindly review the PR. I will be waiting for your feedback. Thank you.

lemilonkh (approved):

lemilonkh:
@saviobosco Thanks! :sparkles:

PR #486, opened 2024-05-10
fix(api): set HOST and NEXTAUTH_URL to correct values on test and prod envs

lemilonkh:
Reset password was broken on test and prod envs as the email contained a link to the dev version of CityCatalyst.
This was caused by the HOST env variable being set to the same URL for all envs in the k8s job file.

`HOST` and `NEXTAUTH_URL` (used for some redirects) are now being overwritten in the Github Actions jobs `web-test.yml` and `web-tab.yml`.


lemilonkh:
Has been fixed in other PRs

PR #494, opened 2024-05-22
fix:convertion_units_activity_values

mfonsecaOEF:
This PR contains:
- Conversion units fix in transformation file
- New processed csv file
- Folder name changed


amanda-eames (approved):
Rested the values look like the right range!

PR #495, opened 2024-05-22
fix:column_types

mfonsecaOEF:
This PR contains:
- new migration file to change column types to string to float in country table for emissions_value and activity_value


amanda-eames (approved):
testing passed

PR #496, opened 2024-05-23
Implement translations for data source seeder/ data catalogue route

mfonsecaOEF:
This PR contains:
- yaml file with the translations in `en` and `es` for different columns in the datasource table
- migration file with change of text type to jsonb type for those columns


amanda-eames (commented):
database testing passed @lemilonkh is there anything required on your end if the datatypes change from text to jsonb, if not we can approve.

mfonsecaOEF:
> database testing passed @lemilonkh is there anything required on your end if the datatypes change from text to jsonb, if not we can approve.

mfonsecaOEF:
@lemilonkh I changed the format of the file, csv to yaml.
Please LMK if I need to add something else to this PR

lemilonkh:
@mfonsecaOEF is the data catalogue route updated to handle these new data types? Did you test it?

lemilonkh:
@mfonsecaOEF let's make sure the `v0` endpoints return only the English strings, and add a v1 of the endpoint that returns `{ "en": "Name", "es": "Nombre", ... }` for every translatable string (that will be shown to the end user, so no formula identifiers etc.).

lemilonkh (approved):
Great, that should keep it compatible with the current CityCatalyst version! :sparkles:

PR #501, opened 2024-05-28
CDP API integration

evanp:
This is the first step towards CDP API integration. It includes a service for calling the CDP API, and an API endpoint for submitting CityCatalyst data to CDP.

Without an API key, it's not possible to test this code, so I'm going to need to park it for now. I'll ping CDP on Tue to see if they can get us a key in time for testing on Friday. Note: the test on Friday only needs to submit test data, not full emissions.

After that, I'll work on the next stage: including emissions data for the correct questions.


app/tests/api/inventory.test.ts
```diff
@@ -366,4 +369,15 @@ describe("Inventory API", () => {
...
   });
+
+  it("should submit an inventory to the CDP test API", async () => {
+    const req = mockRequest({});
+    const res = await submitInventory(req, {
+      params: { inventory: inventory.inventoryId },
+    });
+    assert.equal(res.status, 200);
+    const json = await res.json();
+    console.dir(json)
```
lemilonkh, line 380: Remove logging

lemilonkh (approved):
There is a lot of global API related code in this PR that would make sense to have separately/ sync the current state of develop and main for it.

evanp:
This was branched from `main` instead of `develop`. I made another patch for `develop` and applied it. So, I think this can close.

PR #502, opened 2024-05-28
Manual Input UI v2 for Stationary Energy

cephaschapa:
Changes

- Adds New UI for the manual input for stationery energy

TODO

- [ ] Translation


app/src/components/Cards/add-data-card.tsx
```diff
@@ -0,0 +1,65 @@
...
+  icon,
+  description,
+  scopeText,
+  step,
+  buttonText,
+  title,
+  inventory,
+}: AddDataCardProps) {
+  const router = useRouter();
+  console.log(inventory);
```
lemilonkh, line 27: Remove logging

app/src/components/Cards/methodology-card.tsx
```diff
@@ -0,0 +1,107 @@
...
+}) => {
+  const [isSelected, setIsSelected] = useState(false);
+  const handleRadioChange = () => {
+    setIsSelected(true);
+    // handleCardSelect(name);
+  };
+
+  const handleCardClick = () => {
+    setIsSelected(!isSelected);
+    // handleCardSelect(name);
```
lemilonkh, line 29: Why is the handle function not called? Not yet implemented is fine

app/src/components/Cards/suggested-activities-card.tsx
```diff
@@ -0,0 +1,95 @@
...
+      cursor="pointer"
+      _hover={{ shadow: "md" }}
+    >
+      <Box display="flex" alignItems="center">
+        <Checkbox
+          borderRadius="full"
+          __css={{
+            "& .chakra-checkbox__control": {
+              bg: "white",
+              color: "#2351DC", // TODO: figure out how to include theme variable in css
```
lemilonkh, line 32: You can get color values from the theme like this:
```ts
import { useToken } from "@chakra-ui/react";

function getColors(): string[] {
  const colors = ["interactive.connected", "interactive.tertiary"];
  const colorValues = useToken("colors", colors);
  return colorValues;
}
```

lemilonkh: We use it in the SegmentedProgress component for a more thorough example.

app/src/components/Modals/add-activity-modal.tsx
```diff
@@ -0,0 +1,440 @@
+"use client";
```
lemilonkh, line 1: What is the difference between add-activity-modal and add-activity-energy-consumption-modal?
We don't want to implement a specific React component for each type of methodology in the future, so it would be good to dynamically construct the form from the `InputMethodology` data once we have that.

lemilonkh (approved):

PR #515, opened 2024-06-07
Fix/manual input stationary energy UI changes

cephaschapa:
Changes
- UI QA changes regarding styling, flow, state and functionality


app/src/components/Modals/add-activity-modal.tsx
```diff
@@ -214,6 +216,11 @@ const AddActivityModal: FC<AddUserModalProps> = ({
                           placeholder={t("activity-data-amount-placeholder")}
                           borderRightRadius={0}
                           bgColor="base.light"
+                          _focus={{
+                            borderWidth: "1px",
+                            shadow: "none",
+                            borderColor: "content.link",
+                          }}
```
lemilonkh, line 223: These styles can potentially be moved to the app theme, so we can apply them as the default for most form inputs. Fine for now though.

lemilonkh (approved):

PR #539, opened 2024-06-26
Updated the README.md file with additional hints

mircorudolph:
I updated the README.md file with additional hints based on my experience when setting up the local dev environment on my system.

### DIFF
--- a/app/README.md
+++ b/app/README.md
@@ -3,6 +3,15 @@
 This is the main CityCatalyst Web app. This directory includes both the
 frontend and backend code for the app, as well as database management scripts.
 
+## Prerequisites
+
+This guide assumes you have installed and access to the following tools:
+
+1. git command line interface. Please follow [this link](https://git-scm.com/downloads) for installation instructions.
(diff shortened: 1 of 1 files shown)


app/README.md
```diff
@@ -96,6 +123,10 @@ npm run create-admin
 npm run dev
 ```
 
+The standard port is 3000 and the application can be opend at http://localhost:3000. Please check the log outputs, if a different port has been used.
+
+Use `johndeo@example.com` and `password` to login.
```
lemilonkh, line 128: I don't think we create this example account? You need to sign up first.

lemilonkh: Also it's johndoe

lemilonkh (approved):
Thanks 🙌🏼✨

cephaschapa (approved):

app/README.md
mircorudolph: In the example.env file it is listed as:

DEFAULT_ADMIN_EMAIL=johndeo@example.com
DEFAULT_ADMIN_PASSWORD=password

For me it worked like this

mircorudolph: That is happening in the script 

```
## set up login in credentials
npm run create-admin
```

PR #553, opened 2024-07-05
Feature/on 1874 climate advisor

mircorudolph:
Added inventory context to the climate advisor (chatbot)

### DIFF
--- a/app/src/app/[lng]/[inventory]/cdp/page.tsx
+++ b/app/src/app/[lng]/[inventory]/cdp/page.tsx
@@ -12,9 +12,21 @@ import {
 } from "@chakra-ui/react";
 import React from "react";
 import { useTranslation } from "@/i18n/client";
+import { api } from "@/services/api";
 
-function Page({ params: { lng } }: { params: { lng: string } }) {
+function Page({
+  params: { lng, inventory },
+}: {
+  params: { lng: string; inventory: string };
+}) {
   const { t } = useTranslation(lng, "cdp");
+  const [connectToCDP] = api.useConnectToCDPMutation();
+  const handleConnectToCDP = async () => {
+    await connectToCDP({ inventoryId: inventory }).then((res) =>
+      console.log(res),
+    );
+    console.log(inventory);
+  };
   return (
     <Box>
       <NavigationBar lng="" />
@@ -32,7 +44,7 @@ function Page({ params: { lng } }: { params: { lng: string } }) {
             <Text>{t("submit-data-to-cdp")}</Text>
           </CardBody>
           <CardFooter p="0">
-            <Button className="w-[100%]" h="50px">
+            <Button onClick={handleConnectToCDP} className="w-[100%]" h="50px">
               {t("submit-data-to-cdp")}
             </Button>
           </CardFooter>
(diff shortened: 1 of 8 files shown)


app/src/app/api/v0/chat/[inventory]/route.ts
```diff
@@ -0,0 +1,277 @@
...
+]);
+const messagesSchema = z.array(
+  z.object({ content: z.string(), role: roleSchema }),
+);
+const chatRequest = z.object({
+  messages: messagesSchema,
+});
+type Messages = z.infer<typeof messagesSchema>;
+
+// Function below is not being referenced/implemented
```
Comment (lemilonkh) [inline app/src/app/api/v0/chat/[inventory]/route.ts:44]: Feel free to remove the function instead of commenting it out. It was mainly used for debugging as you can't directly print out the OpenAI response to the console (since it's a stream).

If we need it again, we can find it in the git history.

app/src/app/api/v0/chat/[inventory]/route.ts
```diff
@@ -0,0 +1,277 @@
...
+}
+
+/**
+ * Creates a system message with context included
+ */
+function create_prompt_template(inventory: Inventory): string {
+  /**
+   * Helper function to extract the latest population value
+   */
+  function find_last_city_population(
```
Comment (lemilonkh) [inline app/src/app/api/v0/chat/[inventory]/route.ts:107]: Usually, you don't want the last one for all of these, rather the one that's closest to the selected inventory year (or even better, use interpolation/ extrapolation).

There are already helper methods in the codebase that do this (at least the first part). See getClosestPopulationYear or similar in a util.ts file.

Also, method names in our TS code style are not snake_case but should be camelCase.

Comment (mircorudolph) [reply app/src/app/api/v0/chat/[inventory]/route.ts:107]: `findClosestYearToInventory` returns the Population object  that contains the (city) population closest to the given year. This object may or may not contain populated fields for countyPopulation or regionPopulation.

`findClosestYear` returns he Population object back closest to the given year which contains any of the three populations fields. 

None of these two functions return all 3 values closest to the given year (e.g. one entry might only have 2 values, another entry would have the missing 3rd entry).

I will change my file to use `findClosestYearToInventory` for the latest (city) population for now.

app/src/app/api/v0/chat/[inventory]/route.ts
Comment (mircorudolph) [reply app/src/app/api/v0/chat/[inventory]/route.ts:44]: Thanks and done

app/src/app/[lng]/[inventory]/cdp/page.tsx
```diff
@@ -12,9 +12,21 @@ import {
...
+}: {
+  params: { lng: string; inventory: string };
+}) {
   const { t } = useTranslation(lng, "cdp");
+  const [connectToCDP] = api.useConnectToCDPMutation();
+  const handleConnectToCDP = async () => {
+    await connectToCDP({ inventoryId: inventory }).then((res) =>
+      console.log(res),
+    );
+    console.log(inventory);
```
Comment (cephaschapa) [inline app/src/app/[lng]/[inventory]/cdp/page.tsx:28]: Let's remove the console now 👍

cephaschapa (approved):

app/src/app/[lng]/[inventory]/cdp/page.tsx
Comment (mircorudolph) [reply app/src/app/[lng]/[inventory]/cdp/page.tsx]: Done

PR #558, opened 2024-07-08
Fix/manual input refactor

cephaschapa:
Changes
- Adds reusable components for dynmamic rendering
- Reduces subsector pages from 2k to less than 300 lines
- Adds a json file to manage form for each scope
- Connects form to activityData endpoint
- Adds toasts for form submition
- Moved root subsector page state into redux subsectorSlice


app/src/app/[lng]/[inventory]/data/[step]/page.tsx
```diff
@@ -811,6 +811,7 @@ export default function AddDataSteps({
                           sectorId: subSector.sectorId,
                           subsectorId: subSector.subsectorId,
                           referenceNumber: subSector.referenceNumber,
+                          subCatetories: subSector.subCategories,
```
Comment (evanp) [inline app/src/app/[lng]/[inventory]/data/[step]/page.tsx:814]: subcategories misspelled

app/src/components/Modals/add-activity-modal.tsx
```diff
@@ -189,13 +225,8 @@ const AddActivityModal: FC<AddUserModalProps> = ({
...
                 <FormControl>
                   <BuildingTypeSelectInput
-                    options={[
-                      t("all-fuels"),
-                      t("natural-gas"),
-                      t("propane"),
-                      t("heating-oil"),
-                    ]}
-                    title={t("fuel-type")}
+                    options={formInputs.fields[1].options}
```
evanp, line 228: You are using literal numbers for the indexes on the Fields array. do we have a fixed number of elements here? should this be a loop, or maybe use constants that indicate which field is being shown?

evanp (changes requested):
Looks good. A couple of changes suggested.

app/src/components/Modals/add-activity-modal.tsx
cephaschapa: Yes we have a fixed number of elements.

amanda-eames (approved):
Looks good Cephas!

PR #565, opened 2024-07-16
simple bug fix and tested access to inventory data

mircorudolph:
bug fix: when a city had no population, using the Climate Advisor caused a 500 error. Added optional chaining operator "?" to fix this.
feature: Added minimal inventory context data to test if it is generally available. More will be added once the manual input 2.0 is working properly.


lemilonkh (approved):
Thanks 🙌🏼✨

cephaschapa (approved):
Looks good we may need to check for other missing data too in a future PR. 👌

PR #572, opened 2024-07-18
Fix/manual input changes

cephaschapa:
changes
- Styling changes
- Adjusts form schema


app/src/app/[lng]/[inventory]/data/[step]/page.tsx
```diff
@@ -624,21 +624,24 @@ export default function AddDataSteps({
...
-        <div className=" w-[1090px] max-w-full mx-auto px-4">
-          <Box w="full" display="flex" alignItems="center" gap="16px" mb="35px">
+        <div className=" w-[1090px] mx-auto px-4  ">
+          <Box
+            w="full"
+            display="flex"
+            alignItems="center"
+            gap="16px"
+            mb="28px"
+            className={` ${scrollPosition > 0 ? "hidden" : "flex"} transition-all duration-50 ease-linear`}
```
Comment (lemilonkh) [inline app/src/app/[lng]/[inventory]/data/[step]/page.tsx:640]: You use scrollPosition > 0 so much, it might be worth defining a variable for it above  
E.g. `const isExpanded = scrollPosition > scrollResizeHeaderThreshold;`

app/src/components/Modals/add-activity-modal.tsx
```diff
@@ -181,7 +185,23 @@ const AddActivityModal: FC<AddUserModalProps> = ({
...
+
+  const [isEmissionFactorInputDisabled, setIsEmissionFactorInputDisabled] =
+    useState<boolean>(true);
+
+  const onEmissionFactorTypeChange = (e: any) => {
+    const emissionFactorType = e.target.value;
+    if (
+      emissionFactorType === "Local" ||
+      emissionFactorType === "Regional" ||
+      emissionFactorType === "National"
```
lemilonkh, line 198: This condition might need to be adjusted once we have data for national emissions factors (e.g. for the US we already do I believe).

app/src/components/Modals/add-activity-modal.tsx
```diff
@@ -333,10 +361,11 @@ const AddActivityModal: FC<AddUserModalProps> = ({
                       {...register("activity.emissionFactorType")}
```
lemilonkh, line 361: I think you need to change the keys that are being written to depending on the `name` property of the formInput field?

app/src/components/Tabs/Activity/activity-tab.tsx
```diff
@@ -423,7 +425,7 @@ const ActivityTab: FC<ActivityTabProps> = ({
                         methodology used to collect the data and calculate your
                         emissions.{" "}
                         <Link
-                          href="/"
+                          href="https://docs.google.com/spreadsheets/d/1cKN9jjxs6gLHTbuJyclsJaVDA_0CQxSg8qSjTSDr5eU/edit?pli=1#gid=2028597557 No, this one https://ghgprotocol.org/ghg-protocol-cities"
```
lemilonkh, line 428: Are we sure we want to link to a spreadsheet here?

I think it should be the link a the end?
```suggestion
                          href="https://ghgprotocol.org/ghg-protocol-cities"
```

app/src/features/city/subsectorSlice.ts
```diff
@@ -1,8 +1,8 @@
 import { RootState } from "@/lib/store";
 import { SubSectorAttributes } from "@/models/SubSector";
 import { PayloadAction, createSlice } from "@reduxjs/toolkit";
-import stationaryEnergyForm from './data/stationary-form-inputs.json'
 import { SubCategory, SubCategoryAttributes } from "@/models/SubCategory";
+import { SectorFormSchema } from "@/util/form-schema";
```
lemilonkh, line 5:
```suggestion
import { type SectorFormSchema } from "@/util/form-schema";
```

lemilonkh: Same for e.g. SubCatergoryAttributes, as it is only used as a type definition and doesn't need to be part of the compiled JS output.

app/src/lib/store.ts
```diff
@@ -33,7 +33,7 @@ const persistConfig = {
   key: "root",
   version: 1,
   storage,
-  blacklist: ["api"],
+  blacklist: ["api","subsector"],
```
lemilonkh, line 36: We should probably add some more keys here, as I've noticed it gets quite large after a while... Like e.g. the OpenClimate API one.

app/src/components/Modals/add-activity-modal.tsx
cephaschapa: Great i'll leave a comment here

app/src/app/[lng]/[inventory]/data/[step]/[subsector]/page.tsx
```diff
@@ -99,21 +136,22 @@ function SubSectorPage({
...
         bg="background.backgroundLight"
-        borderColor="border.neutral"
-        borderBottomWidth={scrollPosition > 0 ? "1px" : ""}
-        className="fixed z-10 top-0 w-full pt-[180px] h-[400px]"
-        mt={scrollPosition > 0 ? "-230px" : ""}
-        animate={{
-          y: scrollPosition > 0 ? 0 : -50,
-        }}
-        transition={{ duration: 0.2 }}
+        className={`fixed z-10 top-0 w-full ${scrollPosition > 0 ? "pt-[50px] h-[200px]" : "pt-[100px] h-[400px]"} transition-all duration-50 ease-linear`}
```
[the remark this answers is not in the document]
Comment (cephaschapa) [reply app/src/app/[lng]/[inventory]/data/[step]/[subsector]/page.tsx]: I actually used framer at first, then discovered it somehow wouldn't let me transition the texts during scrolling, but ill keep exploring it.

app/src/components/Modals/add-activity-modal.tsx
```diff
@@ -429,7 +419,7 @@ const AddActivityModal: FC<AddUserModalProps> = ({
                         h="48px"
                         shadow="1dp"
                         borderRightRadius={0}
-                        {...register("activity.co2EmissionFactor")}
+                        {...register(formInputs?.fields[4]?.name)}
```
lemilonkh, line 422: Does it need to have `"activity."` before the field name?

lemilonkh: Same for the other fields

lemilonkh (approved):

PR #587, opened 2024-07-23
feat(ui): adds i18n number formatting

cephaschapa:
changes
- Adds number formatting to json
- Applies number formatting to total emission

### DIFF
--- a/app/src/i18n/locales/en/data.json
+++ b/app/src/i18n/locales/en/data.json
@@ -287,5 +287,6 @@
   "file-deletion-error-description": "Something went wrong during file deletion",
   "file-deletion-success": "file deleted successfully",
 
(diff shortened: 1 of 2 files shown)


app/src/i18n/locales/en/data.json
```diff
@@ -287,5 +287,6 @@
   "file-deletion-error-description": "Something went wrong during file deletion",
   "file-deletion-success": "file deleted successfully",
 
-  "coming-soon": "Coming Soon"
+  "coming-soon": "Coming Soon",
+  "intlNumber": "{{val, number}}"
```
lemilonkh, line 291: We could move this to translation.json, so it is available on any page? Fine otherwise though.

lemilonkh (approved):
Thanks ✨

PR #597, opened 2024-07-29
fix(ui): adds addon options for emission factor values

cephaschapa:
Changes
- Adds options for addons which fixes input issue for emission factor values
- Updates form modal


lemilonkh (approved):
Thanks :sparkles:

PR #598, opened 2024-07-29
fix: remove empty inventory string from request [OC-1429]

thehighestprimenumber:
When no inventoryId was found, we were sending {... inventory: ""} in the request, this triggered an error from zod. Now we only add the inventory key if the field is not null and not an empty string.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -153,3 +153,6 @@ global-api/data/**
 
 #Python virtual env
 global-api/env
+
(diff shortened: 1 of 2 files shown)


thehighestprimenumber:
https://openearth.atlassian.net/browse/OC-1429

lemilonkh (approved):
Great spot, thanks :sparkles:

PR #601, opened 2024-07-30
fix: sector-form-schema.json

isaaccodekill:
Fixes an issue with the sector-form-schema.json file causing the app to crash, also adds @next/env to package.json (necessary for the create-admin script to work)

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -33,6 +33,7 @@
     "@chakra-ui/next-js": "^2.2.0",
     "@chakra-ui/react": "^2.8.2",
     "@huggingface/inference": "^2.8.0",
+    "@next/env": "^14.2.5",
     "@react-email/components": "^0.0.19",
     "@reduxjs/toolkit": "^2.2.3",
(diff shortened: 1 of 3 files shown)


lemilonkh (approved):

app/package.json
```diff
@@ -33,6 +33,7 @@
     "@chakra-ui/next-js": "^2.2.0",
     "@chakra-ui/react": "^2.8.2",
     "@huggingface/inference": "^2.8.0",
+    "@next/env": "^14.2.5",
```
lemilonkh, line 36: Let's move this to a separate PR, as it causes issues in our CI pipeline it seems

lemilonkh (changes requested):

app/package.json
lemilonkh: Or commit the required update for `package-lock.json`, then we can keep it :sweat_smile:

lemilonkh (approved):

PR #602, opened 2024-07-30
style: run all ts and json files through prettier

evanp:
I set up my `prettier` config, but I wanted to check how we've been doing so far. I ran `prettier` on the `app` dir, and got these file changes. Probably worth doing on a regular basis!

I can see a good argument for just leaving these changes until they're needed, but I figured it might be worth it to get us on a good footing. Happy to go either way.


lemilonkh (approved):
Thanks ✨

PR #620, opened 2024-08-05
translate methodology cards

thehighestprimenumber:


lemilonkh (approved):

app/src/i18n/locales/en/data.json
```diff
@@ -288,5 +288,153 @@
   "file-deletion-success": "file deleted successfully",
 
   "coming-soon": "Coming Soon",
-  "intlNumber": "{{val, number}}"
+  "intlNumber": "{{val, number}}",
+  "I.1.1-direct-measure": "Direct Measure",
+  "I.1.1-direct-measure-description": "Direct metering or monitoring of emissions in residential buildings.",
+  "I.1.2-direct-measure": "Direct Measure",
```
lemilonkh, line 294: Not sure about all of these repeated strings, we could only use the ref number for the description? Fine for now anyways...

lemilonkh (approved):

app/src/i18n/locales/en/data.json
thehighestprimenumber: The description key is not in the json, it's created in the code by appending `-description` to the ID. Since we need different descriptions, we need different id's.

PR #621, opened 2024-08-06
feat: added data files for waste default values

amanda-eames:


amanda-eames:
- Need to review waste structure with Mau

app/seed-data/formula_values/scripts/utils.py
```diff
@@ -0,0 +1,107 @@
...
+def uuid_generate_v4():
+    """generate a version 4 UUID"""
+    return uuid.UUID(int=random.getrandbits(128), version=4)
+
+
+def uuid_generate_v3(name, namespace=uuid.NAMESPACE_OID):
+    """generate a version 3 UUID from namespace and name"""
+    assert isinstance(name, str), "name needs to be a string"
+    assert isinstance(namespace, uuid.UUID), "namespace needs to be a uuid.UUID"
+    return str(uuid.uuid3(namespace, name))
```
lemilonkh, line 51: I think we have these functions repeated quite a bit in different data import scripts for the seed data. Maybe worth moving this file into a shared folder, e.g. `app/seed-data/shared`.
Since these are quite short it's not too bad though.

Same for the write dic to csv and make dir functions.

app/seed-data/formula_values/scripts/EFDB_IPCC_Waste.py
```diff
@@ -0,0 +1,624 @@
...
+            "Abatement / Control Technologies" AS control_paractises,
+            "Other properties" AS properties,
+            Value AS emissionsfactor_value,
+            Unit AS emissionsfactor_units,
+            Equation AS ipcc_equation,
+            "IPCC Worksheet" as ipcc_worksheet,
+            "Technical Reference" as technical_reference,
+            "Source of data" as dataset_name,
+            "Data provider" AS data_source
+            FROM ST_READ('/Users/amandaeames/Downloads/EFDB_output.xlsx')
```
lemilonkh, line 92: Replace hardcoded path with either a relative path or at least a constant on the top. This will not work on other machines.

lemilonkh (changes requested):

app/seed-data/formula_values/scripts/utils.py
amanda-eames: This is a good point, maybe we change this over if these get larger?

amanda-eames:
- I need to change global to world that's whats being used in emission factors

lemilonkh (approved):

app/seed-data/formula_values/data_processed/EFDB_2006_IPCC_guidelines/FormulaInputs.csv
```diff
@@ -0,0 +1,237 @@
+gas,parameter_code,parameter_name,methodology,gpc_refno,year,formula_input_value,formula_input_units,formula_name,metadata,region,actor_id,datasource,formulainput_id
```
lemilonkh, line 1: We don't have this table yet. Can you provide a SQL data schema for it so we can add it? Otherwise this will fail to seed I think.

lemilonkh (changes requested):

PR #623, opened 2024-08-06
feat: sector I and III

mfonsecaOEF:


evanp:
@mfonsecaOEF this won't merge because you changed a bunch of unique IDs for direct measure to just "direct measure".

That was a mistake. We need unique IDs for Direct Measure for each and every sector because the labels are different: "Direct measure for rail transport", etc.

thehighestprimenumber:
I'm creating the direct measure myself in the code, so you don't need to worry about the ids.

evanp:
@thehighestprimenumber I don't believe you can implement "Direct measure for X" in a language-independent way.

thehighestprimenumber:
> @thehighestprimenumber I don't believe you can implement "Direct measure for X" in a language-independent way.
@evanp 
I already am doing it, since I didn't have the strings in the json I'm building them myself and they are already being translated as we saw today

evanp:
Awesome. I should have been clearer; I don't think it's a good idea to do something like this:

```
const dm = getString('direct-measure-for');
const t = getString(topicId);

const label = dm + ' ' + t;
```

Assembling strings like this would work for en, es, and maybe pt, but not other languages.

I think you are (were) just building the ID, not trying to build the string from parts. So that's fine!

PR #625, opened 2024-08-06
Changed loading API key during build to loading during runtime

mircorudolph:


app/src/app/api/v0/assistants/route.ts
```diff
@@ -1,8 +1,10 @@
 import { apiHandler } from "@/util/api";
-import { openai } from "@/util/openai";
+import { setupOpenAI } from "@/util/openai";
 import { NextResponse } from "next/server";
 import { Roles } from "@/lib/auth";
 
+const openai = setupOpenAI();
```
lemilonkh, line 6: I think this might have the same issue but let's see 😅

lemilonkh (approved):

PR #626, opened 2024-08-06
moved function 'setupOpenAI' inside api handler

mircorudolph:


lemilonkh (approved):
Maybe remove the commented code passages, otherwise this looks good to me 😊👍🏼

PR #628, opened 2024-08-06
feat(ui): adds optional props to to add activity modal that sets valu…

cephaschapa:
Changes
- Adds optional prop to add activity data modal that sets the activity dropdown value when a suggested activity is selected

### DIFF
--- a/app/src/components/Tabs/Activity/activity-tab.tsx
+++ b/app/src/components/Tabs/Activity/activity-tab.tsx
@@ -492,6 +492,7 @@ const ActivityTab: FC<ActivityTabProps> = ({
         setHasActivityData={setHasActivityData}
         methodology={methodology!}
         inventoryId={inventoryId}
+        selectedActivity="Example Activity"
       />
 
       <ChangeMethodology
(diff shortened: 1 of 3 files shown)


app/src/components/Tabs/Activity/activity-tab.tsx
```diff
@@ -77,8 +83,8 @@ const ActivityTab: FC<ActivityTabProps> = ({
 
   const { methodologies, directMeasure } = getMethodologies();
 
-  const getSuggestedActivities = () => {
-    if (!selectedMethodology) return [];
+  const getSuggestedActivities: () => SuggestedActivity[] = () => {
+    if (!selectedMethodology) return [] as SuggestedActivity[];
```
lemilonkh, line 87:
```suggestion
  const getSuggestedActivities = (): SuggestedActivity[] => {
    if (!selectedMethodology) return [];
```

And you can remove the cast below as well if you define the return type of the lambda I think.

lemilonkh (approved):

PR #633, opened 2024-08-08
Allow creating InventoryValue from POST ActivityValue endpoint

lemilonkh:
- Creates GPCService for GPC-related querying functionality (for now getting the subsector/ subcategory/ sector ID for a given GPC reference number)
- Deprecates a bunch of fields that are now in ActivityValue (to be removed later)

### DIFF
--- a/app/src/app/api/v0/inventory/[inventory]/activity-value/route.ts
+++ b/app/src/app/api/v0/inventory/[inventory]/activity-value/route.ts
@@ -1,4 +1,5 @@
 import CalculationService from "@/backend/CalculationService";
+import GPCService from "@/backend/GPCService";
 import UserService from "@/backend/UserService";
 import { db } from "@/models";
 import type { ActivityValue } from "@/models/ActivityValue";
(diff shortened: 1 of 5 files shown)


app/src/app/api/v0/inventory/[inventory]/activity-value/route.ts
```diff
@@ -18,10 +20,12 @@ export const POST = apiHandler(async (req, { params, session }) => {
 
   const result = await db.sequelize?.transaction(
```
Comment (isaaccodekill) [inline app/src/app/api/v0/inventory/[inventory]/activity-value/route.ts:21]: This transaction is getting really big, A good idea would be to break out this logic into a separate ActivityValue Service

isaaccodekill (approved):

isaaccodekill (approved):

app/src/backend/GPCService.ts
```diff
@@ -0,0 +1,45 @@
...
+    const subSectorId = subcategory?.subsector?.subsectorId;
+    const sectorId = subcategory?.subsector?.sector?.sectorId;
+
+    if (!sectorId || !subSectorId || !subCategoryId) {
+      throw new createHttpError.BadRequest(
+        "Couldn't find sector/ subsector/ subcategory for given GPC reference number",
+      );
+    }
+
+    return { sectorId, subSectorId, subCategoryId };
```
evanp, line 43: One thing I'd want to check is for scope number (1, 2, 3). Sometimes the third element is the scope number; other times it's not. Worth including here!

lemilonkh: Fair, in case this method gets used in other places we can include the scope.

For the current use case we just need the three IDs to create an inventory value.

Since we need to do another join here to get the actual scope number from the Scope table, not sure if we need to do this now for an eventual use in the future...

lemilonkh: We could have a separate method GPCService.getScopeForReferenceNumber, since I don't think we use both our own IDs and GPC scope at the same time usually.

PR #636, opened 2024-08-09
Update PATCH ActivityValue endpoint and test all ActivityValue endpoints

isaaccodekill:
Update the Patch ActivityValue Endpoint with calculations for CO2 emission, with Tests for all activity value endpoints


app/src/app/api/v0/inventory/[inventory]/activity-value/[id]/route.ts
```diff
@@ -2,124 +2,44 @@ import UserService from "@/backend/UserService";
...
   await UserService.findUserInventory(params.inventory, session);
 
-  const activityValue = await db.models.ActivityValue.findOne({
-    where: { id },
+  const result = await ActivityService.updateActivity({
+    id,
+    activityValueParams: {
+      activityData: data.activityData,
+      metadata: data.metadata,
+    },
```
Comment (lemilonkh) [inline app/src/app/api/v0/inventory/[inventory]/activity-value/[id]/route.ts:35]:
```suggestion
    activityValueParams: data,
```

app/src/app/api/v0/inventory/[inventory]/activity-value/route.ts
```diff
@@ -3,7 +3,10 @@ import UserService from "@/backend/UserService";
 import { db } from "@/models";
 import type { InventoryValue } from "@/models/InventoryValue";
 import { apiHandler } from "@/util/api";
-import { createActivityValueRequest } from "@/util/validation";
+import {
+  createActivityValueRequest,
+  updateActivityValueRequest,
```
Comment (lemilonkh) [inline app/src/app/api/v0/inventory/[inventory]/activity-value/route.ts:8]: I don't think you need to import this here, just in the PATCH route file.

app/src/backend/ActivityService.ts
```diff
@@ -21,9 +21,241 @@ import type { Transaction } from "sequelize";
 
 type GasValueInput = Omit<GasValueCreationAttributes, "id"> & {
   emissionsFactor?: Omit<EmissionsFactorAttributes, "id">;
+  id?: string;
```
lemilonkh, line 24:
```suggestion
type GasValueInput = GasValueCreationAttributes & {
  emissionsFactor?: EmissionsFactorAttributes;
```

No need to omit id and then add it again.

lemilonkh: And I guess you might want to add id for the emissions factor so it knows which one to update.
Not sure if id needs to be optional for EmissionsFactor so we can also create new ones or if that always already happened in the POST endpoint and we can just assume they are present in the PATCH endpoint.

app/src/backend/ActivityService.ts
```diff
@@ -21,9 +21,241 @@ import type { Transaction } from "sequelize";
...
+        where: { datasourceId: activityValue.datasourceId },
+        transaction,
+      });
+      if (!dataSource) {
+        throw new createHttpError.NotFound(
+          "Data source for ActivityValue not found",
+        );
+      }
+      Object.assign(dataSource, dataSourceParams);
+      await dataSource?.save({ transaction });
```
lemilonkh, line 48: Sequelize has a `.update()` method you can use here to simplify the logic a bit.
You can just pass the dataSourceParams and the existing fields will stay the same.
There is also `.upsert()` if you're not sure if it exists yet, but I think here you can assume it does.

app/src/backend/ActivityService.ts
```diff
@@ -21,9 +21,241 @@ import type { Transaction } from "sequelize";
...
+      return dataSource.datasourceId;
+    }
+  }
+
+  private static async updateInventory({
+    activityValue,
+    inventoryValueParams,
+  }: {
+    activityValue: ActivityValue;
+    inventoryValueParams?: Omit<InventoryValueAttributes, "id"> | undefined;
```
lemilonkh, line 67:
```suggestion
    inventoryValueParams?: InventoryValueAttributes | undefined;
```
You need the id in order to know which one you need to update.

lemilonkh: I guess you can also get it from the activity value like you do below...

app/src/backend/ActivityService.ts
```diff
@@ -21,9 +21,241 @@ import type { Transaction } from "sequelize";
...
+  }) {
+    let inventoryId = activityValue.inventoryValueId;
```
lemilonkh, line 69: We also have a table called Inventory so best to call this variable `inventoryValueId`

app/src/backend/ActivityService.ts
```diff
@@ -21,9 +21,241 @@ import type { Transaction } from "sequelize";
...
+          ...dataSourceParams,
+          datasourceId: randomUUID(),
+        },
+        { transaction },
+      );
+      return dataSource.datasourceId;
+    }
+  }
+
+  private static async updateInventory({
```
lemilonkh, line 62:
```suggestion
  private static async updateInventoryValue({
```

app/src/backend/ActivityService.ts
```diff
@@ -21,9 +21,241 @@ import type { Transaction } from "sequelize";
...
+    activityValueParams,
+    gasValues,
+    dataSourceParams,
+  }: {
+    id: string;
+    activityValueParams: Omit<ActivityValueAttributes, "id">;
+    inventoryValueId: string | undefined;
+    inventoryValueParams: Omit<InventoryValueAttributes, "id"> | undefined;
+    gasValues: GasValueInput[] | undefined;
+    dataSourceParams: Omit<DataSourceAttributes, "datasourceId"> | undefined;
```
lemilonkh, line 197: I think generally you don't want to omit the IDs here but you might have found ways around that by using the relations...

app/src/backend/ActivityService.ts
```diff
@@ -21,9 +21,241 @@ import type { Transaction } from "sequelize";
...
+            activityValue,
+            inventoryValue.inputMethodology as string,
+          );
+
+        inventoryValue.co2eq = totalCO2e;
+        inventoryValue.co2eqYears = Math.max(0, totalCO2eYears);
+        await inventoryValue.save({ transaction });
+
+        activityValue.co2eq = totalCO2e;
+        activityValue.co2eqYears = totalCO2eYears;
```
lemilonkh, line 244: activityValue needs to be saved as well, otherwise this never makes it to the database...

lemilonkh:
```suggestion
        activityValue.co2eq = totalCO2e;
        activityValue.co2eqYears = totalCO2eYears;
        await activityValue.save({ transaction });
```

lemilonkh (commented):
Great work :raised_hands: 
And thanks for refactoring along the way as well :sparkles:

app/src/backend/ActivityService.ts
isaaccodekill: For the activityValue ID, I prefer to use the ID from the url, and my reasoning for using the relations for the inventoryValueId was to cater for situation where the user doesn't pass the inventoryValueParams as part of the patch body

app/src/models/ActivityValue.ts
isaaccodekill, line 1: the after create hook transforms the stored jsonb data

app/migrations/20240814014055-ActivityValue-backfill-activity-data-jsonb.cjs
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up (queryInterface, Sequelize) {
+    await queryInterface.sequelize.query(`
+      UPDATE "ActivityValue"
+      SET "activity_data_jsonb" = "activity_data"::jsonb
+      WHERE "activity_data_jsonb" IS NULL;
```
lemilonkh, line 9: I think you want IS NOT NULL here? Otherwise you just end up writing the null values to the new column.

app/src/backend/ActivityService.ts
lemilonkh: Original comment has not been resolved

app/src/backend/ActivityService.ts
```diff
@@ -21,9 +21,225 @@ import type { Transaction } from "sequelize";
...
+        `Inventory value ${inventoryValue.id} is missing an input methodology`,
+      );
+    }
+
+    // if inventoryValueParams exist update the inventoryValues
+    if (inventoryValueParams) {
+      const { sectorId, subSectorId, subCategoryId } =
+        await GPCService.getIDsFromReferenceNumber(
+          inventoryValueParams.gpcReferenceNumber!,
+        );
```
lemilonkh, line 101: Not sure if we need to support moving it to a different subsector/ sector etc. after creating it.
Doesn't seem like a requirement from the UI side, so maybe we can avoid the additional query...

app/src/backend/ActivityService.ts
```diff
@@ -81,11 +304,16 @@ export default class ActivityService {
           );
         }
 
+        // activityValueParams.activityData = JSON.stringify(
+        //   activityValueParams.activityData,
+        // ) as unknown as Record<string, any>;
+
```
lemilonkh, line 310: Remove commented code

app/src/models/ActivityValue.ts
```diff
@@ -139,6 +147,13 @@ export class ActivityValue
             fields: [{ name: "id" }],
           },
         ],
+        hooks: {
+          // afterCreate: (attributes) => {
+          //   attributes.activityData = JSON.parse(
+          //     attributes.activityData as unknown as string,
+          //   );
+          // },
+        },
```
lemilonkh, line 156: Remove as it's no longer required

app/migrations/20240814014055-ActivityValue-backfill-activity-data-jsonb.cjs
isaaccodekill: The Is NULL, ensures that we only fill in columns that are null. So we don't override any "activity_data_jsonb" field that has it's own values

lemilonkh: Ah fair, I misunderstood the purpose of the query. Makes sense then :thumbsup::skin-tone-3:

lemilonkh (approved):

PR #646, opened 2024-08-12
use i18n version of the catalogue

thehighestprimenumber:
image

image


app/migrations/20240809141227-create-datasourceI18n.cjs
```diff
@@ -0,0 +1,50 @@
...
+    start_year                 integer,
+    end_year                   integer,
+    methodology_description    jsonb,
+    transformation_description jsonb,
+    dataset_name               jsonb
+);
+
+alter table public."DataSourceI18n"
+    owner to citycatalyst;
+
```
lemilonkh, line 38: Don't include this here. This should be the default, otherwise your DB is misconfigured (Citycatalyst needs to be able to have all rights on schema public in the citycatalyst database.

app/migrations/20240809141227-create-datasourceI18n.cjs
```diff
@@ -0,0 +1,50 @@
+"use strict";
+
+const sql_up = `create table if not exists public."DataSourceI18n"
+(
+    datasource_id uuid not null constraint "DataSourceI18n_pkey" primary key,
+    datasource_name            varchar(255),
```
lemilonkh, line 6: Is this column not translated?

app/migrations/20240809141227-create-datasourceI18n.cjs
```diff
@@ -0,0 +1,50 @@
...
+    "URL"                      varchar(255),
+    dataset_description        jsonb,
+    access_type                varchar(255),
```
lemilonkh, line 9: Let's use `text` over `varchar(255)`. Varchar is annoying to deal with once the values start going over this limit and it's not really that much more performant in modern Postgres.

app/migrations/20240809141227-create-datasourceI18n.cjs
```diff
@@ -0,0 +1,50 @@
...
+    methodology_url            varchar(255),
+    publisher_id uuid constraint "FK_DataSourceI18n.publisher_id" references public."Publisher" on update cascade on delete set null,
+    retrieval_method           varchar(255),
+    api_endpoint               varchar(255),
+    created                    timestamp,
+    last_updated               timestamp,
+    source_type                varchar(255),
+    sector_id uuid constraint "FK_DataSourceI18n.sector_id" references public."Sector" on update cascade on delete cascade,
+    subsector_id uuid constraint "FK_DataSourceI18n.subsector_id" references public."SubSector" on update cascade on delete cascade,
+    subcategory_id uuid constraint "FK_DataSourceI18n.subcategory_id" references public."SubCategory" on update cascade on delete cascade,
```
lemilonkh, line 28: Not sure if we want to cascade here. If these get deleted, something went very wrong as this is our seed data that every inventory references. So in this case let's do `on update cascade on delete set null`, just to make sure we don't need to restore user data from backups.

app/migrations/20240809142521-prefill-datasource.cjs
```diff
@@ -0,0 +1,46 @@
...
+}
+
+/** @type {import("sequelize-cli").Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    const rowsWithUserData = await getRowsWithUserData(queryInterface);
+    const formattedRows = getFormattedRows(rowsWithUserData);
+    return queryInterface.sequelize.transaction(async (transaction) => {
+      for (const row of formattedRows) {
+        await queryInterface.insert(null, "DataSourceI18n", row, { transaction });
```
lemilonkh, line 38: Not sure if we want to save user data in DataSourceI18n? Let's keep it in DataSource and rename that to UserDataSource in a future refactoring (or just move the 2 relevant fields into ActivityValue tbh, that makes saving easier too). But this should be done in a follow-up PR if possible when we clean up the deprecated fields/ tables in the DB.

app/scripts/catalogue-sync.ts
```diff
@@ -68,10 +68,8 @@ async function syncDataCatalogue() {
     });
   }
   const previousUpdate = catalogue.lastUpdate?.getTime() || 0;
-
-  const lastUpdateResponse = await fetch(
-    `${GLOBAL_API_URL}/api/v0/catalogue/last-update`,
-  );
+  const catalogUrl = `${GLOBAL_API_URL}/api/v0/catalogue`;
```
lemilonkh, line 71:
```suggestion
  const catalogUrl = `${GLOBAL_API_URL}/api/v0/catalogue/i18n`;
```
Why are you still using the old route?

app/scripts/catalogue-sync.ts
```diff
@@ -180,7 +180,7 @@ async function syncDataCatalogue() {
    * https://github.com/sequelize/sequelize/issues/13545
    */
   for (const source of sources) {
-    await db.models.DataSource.upsert(source);
+    await db.models.DataSource.upsert(source); // TODO NINA DataSourceI18n
```
lemilonkh, line 183: This would be good to switch before merging :)

app/src/app/[lng]/[inventory]/data/[step]/ActivityDataTab.tsx
```diff
@@ -49,9 +50,9 @@ const activityDataUnits: Record<string, string[]> = {
 ];*/
 
 export function determineEmissionsFactorType(factor: EmissionsFactorData) {
-  let sourceName = factor.dataSources
-    ? factor.dataSources[0].datasetName || "Unknown data source"
-    : "Unknown data source";
+  const sourceName =
+    getTranslationFromDict(factor.dataSources?.[0]?.datasetName || {}) ||
+    "Unknown data source";
```
Comment (lemilonkh) [inline app/src/app/[lng]/[inventory]/data/[step]/ActivityDataTab.tsx:55]: Use TypeScript's `??` default operator if possible as there are less weird type coercions with it

app/src/components/Modals/add-activity-modal.tsx
```diff
@@ -79,19 +80,6 @@ export type Inputs = {
...
-    : "Unknown data source";
-  if (sourceName.includes("IPCC") && sourceName.includes("US")) {
-    return "National (US)";
-  } else if (sourceName.includes("IPCC")) {
-    return "IPCC";
-  }
-
-  return sourceName;
-}
-
```
lemilonkh, line 94: I think we might still need this method? Although we are using the names of the data sources now, so maybe that helps...

app/src/i18n/index.ts
```diff
@@ -32,3 +33,10 @@ export async function useTranslation(
     i18n: i18nextInstance,
   };
 }
+
+export const getTranslationFromDict = (
+  translations: Record<string, string> | string | undefined,
+): string | undefined => {
+  const lang = i18next.language;
```
lemilonkh, line 40: I think this means that it won't update on the fly when the language changes? So maybe better to `useEffect` that registers this callback: https://www.i18next.com/overview/api#onlanguagechanged
And then unregisters it in the returned function.
If it automatically re-renders when you select a different language in the language dropdown in the navigation bar then this is fine :sparkles:

app/src/models/DataSourceI18n.ts
```diff
@@ -0,0 +1,929 @@
...
+        subcategoryId: {
+          type: DataTypes.UUID,
+          allowNull: true,
+          references: {
+            model: "SubCategory",
+            key: "subcategory_id",
+          },
+          field: "subcategory_id",
+        },
+      },
```
lemilonkh, line 911: Need to have `DataTypes.JSONB` column `type` here for the relevant columns

lemilonkh: Also using TEXT over STRING(255) as mentioned on the migration

app/src/util/helpers.ts
```diff
@@ -1,3 +1,25 @@
+export const getTranslationFromDictionary = (
+  translations: Record<string, string> | string | undefined,
+  lang?: string,
```
lemilonkh, line 3: Usually abbreviated as `lng` in the codebase currently, but lang works too I suppose...

app/migrations/20240809141227-create-datasourceI18n.cjs
thehighestprimenumber: it isn't 
image

app/src/components/Modals/add-activity-modal.tsx
thehighestprimenumber: it was duplicated code, we have the exact same logic in determineEmissionsFactorType in ActivityDataTab.tsx

app/src/i18n/index.ts
thehighestprimenumber: I just checked, in automatically re-renders

app/migrations/20240809142521-prefill-datasource.cjs
thehighestprimenumber: Created ticket here:
https://openearth.atlassian.net/browse/OC-1431

app/scripts/catalogue-sync.ts
thehighestprimenumber: we have to keep it without the i18n for the last_updated check. then we add the /i18n when we query for the data.

app/src/models/DataSourceI18n.ts
thehighestprimenumber: > Need to have `DataTypes.JSONB` column `type` here for the relevant columns

Is there somewhere I'm missing? the only fields that come translated are
transformation_description, dataset_description, methodology_description, dataset_name

lemilonkh: Then not, I suppose it's mostly proper names, which usually don't need to be translated.

app/migrations/20240809141227-create-datasourceI18n.cjs
lemilonkh: See this snippet from schema/reset.sql in the repo:
```postgresql
GRANT ALL ON SCHEMA public TO postgres;
GRANT ALL ON SCHEMA public TO public;
```

lemilonkh: This is done automatically when you use the docker script to run the DB instead of running a system-wide one, where you need to run these to get full access to the DB with the citycatalyst user.

lemilonkh (commented):
LGTM now :+1: Let's have @evanp take a final look and then merge :sparkles:

app/migrations/20240809142521-prefill-datasource.cjs
evanp: i think the idea was to leave ds alone. if we want to add uds, lets create that and copy

app/migrations/20240812130411-adjust-fk.cjs
evanp, line 1: should we use sqlize method `removeConstraint`

evanp (approved):
looks good to me. id recmd using sqlz mthds but not req'd

evanp (approved):

PR #665, opened 2024-08-20
remove transaction

thehighestprimenumber:
So that if one thing fails, the rest doesn't


lemilonkh:
I'm not sure if that is the desired behavior? I think if one of those doesn't succeed we want the migration to fail. Same as most of our other migrations that have multiple queries, so use a transaction to make sure either all or none of them succeed.

lemilonkh (approved):
Approved so this migration goes through in dev/ qa/ prod

PR #666, opened 2024-08-20
fix: updated logic for US populating with the wrong actor id

amanda-eames:
- updating mapping logic from region to actor_id
- check methodology-name values
- need to add methodology id (will require some changes to datafile strcuture)


amanda-eames:
added methodology_id to the datafile and methodology csv but this break the data seeding script.

thehighestprimenumber (approved):

lemilonkh (approved):

lemilonkh:
Is this ready to be merged or do we need to wait for #662 to be finished and merged?
@amanda-eames does it break the current version of our seeders?
If so, it's part of that PR so we could merge both at the same time by closing this and merging the other one once it's ready?

PR #668, opened 2024-08-21
e2e: Manual input test

isaaccodekill:
Includes playwright tests for direct-measure in every sector

image


lemilonkh:
The new integration tests seem to be quite slow from your screenshot. Can you try to parallelize them as suggested in the yellow notice next?

Otherwise I guess we can start running our Playwright tests as a separate step in parallel to our regular tests...

app/src/app/[lng]/auth/login/page.tsx
```diff
@@ -74,6 +74,7 @@ export default function Login({
 
       if (!res?.error) {
         router.push(callbackUrl);
+        // what happens if there is no callback url ?
```
Comment (lemilonkh) [inline app/src/app/[lng]/auth/login/page.tsx:77]: Correct it should have a default value of e.g. /

app/src/util/check-user-session.ts
```diff
@@ -10,4 +10,6 @@ export const CheckUserSession = () => {
       router.push("/auth/login");
     },
   });
+
+  console.log("user session", data);
```
lemilonkh, line 14: Remove logging once you're done

app/e2e/login.spec.ts
```diff
@@ -50,7 +26,7 @@ test.describe("Login page", () => {
     // TODO how to ensure that session route was called?
     await page.waitForResponse("/api/auth/session");
 
-    await expect(page).toHaveURL("/en/onboarding/");
+    // await expect(page).toHaveURL("/en/onboarding/");
```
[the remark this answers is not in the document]
isaaccodekill: Yeah this was failing because at the points the tests run there is usually a default inventory.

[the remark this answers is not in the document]
isaaccodekill: modified this to `    await expect(page).not.toHaveURL("/en/auth/login/");` this way we test that the login action actually moved the user to within the application and not the auth screen anymore

app/src/app/[lng]/auth/login/page.tsx
Comment (isaaccodekill) [reply app/src/app/[lng]/auth/login/page.tsx]: I think this is already the default behaviour. I think this was a stray comment I left when trying to debug some auth issues I was having locally. I'll mark as resolved

isaaccodekill:
> The new integration tests seem to be quite slow from your screenshot. Can you try to parallelize them as suggested in the yellow notice next?
> 
> Otherwise I guess we can start running our Playwright tests as a separate step in parallel to our regular tests...

Parallelizing them cause the tests to break in headless mode.  If it is a hard requirement to make them run in parallel we can create a separate ticket to accomplish that

Let me know what you think

lemilonkh (approved):
Thanks :sparkles:

PR #669, opened 2024-08-23
fix: adds target property to open page gpc protocal page in different tab

cephaschapa:
Changes
- Adds target prop that opens tab in different tab when clicked.


app/src/components/Tabs/Activity/activity-tab.tsx
```diff
@@ -542,6 +542,8 @@ const ActivityTab: FC<ActivityTabProps> = ({
                           color="content.link"
                           fontWeight="bold"
                           textDecoration="underline"
+                          target="_blank"
+                          rel="noreferrer"
```
lemilonkh, line 546: I think adding noopener helps here for security

lemilonkh (approved):

PR #680, opened 2024-08-26
[ON-2101] Get scopes from db

thehighestprimenumber:


app/src/app/api/v0/sector/[sectorId]/required-scopes/route.ts
```diff
@@ -0,0 +1,29 @@
+import { db } from "@/models";
+import { apiHandler } from "@/util/api";
+import createHttpError from "http-errors";
+import { NextRequest, NextResponse } from "next/server";
+import { QueryTypes } from "sequelize";
+
+const getRequiredScopes = async (sectorId: string) => {
```
Comment (lemilonkh) [inline app/src/app/api/v0/sector/[sectorId]/required-scopes/route.ts:7]: This method could be moved to GPCService (as a static class method), as that's part of that domain.

lemilonkh (approved):
Thanks 🙌🏼✨

PR #681, opened 2024-08-26
feat: added new route for modelled global-api structure

amanda-eames:
added tables for the modeled schema
added a new emissions route that give the total emissions and the emissions by subcategory
added a parameter for the global warming potential ar2, ..., ar6 this can be used based a users preference


lemilonkh (approved):
Amazing :sparkles:

PR #684, opened 2024-08-29
Manual input validation

isaaccodekill:
This PR contains:
-  Manual input validation implementations
- Ui updates to make the Frontend match the design more closely
- Implement Dynamic fields for the activity value form


app/src/backend/ManualnputValidationService.ts
```diff
@@ -0,0 +1,251 @@
...
+      );
+
+      const errorBody = {
+        code: ManualInputValidationErrorCodes.UNIQUE_BY_CONFLICT,
+        targetFields: duplicateFields,
+      };
+
+      throw new ManualInputValidationError(errorBody);
+    }
+    // check if the uniqueBy fields are unique
```
lemilonkh, line 196: Is there something missing here or why is the comment here?

app/src/lib/custom-errors.ts/manual-input-error.ts
lemilonkh, line 1: Why is the folder here named custom-error.ts? Can we just name it custom-errors instead to prevent confusion?

app/src/lib/custom-errors.ts/manual-input-error.ts
```diff
@@ -0,0 +1,25 @@
+import createHttpError from "http-errors";
```
lemilonkh, line 1: I don't think this import is being used?

app/src/components/Modals/activity-modal/activity-form-modal.tsx
```diff
@@ -253,30 +242,50 @@ const AddActivityModal: FC<AddActivityModalProps> = ({
...
+        activityDataAmount: 0,
+        activityDataUnit: null,
+        emissionFactorType: "",
+        totalFuelConsumption: "",
+        totalFuelConsumptionUnits: "",
+        co2EmissionFactorUnit: "",
+        n2oEmissionFactorUnit: "",
+        ch4EmissionFactorUnit: "",
+      },
+    });
```
lemilonkh, line 279: It's helpful to store this as a "default value" object next to the type definition for the form state. Makes it easier if you need to reset the form from different places in the code.

app/src/components/Modals/activity-modal/activity-form-modal.tsx
```diff
@@ -253,30 +242,50 @@ const AddActivityModal: FC<AddActivityModalProps> = ({
...
   return (
     <>
       <Modal
         blockScrollOnMount={false}
         isOpen={isOpen}
         onClose={() => {
-          onClose();
-          resetSelectedActivityValue();
+          closeModalFunc();
         }}
```
lemilonkh, line 289: `onClose={closeModalFunc}`

app/src/components/building-select-input.tsx
```diff
@@ -44,8 +49,8 @@ const BuildingTypeSelectInput: FC<BuildingTypeSelectInputProps> = ({
         borderRadius="4px"
         borderWidth={errors?.[activity] ? "1px" : 0}
         border="inputBox"
-        borderColor={errors?.[activity] ? "sentiment.negativeDefault" : ""}
-        background={errors?.[activity] ? "sentiment.negativeOverlay" : ""}
+        borderColor={error ? "sentiment.negativeDefault" : ""}
+        background={error ? "sentiment.negativeOverlay" : ""}
```
lemilonkh, line 53: Why switch to just error here? This would mean it's no longer specific to the activity you're in? If we need to make this change, also change it on the `borderWidth`, otherwise that could get inconsistent.

app/src/hooks/activity-value-form/use-activity-form.ts
```diff
@@ -0,0 +1,78 @@
...
+          N2OEmissionFactor: 0,
+          activityDataAmount: 0,
+          activityDataUnit: null,
+          emissionFactorType: "",
+          totalFuelConsumption: "",
+          totalFuelConsumptionUnits: "",
+          co2EmissionFactorUnit: "",
+          n2oEmissionFactorUnit: "",
+          ch4EmissionFactorUnit: "",
+        },
```
lemilonkh, line 62: If you store this default value object for the form next to the type definition, you can use is above as well to set the defaults for things you're not overwriting.

app/src/hooks/activity-value-form/use-activity-form.ts
```diff
@@ -0,0 +1,78 @@
...
+        activity: {
+          ...targetActivityValue.activityData,
+          dataQuality: targetActivityValue?.dataSource?.dataQuality,
+          sourceReference: targetActivityValue?.dataSource?.notes,
+          CH4EmissionFactor: targetActivityValue?.activityData?.ch4_amount,
+          CO2EmissionFactor: targetActivityValue?.activityData?.co2_amount,
+          N2OEmissionFactor: targetActivityValue?.activityData?.n2o_amount,
+          activityDataAmount: 0,
+          activityDataUnit: null,
+          emissionFactorType: "",
```
lemilonkh, line 36: This needs to be loaded from the ActivityValue/ GasValue as well. The type for the drop down in this case is the `name` of the data source associated with the EmissionsFactor (as you can see in how the data for the dropdown is created).

All of the other fields need to be loaded here as well, not just the ones above.

app/src/hooks/activity-value-form/use-activity-form.ts
```diff
@@ -0,0 +1,78 @@
...
+  useEffect(() => {
+    if (targetActivityValue) {
+      reset({
+        activity: {
+          ...targetActivityValue.activityData,
+          dataQuality: targetActivityValue?.dataSource?.dataQuality,
+          sourceReference: targetActivityValue?.dataSource?.notes,
+          CH4EmissionFactor: targetActivityValue?.activityData?.ch4_amount,
+          CO2EmissionFactor: targetActivityValue?.activityData?.co2_amount,
+          N2OEmissionFactor: targetActivityValue?.activityData?.n2o_amount,
```
lemilonkh, line 33: This seems to be wrong. These need to come from the `emissionsPerActivity` property of the `EmissionsFactor` associated with each `GasValue`.

The fields you're using here are the manual input ones, which don't apply for activity based reporting.

lemilonkh (commented):
Great work :raised_hands:

app/src/backend/ActivityService.ts
```diff
@@ -206,16 +214,30 @@ export default class ActivityService {
...
-            updatedActivityValue,
+            activityValue,
             inventoryValue.inputMethodology as string,
           );
 
-        inventoryValue.co2eq = totalCO2e;
-        inventoryValue.co2eqYears = Math.max(0, totalCO2eYears);
+        const currentCO2e =
+          BigInt(inventoryValue.co2eq as bigint) -
+            BigInt(activityValue.co2eq as bigint) ?? 0n;
```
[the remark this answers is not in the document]
isaaccodekill: This is because the InventoryValue.co2eq can be undefined. So to properly use the BigInt Constructor I need to cast it

app/src/app/api/v0/inventory/[inventory]/value/subsector/[subsector]/route.ts
```diff
@@ -0,0 +1,20 @@
+import { apiHandler } from "@/util/api";
+import { db } from "@/models";
+import UserService from "@/backend/UserService";
+import { NextResponse } from "next/server";
+
+export const GET = apiHandler(async (_req, { params, session }) => {
```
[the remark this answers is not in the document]
Comment (isaaccodekill) [reply app/src/app/api/v0/inventory/[inventory]/value/subsector/[subsector]/route.ts:6]: Done!

app/src/components/Modals/activity-modal/activity-form-modal.tsx
isaaccodekill: Smart idea

.DS_Store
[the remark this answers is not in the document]
isaaccodekill: the rule exists in the gitignore and I can seem to find this file locally

app/src/hooks/activity-value-form/use-activity-form.ts
isaaccodekill: So these values are currently not being stored in the DB, I've opted to place them in the metadata column for now

app/src/components/Modals/activity-modal/activity-form-modal.tsx
```diff
@@ -168,13 +171,18 @@ const AddActivityModal: FC<AddActivityModalProps> = ({
...
-      activityData: {
-        co2_amount: gasValues[1].factor,
-        ch4_amount: gasValues[0].factor,
-        n2o_amount: gasValues[2].factor,
-        ...values,
+      activityData: methodology?.id.includes("direct-measure")
+        ? {
+            co2_amount: gasValues[1].factor,
+            ch4_amount: gasValues[0].factor,
+            n2o_amount: gasValues[2].factor,
```
lemilonkh, line 178: Why are we saving emissions factors as gas measurements?
I think there is a confusion here of emissions factors vs. directly measured total gas amounts (both of these are present for CO2, CH4, N2O etc., but they serve very different purposes in the calculations.

app/src/components/Modals/activity-modal/activity-form-modal.tsx
```diff
@@ -168,13 +171,18 @@ const AddActivityModal: FC<AddActivityModalProps> = ({
...
+            ...values,
+          }
+        : { ...values },
+      metadata: {
+        emissionFactorType: activity.emissionFactorType,
+        totalFuelConsumption: activity.totalFuelConsumption,
```
lemilonkh, line 184: `totalFuelConsumption` should be in the activity data, not metadata. Emissions factor type makes sense though, since it's not directly used in the calculation (only used to restore form state when editing, since choosing an emissions factor means its values will be loaded into the 3 emissions factor form inputs.
I think these field names should all be loaded from the manual input hierarchy JSON file?
Because not every activity has totalFuelConsumption, only some methodologies use this.

app/src/components/Modals/activity-modal/activity-modal-body.tsx
```diff
@@ -278,6 +278,7 @@ const ActivityModalBody = ({
               <FormControl
                 isInvalid={!!resolve(prefix + "activityDataAmount", errors)}
               >
+                {/*TODO dynamically render required fields*/}
```
lemilonkh, line 281: This is a big and quite important task. Please create a Jira ticket for the next sprint with highest priority so we can finish this properly.

app/src/components/Modals/activity-modal/activity-modal-body.tsx
```diff
@@ -448,12 +450,13 @@ const ActivityModalBody = ({
...
                     color="content.tertiary"
                     h="48px"
+                    fontSize="14px"
                     shadow="1dp"
                     pos="relative"
                     zIndex={10}
                     {...register("activity.co2EmissionFactorUnit")}
                   >
-                    {""}
+                    C02/{""}
```
lemilonkh, line 459: Why the empty string here? Is there a unit missing here?

app/src/components/Modals/activity-modal/activity-modal-body.tsx
```diff
@@ -540,12 +544,14 @@ const ActivityModalBody = ({
...
                     h="48px"
+                    fontSize="14px"
                     shadow="1dp"
                     pos="relative"
                     zIndex={10}
                     {...register("activity.ch4EmissionFactorUnit")}
                   >
-                    {""}
+                    CH4/{""}
+                    {/*  TODO input the right units*/}
```
lemilonkh, line 554: You can get the units from the manual input hierarchy file. Please also create a Jira ticket with highest priority for this.

app/src/hooks/activity-value-form/use-activity-form.ts
```diff
@@ -4,12 +4,44 @@ import { ActivityValue } from "@/models/ActivityValue";
...
+    CO2EmissionFactor: 0,
+    N2OEmissionFactor: 0,
+    emissionFactorType: "",
+    totalFuelConsumption: "",
+    totalFuelConsumptionUnits: "",
+    co2EmissionFactorUnit: "",
+    n2oEmissionFactorUnit: "",
+    ch4EmissionFactorUnit: "",
+  };
+};
```
lemilonkh, line 25: This could have just been an object that you spread and overwrite the default values that way, but it's ok.

app/src/hooks/activity-value-form/use-activity-form.ts
```diff
@@ -4,12 +4,44 @@ import { ActivityValue } from "@/models/ActivityValue";
...
+
+const extractGasAmount = (gas: string, activity: ActivityValue) => {
```
lemilonkh, line 27: This should be named `extractEmissionsFactor` or rather `findEmissionsFactor`

app/src/hooks/activity-value-form/use-activity-form.ts
```diff
@@ -28,41 +60,39 @@ const useActivityForm = ({
...
+            methodologyName === "direct-measure"
+              ? targetActivityValue?.activityData?.co2_amount
+              : extractGasAmount("CO2", targetActivityValue).amount,
+          N2OEmissionFactor:
+            methodologyName === "direct-measure"
+              ? targetActivityValue?.activityData?.n2o_amount
+              : extractGasAmount("N2O", targetActivityValue).amount,
+          emissionFactorType: targetActivityValue.metadata?.emissionFactorType, // TODO confirm the source of this value
+          totalFuelConsumption:
+            targetActivityValue?.metadata?.totalFuelConsumption, // TODO confirm the source of this value
```
lemilonkh, line 77: This is an input provided by the user (input needs to be generated from the manual input hierarchy data for the chosen methodology).
I would put this into `activityData.totalFuelConsumption` instead of in `metadata` as it is used for calculations. Generally everything used in the formulas should be in `activityData`, everything that's exclusively used to restore the state of forms or other UI elements etc. should be in `metadata`.

lemilonkh: This means having the emissionsFactorType in metadata as shown above is correct :+1:

lemilonkh (approved):
Great, thanks :) Some feedback about where data is stored and emissions factor values vs. direct measure gas amounts.

app/src/components/Modals/activity-modal/activity-modal-body.tsx
isaaccodekill: Done!

app/src/components/Modals/activity-modal/activity-form-modal.tsx
isaaccodekill: This falls under this ticket: https://openearth.atlassian.net/browse/ON-2331?atlOrigin=eyJpIjoiMjdhMDdmNWNlMTRlNDFlZjkwNDM3ZGJmMjQzYzhmMjkiLCJwIjoiaiJ9

PR #685, opened 2024-08-30
feat(api): implement methodology formulas for methane commitment and VKT-1

lemilonkh:
So far, this contains these formulas:
- methane-commitment
- VKT-1 (also referred to as `induced-activity-1`)


app/src/backend/CalculationService.ts
isaaccodekill, line 1: This looks really good to me.

app/src/backend/CalculationService.ts
```diff
@@ -73,70 +81,193 @@ export default class CalculationService {
...
+    }
+
     return {
       totalCO2e,
       totalCO2eYears,
       gases,
     };
   }
+
+  static handleVkt1Formula(
```
isaaccodekill, line 136: Can we move the formulas to another class and have the calculation service depend on this class. We will be implementing a lot of forumulas in the coming weeks, which means the file will get really large

app/src/backend/CalculationService.ts
```diff
@@ -73,70 +81,193 @@ export default class CalculationService {
...
+    activityValue: ActivityValue,
+    gasToCO2Eqs: GasToCO2Eq[],
+  ): Gas[] {
+    const data = activityValue.activityData;
+    if (!data) {
+      throw new createHttpError.BadRequest(
+        "Activity has no data associated, so it can't use the formula",
```
thehighestprimenumber, line 143: how will this message be displayed in the frontend? I'm thinking we need an error code so that it can be translated. Also, we might want a native speaker's POV because to me the sentence sounds a little weird... maybe "No data for this activity, formula can't be applied"

app/src/backend/CalculationService.ts
```diff
@@ -73,70 +81,193 @@ export default class CalculationService {
...
+      if (!gasValue.gas) {
+        throw new createHttpError.BadRequest(
+          "Activity has a GasValue with no `gas` name",
+        );
+      }
+      const emissionsFactor = gasValue.emissionsFactor;
+      const emissions =
+        data["activity-value"] *
+        data["intensity"] *
+        (emissionsFactor.emissionsPerActivity ?? 0); // TODO throw BadRequest error if no EF present?
```
thehighestprimenumber, line 157: ticket?

app/src/backend/CalculationService.ts
```diff
@@ -73,70 +81,193 @@ export default class CalculationService {
...
+    activityValue: ActivityValue,
+    gasToCO2Eqs: GasToCO2Eq[],
+  ): Gas[] {
+    const data = activityValue.activityData;
+    if (!data) {
+      throw new createHttpError.BadRequest(
+        "Activity has no data associated, so it can't use the formula",
+      );
+    }
+    const foodFraction = (data["food-fraction"] || 0) / 100.0;
```
thehighestprimenumber, line 174: a more DRY approach could be

const getFraction = (key) => (data[key] || 0) / 100.0;

[foodFraction, gardenWasteFraction, paperFraction, woodFraction, textilesFraction, industrialWasteFraction] = ["food-fraction", "garden-waste-fraction",
"paper-fraction", "wood-fraction", "textiles-fraction", "industrial-waste-fraction"].map(getFraction)

app/src/backend/CalculationService.ts
```diff
@@ -73,70 +81,193 @@ export default class CalculationService {
...
+
+    const recoveredMethaneFraction = data["recovered-methane-fraction"] || 0;
+    const oxidationFactor = data["landfill-type"].startsWith("managed-well")
+      ? 0.1
+      : 0;
+    const totalSolidWaste = data["total-solid-waste"] || 0;
+
+    // Degradable organic carbon in year of deposition, fraction (tonnes C/tonnes waste)
+    const degradableOrganicCarbon =
+      0.15 * foodFraction +
```
thehighestprimenumber, line 193: maybe we should store these values as constants to avoid "magic numbers"

app/src/backend/CalculationService.ts
```diff
@@ -73,70 +81,193 @@ export default class CalculationService {
...
+    return [{ gas: "CH4", amount: BigInt(ch4Emissions) }];
+  }
+
+  private static handleActivityAmountTimesEmissionsFactorFormula(
+    activityValue: ActivityValue,
+    gasToCO2Eqs: GasToCO2Eq[],
+  ): Gas[] {
+    let totalCO2e = 0n;
+    let totalCO2eYears = 0;
+    // TODO add actvityAmount column to ActivityValue
```
thehighestprimenumber, line 227: ticket for these?

thehighestprimenumber (approved):
Approving with some comments, mainly:
* saving hardcoded values to constants
* handling translation of error codes
* creating tickets for pending TODOs

Also, I think this logic deserves a couple of tests

app/src/backend/CalculationService.ts
lemilonkh: Fair, that's why I originally created the CalculationService but yes it makes sense to split the public methods from the private implementation details. So I guess we can either have a FormulaService, or we can just have a formulas.ts where these are exported methods?

app/src/backend/CalculationService.ts
```diff
@@ -73,70 +81,193 @@ export default class CalculationService {
...
-          return { gas: gasValue.gas!, amount };
-        });
         break;
+      case "methane-commitment":
+        gases = CalculationService.handleMethaneCommitmentFormula(
+          activityValue,
+          gasToCO2Eqs,
+        );
+        break;
+      case "induced-activity-1": // TODO or VKT-1? What is it in the hierarchy.json file?
```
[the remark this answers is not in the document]
lemilonkh: Needs to be resolved on this PR imo as it doesn't work otherwise.

app/src/backend/CalculationService.ts
lemilonkh: Agreed. This is the case for our whole backend, so I'll create a ticket for this so we can apply i18n to backend error messages and display them in the frontend as toasts.

Currently the frontend is handling these errors and then creating custom toast messages from the designs that are translatable.

app/src/backend/CalculationService.ts
```diff
@@ -73,70 +81,193 @@ export default class CalculationService {
     const gasToCO2Eqs = await db.models.GasToCO2Eq.findAll();
     let totalCO2e = 0n;
     let totalCO2eYears = 0;
-    let gases: { gas: string; amount: bigint }[] = [];
+    let gases: Gas[] = [];
+
     switch (formula) {
+      // TODO use Record<string, (activityValue, gasToCO2Eqs) => FormulaResult> for this? To avoid adding new code here for each new formula...
+      // basically like a function pointer table in C++...
       case "direct-measure":
```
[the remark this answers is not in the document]
lemilonkh: This will be resolved during the other refactoring (as they will be the keys of a constant object, which we can then also export as constants if we want.

app/src/backend/CalculationService.ts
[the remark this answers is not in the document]
lemilonkh: This name has been communicated to the data team.

app/src/backend/CalculationService.ts
lemilonkh: Created ticket for this concerning the whole backend.

PR #690, opened 2024-09-03
Saving and Loading for activity based methodologies

isaaccodekill:
Depends on this pr: 

https://github.com/user-attachments/assets/f08fac07-c940-4484-8cf2-d480ba5432dc

https://github.com/Open-Earth-Foundation/CityCatalyst/pull/684


.DS_Store
lemilonkh, line 1: Delete this file.

app/package.json
```diff
@@ -56,6 +56,7 @@
     "ai": "^3.2.35",
     "autoprefixer": "10.4.16",
     "bcrypt": "^5.1.1",
+    "chakra-react-select": "^4.9.1",
```
lemilonkh, line 59: We had a component like this before on the file upload form. Can you check if you can reuse the same thing as there instead of introducing this dependency?

app/src/components/Modals/activity-modal/activity-form-modal.tsx
```diff
@@ -182,21 +157,43 @@ const AddActivityModal: FC<AddActivityModalProps> = ({
...
     const gasValues = extractGasesAndUnits(activity);
+
+    // extract field values
+    const values: Record<string, any> = {};
+    fields?.forEach((field) => {
+      if (field.id in activity) {
+        values[field.id] = (activity as any)[field.id];
+      }
+      if (field.units) {
+        values[`${field.id}unit`] = (activity as any)[`${field.id}unit`];
```
lemilonkh, line 168:
```suggestion
        values[`${field.id}Unit`] = (activity as any)[`${field.id}Unit`];
```
Would be nicer for readability, so it's properly camelCased.

app/src/components/Modals/activity-modal/activity-form-modal.tsx
```diff
@@ -182,21 +157,43 @@ const AddActivityModal: FC<AddActivityModalProps> = ({
...
-      inventoryValue: {
-        inputMethodology: getInputMethodology(methodology?.id), // extract methodology name
-        gpcReferenceNumber: referenceNumber,
-        unavailableReason: "",
-        unavailableExplanation: "",
+      activityData: methodology?.id.includes("direct-measure")
+        ? {
+            co2_amount: gasValues[1].factor,
+            ch4_amount: gasValues[0].factor,
+            n2o_amount: gasValues[2].factor,
```
lemilonkh, line 178: Emissions factors are not direct measure gas values, see feedback on #648

app/src/components/Modals/activity-modal/activity-modal-body.tsx
```diff
@@ -570,7 +723,7 @@ const ActivityModalBody = ({
               borderColor: "content.link",
             }}
             placeholder={t("source-reference-placeholder")}
-            {...register("activity.sourceReference", {
+            {...register(`activity.sourceReference` as any, {
```
lemilonkh, line 726: Why as any? This takes a string.

app/tests/api/activity_value.test.ts
```diff
@@ -478,6 +486,5 @@ describe("Activity Value API", () => {
 
     const { data } = await res3.json();
     assert.equal(res3.status, 200);
-    assert.equal(data.deletedCount, 3);
```
lemilonkh, line 481: Why did you need to remove this assertion? It should still be correct, right?

lemilonkh (approved):

app/package.json
isaaccodekill: This has been changed in a follow up pr

app/src/components/Modals/activity-modal/activity-form-modal.tsx
isaaccodekill: Created a PR for this

app/src/components/Modals/activity-modal/activity-form-modal.tsx
isaaccodekill: So this might be a bit complex to change because it refers to a field in the form reused in multiple places, so I'll improve this in the next pr

app/tests/api/activity_value.test.ts
isaaccodekill: This tests has to be re-modelled, I plan to rewrite them as they are flaky

PR #691, opened 2024-09-03
Feature/on 2136

mircorudolph:
added a timeout function to `handleRequiresAction` that returns a default tooloutput in case there are issues connecting to the internal apis.


lemilonkh (approved):
Thanks :sparkles:

PR #694, opened 2024-09-04
fix(api): gas formula calculation error

isaaccodekill:
Gas values were not being passed to the the calculation service formulas

### DIFF
--- a/app/src/backend/formulas.ts
+++ b/app/src/backend/formulas.ts
@@ -1,6 +1,8 @@
 import type { ActivityValue } from "@/models/ActivityValue";
 import type { Gas } from "./CalculationService";
 import createHttpError from "http-errors";
+import { GasValueCreationAttributes } from "@/models/GasValue";
+import { EmissionsFactorAttributes } from "@/models/EmissionsFactor";
 
 const GAS_NAMES = ["CO2", "N2O", "CH4"];
 const METHANE_CORRECTION_FACTORS: Record<string, number> = {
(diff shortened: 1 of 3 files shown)


app/src/backend/formulas.ts
```diff
@@ -133,7 +147,8 @@ export function handleActivityAmountTimesEmissionsFactorFormula(
     }
     if (emissionsFactor.emissionsPerActivity == null) {
       throw new createHttpError.BadRequest(
-        `Emissions factor ${emissionsFactor.id} has no emissions per activity`,
+        // TODO resolve type issues with extracting id
+        `Emissions factor ${emissionsFactor?.gas} has no emissions per activity`,
```
lemilonkh, line 151:
```suggestion
        `Emissions factor for ${emissionsFactor?.gas} has no emissions per activity`,
```

app/src/backend/formulas.ts
```diff
@@ -37,7 +46,7 @@ export function handleVkt1Formula(activityValue: ActivityValue): Gas[] {
     const emissionsFactor = gasValue.emissionsFactor;
     if (emissionsFactor?.emissionsPerActivity == null) {
       throw new createHttpError.BadRequest(
-        `Emissions factor ${emissionsFactor.id} has no emissions per activity`,
+        `Emissions factor ${emissionsFactor?.gas} has no emissions per activity`,
```
lemilonkh, line 49:
```suggestion
        `Emissions factor for ${emissionsFactor?.gas} has no emissions per activity`,
```

lemilonkh (approved):
This should be refactored later to use and require a JOIN/ include for GasValue and EmissionsFactor in the ActivityValue query for the data that is passed into the formulas.

PR #696, opened 2024-09-05
feat(seed): ON-2204 Add emissions factors for transportation sector

mfonsecaOEF:
this PR contains:
- raw EF data
- transformation script
- processed EF data


app/seed-data/emissions_factors/data_processed/ghgprotocol/Publisher.csv
```diff
@@ -0,0 +1,2 @@
+name,URL,publisher_id
+GPC,https://ghgprotocol.org/,13044ae6-f967-30e8-b722-82ec6ad0df3c
```
lemilonkh, line 2: Do we need this for every data set from the GPC? Because I think we only need one GPC. If this has the same ID as the other GPC entries from the other emissions factor datasets etc., then this is fine.

lemilonkh (approved):

lemilonkh (commented):
Thanks ✨

app/seed-data/emissions_factors/data_processed/ghgprotocol/Publisher.csv
mfonsecaOEF: This publisher is GPC, the other one was IPCC

evanp:
Why is this here instead of the data repository?

lemilonkh:
@evanp because these are emissions factor seed files for CityCatalyst

PR #697, opened 2024-09-07
[on-2352] subsector routing bug

thehighestprimenumber:
STR:
Select Transportation or Waste, then click on a subsector. 
Before: Infinite loading and the url always contained in '/data/1/'
After: Correct routing (url contains  /data/2 or /data/3, depending on the sector), and the page is displayed correctly.


lemilonkh (approved):

lemilonkh (commented):
Great spot 👍 Thanks ✨

PR #706, opened 2024-09-10
Move global API specific ignore settings into global-api/.gitignore

lemilonkh:
Also adds .DS_Store and swap files to global .gitignore


amanda-eames (approved):
Looks good!

PR #708, opened 2024-09-10
feat(script): add ability to skip timestamp check for sync-catalogue script

lemilonkh:
Use `SKIP_TIMESTAMP_CHECK` env var with value `"true"`

E.g. `SKIP_TIMESTAMP_CHECK=true npm run sync-catalogue`


isaaccodekill (approved):
Looks good

PR #711, opened 2024-09-11
feat: add top emissions by subsector

thehighestprimenumber:
image


dinoespinozag:
- The distance between the tabs and the title “Inventory results” is not correct. In the design we have a gap of 48px between each section. Here they are too close together. 
- The color of the title “overview” should be color: var(--content-link, #2351DC);
- The chips the “Year” and “Inventory format” do not have a correct distance between icon and text. It should be 8px and attach and image. 
- 
image

- These chips should not have all wording in uppercase. And they should use the same font format: color: var(--content-secondary, #232640);
/* Label/medium */
font-family: Poppins;
font-size: 12px;
font-style: normal;
font-weight: 500;
line-height: 16px; /* 133.333% */
letter-spacing: 0.5px;

- The comments of the chips apply also for the chips of the sector emissions progress bar.
image


- The Total emission title is repeated. The right component should read “Total sector emissions”. 
- 
image


- The font styles in the table are not according to the design. Use “Poppins” font and the ones we use according to the style library. [](https://www.figma.com/design/3i0o3ytNRy3yhCun1xdCGi/CC-Terra-UI-Guidelines?node-id=2-2&t=GI9C6Mkgf0W4Ma6K-1)

- In the Total emission component, the “?” icon should be in grey #7A7B9A
- 
image


- We do not use title case. Be aware of only having capital letters in the first letter of titles and texts in general.  eg. Instead of "Total Emissions" should be " Total emissions"
 
- The top emission selector by sub-sector and by activity is not according to the design and does not have the right font style and size.

app/src/app/[lng]/[inventory]/InventoryResultTab/TopEmissionsWidget.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+
+const EmissionsTable = ({ topEmissions }: { topEmissions: TopEmission[] }) => {
+  return (
+    <TableContainer my={4}>
+      <Table variant="simple">
+        <Thead>
+          <Tr>
+            <Th sx={{ "font": "bold", color: "black" }}>Subsector</Th>
+            <Th sx={{ "font": "bold", color: "black" }}>Total Emissions (CO2eq)</Th>
```
Comment (isaaccodekill) [inline app/src/app/[lng]/[inventory]/InventoryResultTab/TopEmissionsWidget.tsx:48]: Should there be some translation for Total Emissions ?

app/src/app/[lng]/[inventory]/InventoryResultTab/TopEmissionsWidget.tsx
```diff
@@ -0,0 +1,118 @@
...
+                              inventory
+                            }: {
+  t: Function & TFunction<"translation", undefined>;
+  inventory?: InventoryResponse;
+}) => {
+
+  const { data: results, isLoading: isTopEmissionsResponseLoading } =
+    api.useGetResultsQuery(inventory!.inventoryId!);
+
+  function getPercentagesForProgress(): SegmentedProgressValues[] {
```
Comment (isaaccodekill) [inline app/src/app/[lng]/[inventory]/InventoryResultTab/TopEmissionsWidget.tsx:81]: Can we move this into a useMemo ? to avoid rerenders

```suggestion
  const getPercentagesForProgress = useMemo((): SegmentedProgressValues[] => {
    if (!results) return [];

    // Group results by sector
    const grouped = groupBy(results.totalEmissions?.bySector || [], (e) => e.sectorName) as Record<
      string,
      { co2eq: bigint; percentage: number }[]
    >;

    // Map grouped results to the expected format
    return Object.entries(grouped).map(([name, [{ co2eq, percentage }]]) => ({
      name,
      value: co2eq,
      percentage,
    }));
  }, [results]);
```

app/src/app/api/v0/inventory/[inventory]/results/route.ts
```diff
@@ -0,0 +1,39 @@
+import UserService from "@/backend/UserService";
+import { db } from "@/models";
+import { apiHandler } from "@/util/api";
+import { NextResponse } from "next/server";
+import { getEmissionResults } from "@/backend/ResultsService";
+import sumBy from "lodash/sumBy";
+
+
+export const GET = apiHandler(async (_req, { session, params: { inventory } }) => {
```
Comment (isaaccodekill) [inline app/src/app/api/v0/inventory/[inventory]/results/route.ts:9]: Can we create a ticket to write an API test for this endpoint

isaaccodekill (commented):
Well done!

app/src/components/blue-subtitle.tsx
```diff
@@ -0,0 +1,15 @@
+import { TFunction } from "i18next";
```
lemilonkh, line 1: Use type imports if it's only used for type validation to reduce total bundle size.

lemilonkh (approved):
Thanks ✨

app/src/app/api/v0/inventory/[inventory]/results/route.ts
Comment (thehighestprimenumber) [reply app/src/app/api/v0/inventory/[inventory]/results/route.ts:9]: https://openearth.atlassian.net/browse/ON-2404

PR #723, opened 2024-09-17
feat(ui): create percentage breakdown input for III.1.1 etc.

lemilonkh:
### TL;DR

Added a new percentage breakdown input component for activity forms.

### What changed?

- Introduced a new `PercentageBreakdownInput` component for handling percentage breakdowns in activity forms.
- Updated `activity-form-modal.tsx` and `activity-modal-body.tsx` to incorporate the new input type.
- Added new icons for different waste categories.
- Updated localization strings to include new category labels.
- Modified `useActivityForm` hook to reset form values when methodology changes.

### How to test?

1. Navigate to an activity form that includes a percentage breakdown field.
2. Click on the input to open the popover with category breakdowns.
3. Enter percentages for different categories and ensure the total updates correctly.
4. Verify that the input summary displays correctly and updates as you modify values.
5. Check that form validation works properly for the percentage breakdown.

### Why make this change?

This change introduces a more user-friendly and intuitive way to input percentage breakdowns for activities. It provides a clear visual representation of different categories and their respective percentages, making it easier for users to accurately input and understand the data. The new component also includes built-in validation to ensure the total percentage adds up to 100%, improving data accuracy and user experience.


app/src/components/icons.tsx
```diff
@@ -338,3 +338,399 @@ export const DataConnectIcon = () => (
...
+          fill="#7A7B9A"
+        />
+      </g>
+    </g>
+  </svg>
+);
```
isaaccodekill, line 736: Just a comment for the future. We could looking into having a better process of dealing with our icons.
- Our Icons should have uniform sizing
- We should generate an icon sprite
- Geneate a single icon component `<Icon name=[icon_name] />`

app/src/components/percentage-breakdown-input.tsx
```diff
@@ -0,0 +1,226 @@
...
+  FoodIcon,
+  GardenIcon,
+  IndustrialIcon,
+  PaperIcon,
+  TextilesIcon,
+  WoodIcon,
+} from "./icons";
+import type { TFunction } from "i18next";
+
+const breakdownCategories = [
```
isaaccodekill, line 32: I think it would be a good idea not to hard code this and make it more dynamic, so this percentage breakdown component can be reused easily

isaaccodekill (changes requested):

app/src/components/icons.tsx
lemilonkh: Creating a ticket for this :+1:

app/src/components/percentage-breakdown-input.tsx
lemilonkh: Added a TODO comment. We will do this once there is a use case for multiple different kinds of percentage breakdown inputs.

isaaccodekill (approved):

PR #728, opened 2024-09-19
Format Number Inputs In Activity Forms

isaaccodekill:
image


app/src/components/formatted-number-input.tsx
```diff
@@ -0,0 +1,94 @@
...
+  miniAddon,
+  ...rest
+}: FormattedNumberInputProps) {
+  const format = (nval: number | string) => {
+    let val = parseInt(nval as string);
+    if (val === undefined || val === null || isNaN(val)) return "";
+    return val.toLocaleString();
+  };
+
+  const parse = (val: string) => val.replace(/,/g, "");
```
lemilonkh, line 39: Problem: this depends on the selected locale. Other languages have different decimal and thousands separators (e.g. German has , as decimal and . for thousands, so reversed from English).

Can we let an i18n/ locale library do this work for us instead of having to implement this ourselves?

Otherwise we will need to have the separators for each language in the i18n util.ts file, but this is clearly logic that others have implemented before and lut into libraries (e.g. i18next).

lemilonkh: We can make this a new ticket if necessary, but this needs to be dealt with rather urgently as other language users will be confused here.

app/src/components/formatted-number-input.tsx
```diff
@@ -0,0 +1,94 @@
...
+            <NumberInputField
+              data-testId={testId}
+              placeholder={placeholder}
+              h="48px"
+              type="text" // Use text type to allow formatted input
+              shadow="1dp"
+              borderRightRadius={children ? 0 : "md"} // Adjust border radius
+              bgColor={isDisabled ? "background.neutral" : "base.light"}
+              pos="relative"
+              zIndex={999}
```
lemilonkh, line 68: Could use a lower value here that still works to prevent escalating the numbers too much.

app/src/components/formatted-number-input.tsx
isaaccodekill: I'll localize

lemilonkh (approved):
Amazing! Thanks for the i18n fix :sparkles:

PR #737, opened 2024-09-23
remove selector since "by-activity" will not be implemented for now

thehighestprimenumber:
Since there was only one option ("by-activity") is on hold, I'm removing the selector.

before:
image

after:
image


thehighestprimenumber:
wrong branch name, replaced by https://github.com/Open-Earth-Foundation/CityCatalyst/pull/738

PR #739, opened 2024-09-24
Fix/prefill extra field

cephaschapa:
Changes
- Adds prefill functionality for activities inputs from suggested activity prefills
- This work for both single and multi-select options


app/src/components/MultiSelectInput.tsx
```diff
@@ -137,10 +143,15 @@ const MultiSelectWithCheckbox = ({
...
-                })) || []
+                field.value.length === 0
+                  ? preselectedValue.map((val: string) => ({
+                      label: t(val),
+                      value: val,
+                    }))
+                  : field.value?.map((val: string) => ({
+                      label: t(val),
+                      value: val,
+                    }))
```
lemilonkh, line 154: If you can, do this about the returned markup in the head part of the react component to make it more readable.

lemilonkh (approved):
Thanks :sparkles:

PR #742, opened 2024-09-25
fix(Readme): typo in the login email address changed to johndoe@example.com

engWafula:
The initally indicated email address  johndeo@example.com in the readme file didn't really work.


lemilonkh:
Thanks for the contribution. Since the information in the line is no longer accurate, we will instead remove it.

engWafula:
Alright  then

PR #753, opened 2024-09-30
[ON-2414] sector visualization backend (breakdown)

thehighestprimenumber:
currently we only  have the backend, can be tested by calling this url:

api/v0/inventory/:inventory-id/results/:sector-name-in-kebab-case/

returns data with this shape:

`{  
  data: {  
    activitiesForSectorBreakdown: {  
      byActivity: {  
        [buildingType: string]: {  
          [fuelType: string]: {  
            [unitType: string]: {  
              activityValue: string;  
              activityUnits: string;  
              totalActivityEmissions: string;  
              totalEmissionsPercentage: number;  
            };  
          };  
        };  
      };  
      byScope: {  
        [buildingType: string]: {  
          [scopeNumber: string]: {  
            [unitType: string]: {  
              activityValue: string;  
              activityUnits: string;  
              totalActivityEmissions: string;  
              totalEmissionsPercentage: number;  
            };  
          };  
        };  
      };  
    };  
  };  
}  `

### DIFF
--- a/app/src/app/api/v0/inventory/[inventory]/results/route.ts
+++ b/app/src/app/api/v0/inventory/[inventory]/results/route.ts
@@ -1,37 +1,39 @@
 import UserService from "@/backend/UserService";
-import { db } from "@/models";
-import { apiHandler } from "@/util/api";
-import { NextResponse } from "next/server";
-import { getEmissionResults } from "@/backend/ResultsService";
+import {db} from "@/models";
+import {apiHandler} from "@/util/api";
+import {NextResponse} from "next/server";
+import {getEmissionResults} from "@/backend/ResultsService";
 import sumBy from "lodash/sumBy";
 
 export const GET = apiHandler(
-  async (_req, { session, params: { inventory } }) => {
-    // ensure inventory belongs to user
-    await UserService.findUserInventory(inventory, session, [
-      {
-        model: db.models.InventoryValue,
-        as: "inventoryValues",
-        include: [
-          {
-            model: db.models.DataSource,
-            attributes: ["datasourceId", "sourceType"],
-            as: "dataSource",
-          },
-        ],
-      },
-    ]);
+    async (_req, {session, params: {inventory}}) => {
+        // ensure inventory belongs to user
+        await UserService.findUserInventory(inventory, session, [
+            {
+                model: db.models.InventoryValue,
+                as: "inventoryValues",
+                include: [
+                    {
+                        model: db.models.DataSource,
+                        attributes: ["datasourceId", "sourceType"],
+                        as: "dataSource",
+                    },
+                ],
+            },
+        ]);
 
-    const { totalEmissionsBySector, topEmissionsBySubSector } =
-      await getEmissionResults(inventory);
-    return NextResponse.json({
-      data: {
-        totalEmissions: {
-          bySector: totalEmissionsBySector,
-          total: sumBy(totalEmissionsBySector, (e) => Number(e.co2eq)),
-        },
-        topEmissions: { bySubSector: topEmissionsBySubSector },
-      },
-    });
-  },
+        const {totalEmissionsBySector, topEmissionsBySubSector} =
+            await getEmissionResults(inventory);
+
+
+        return NextResponse.json({
+            data: {
+                totalEmissions: {
+                    bySector: totalEmissionsBySector,
+                    total: sumBy(totalEmissionsBySector, (e) => Number(e.co2eq)),
+                },
+                topEmissions: {bySubSector: topEmissionsBySubSector},
+            },
+        });
+    },
 );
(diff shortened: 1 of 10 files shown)


app/src/app/api/v0/inventory/[inventory]/results/route.ts
```diff
@@ -1,37 +1,39 @@
 import UserService from "@/backend/UserService";
-import { db } from "@/models";
-import { apiHandler } from "@/util/api";
-import { NextResponse } from "next/server";
-import { getEmissionResults } from "@/backend/ResultsService";
+import {db} from "@/models";
```
Comment (lemilonkh) [inline app/src/app/api/v0/inventory/[inventory]/results/route.ts:2]: Run `prettier` on this file please and configure it to format on save if you can.

Comment (lemilonkh) [reply app/src/app/api/v0/inventory/[inventory]/results/route.ts:2]: I think this file just has formatting/ whitespace changes, so maybe just remove it from the PR if there is nothing else...

app/src/backend/ResultsService.ts
```diff
@@ -84,3 +93,177 @@ export async function getEmissionResults(inventoryId: string) {
...
+            JOIN "SubSector" ss ON iv.sub_sector_id = ss.subsector_id
+            JOIN "SubCategory" sc ON iv.sub_category_id = sc.subcategory_id
+            JOIN "Scope" scope ON scope.scope_id = sc.scope_id OR ss.scope_id = scope.scope_id
+        WHERE inventory_id = :inventoryId
+            AND s.sector_name ilike :sectorName
+    `;
+
+  try {
+    return (await db.sequelize!.query(rawQuery, {
+      replacements: { inventoryId, sectorName: sectorName.replace("-", " ") },
```
lemilonkh, line 141: Why are you doing this here?
These are probably translation strings that should be translated in the frontend, not made to look like english strings?

app/src/backend/ResultsService.ts
```diff
@@ -84,3 +93,177 @@ export async function getEmissionResults(inventoryId: string) {
...
+  const methodologyFields = isDirectMeasure
+    ? manualInputHierarchyElement.directMeasure
+    : manualInputHierarchyElement.methodologies?.find(
+        (m) => m.id === input_methodology,
+      );
+
+  if (!methodologyFields) {
+    console.error(
+      `Methodology fields not found for ${reference_number}, ${input_methodology}`,
+    );
```
lemilonkh, line 173: Throw `createHttpError.BadRequest` instead here if this is invalid data and should not be returned.

If you need to ignore the error and continue, this is fine ofc.

app/src/util/form-schema/index.ts
```diff
@@ -57,29 +57,33 @@ export interface SuggestedActivity {
...
 export interface Methodology {
-  id: string;
-  disabled?: boolean;
-  activities?: Activity[];
-  inputRequired?: string[];
-  formula?: string;
-  fields?: any[];
-  suggestedActivities?: SuggestedActivity[];
-  suggestedActivitiesId?: string;
+    id: string;
```
lemilonkh, line 60: Why all of these formatting changes? Please configure your editor for 2 spaces and enable format on save for prettier.

app/src/util/form-schema/index.ts
```diff
@@ -57,29 +57,33 @@ export interface SuggestedActivity {
...
-  id: string;
-  suggestedActivitiesId?: string;
-  suggestedActivities?: Activity[];
-  inputRequired?: string[];
-  "extra-fields"?: ExtraField[];
+    id: string;
+    suggestedActivitiesId?: string;
+    suggestedActivities?: Activity[];
+    inputRequired?: string[];
+    "extra-fields"?: ExtraField[];
```
lemilonkh, line 77: Can we use a JS-compatible camelCased name here for consistency? Or is there no other way because you're doing a lookup somewhere?

app/src/backend/ResultsService.ts
thehighestprimenumber: if we use the translation, we might receive it in spanish, portuguese or german. so I'm planning on sending the translation string directly (eg: stationary-energy), and then I have to remove the "-" so query in the db

app/src/backend/ResultsService.ts
thehighestprimenumber: I want to log the error and continue

app/src/util/form-schema/index.ts
thehighestprimenumber: this is the interface for what's defined in the json so I can't change it

app/tests/api/results.jest.ts
```diff
@@ -1,101 +1,118 @@
...
+import {expectStatusCode, mockRequest, setupTests, testUserID} from "../helpers";
 
-import { Inventory } from "@/models/Inventory";
+import {Inventory} from "@/models/Inventory";
 import {
-  activityValues as activityValuesData,
-  baseInventory,
-  inventoryId,
-  inventoryValues as inventoryValuesData
+    activityValues as activityValuesData,
```
lemilonkh, line 9: This file has inconsistent indentation and formatting

lemilonkh (approved):
Thanks ✨

PR #763, opened 2024-10-02
[ON-2523] Handle automatic update when a user enters Inventory Results

thehighestprimenumber:
When an activity is created/updated/deleted, invalidate the InventoryProgress and ReportResults tags


lemilonkh (approved):
Thanks 👍✨

PR #767, opened 2024-10-02
On 2569 change initialize thread

mircorudolph:
I moved the initialization of the thread (communication with openAI assistant API) from rendering the chat component to sending a message. 
Now the initial answer will take slightly longer, but we do not create a lot of empty threads each time it gets rendered.


lemilonkh (approved):
Thanks :sparkles:

PR #772, opened 2024-10-04
On 2431 record conversations

mircorudolph:
When sending a message, the created threadId gets saved in our database in table 'AssistantThread'


app/migrations/20241003214418-create_assistant_messages_table.cjs
```diff
@@ -0,0 +1,34 @@
+"use strict";
+
+const sql_up = `create type role_enum as enum('user', 'assistant');
```
lemilonkh, line 3: You could use a transaction here to make sure either all of these SQL commands are executed or none are. Usually advisable once you have to do multiple queries in one migration.

app/migrations/20241003214418-create_assistant_messages_table.cjs
```diff
@@ -0,0 +1,34 @@
...
+
+create table if not exists public."AssistantMessage"
+(
+    assistant_message_id  text not null primary key,
+    thread_id             text not null,
```
lemilonkh, line 8: Can you use UUIDs (v4) for the table IDs? All our other tables do this, so it's preferable to keep this pattern.
Comparing and indexing UUID columns is more efficient compared to using arbitrary strings.

app/migrations/20241003213605-create_assistant_thread_table.cjs
```diff
@@ -0,0 +1,23 @@
+"use strict";
+
+const sql_up = `create table if not exists public."AssistantThread"
+(
+    assistant_thread_id text not null primary key,
```
lemilonkh, line 5: Use type `uuid` for ID columns of tables and for foreign keys.
Create it in code with randomUUID() imported using `import { randomUUID } from "node:crypto";`
You can also use auto-generated IDs in SQL queries if that's easier (`uuid_generate_v4()` from the `uuid-ossp` PostgreSQL extension that we probably need to enable still).

Extension can be enabled using this SQL query in a migration:
`CREATE EXTENSION IF NOT EXISTS "uuid-ossp";`
Might need permissions though.

app/src/app/api/v0/threads/route.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    if (!input.threadId || typeof input.threadId !== "string") {
+      return NextResponse.json(
+        { error: "Invalid threadId provided" },
+        { status: 400 },
+      );
+    }
+
+    const response = await db.models.AssistantThread.create({
+      assistantThreadId: input.threadId,
```
lemilonkh, line 20: I think you need to pass the assistant ID as well? Otherwise it will just be null?

app/migrations/20241003213605-create_assistant_thread_table.cjs
lemilonkh: In case this is a constraint of the OpenAI API, no worries.

lemilonkh (approved):
Thanks :sparkles:

app/migrations/20241003213605-create_assistant_thread_table.cjs
thehighestprimenumber: the ids come from openAPI

app/migrations/20241003214418-create_assistant_messages_table.cjs
thehighestprimenumber: the ids come from openAPI

app/src/app/api/v0/threads/route.ts
```diff
@@ -0,0 +1,28 @@
+import { db } from "@/models";
+import { apiHandler } from "@/util/api";
+import createHttpError from "http-errors";
+import { NextRequest, NextResponse } from "next/server";
+
+export const POST = apiHandler(async (req) => {
```
thehighestprimenumber, line 6: can we move this to assistants/threads?
Threads is a word with a lot of meanings and I'd like it to be more specific

app/src/components/ChatBot/chat-bot.tsx
```diff
@@ -97,24 +98,39 @@ export default function ChatBot({
...
         inventoryId: inventoryId,
         content: t("initial-message"),
       }).unwrap();
 
-      // Set the threadIdRef which gets set synchronously
+      // Set the threadIdRef synchronously
       threadIdRef.current = result;
+
+      // Attempt to save threadId in the database asynchronously
+      fetch(`/api/v0/threads/`, {
```
thehighestprimenumber, line 113: can we move this logic to app/src/services/api.ts ?

thehighestprimenumber (changes requested):

app/migrations/20241003214418-create_assistant_messages_table.cjs
mircorudolph: Yes the IDs are in the form of 'thread_deMv.......' therefore we are not using the UUID here

app/src/app/api/v0/threads/route.ts
mircorudolph: moved it to \api\v0\assistants\threads\export

app/src/app/api/v0/threads/route.ts
mircorudolph: The workflow is as follows:
1. When the thread gets created, there is only the threadId. 
2. With the threadID in the Python script I extract all the messages and then later write the messages togehter with the assistantID to the table.
But actually we already have the assistantID in the environment variables so I could set it directly as well thats true.

Will update it

app/migrations/20241003214418-create_assistant_messages_table.cjs
mircorudolph: Done

thehighestprimenumber (approved):

PR #781, opened 2024-10-08
Fix/climate assistant proportions

cephaschapa:
Changes
- Uses vh and vw for height and width respectively to accomodate appropriate proportions to the Climate Assistant


lemilonkh (approved):
Thanks :sparkles:

PR #783, opened 2024-10-08
Fix/persist language in cookie

cephaschapa:
Changes
- Persists selected language in coockie i18next
- Adds cache control in nextjs config
- Adds event listener to update url based on language set in cookie
- Updates nav bar icon URL
- Modifies middleware to handle redirects correctly
image

### DIFF
--- a/app/next.config.mjs
+++ b/app/next.config.mjs
@@ -14,10 +14,10 @@ const nextConfig = {
   env: {
     APP_VERSION: packageInfo.version,
   },
-  async headers() {
+   async headers() {
     return [
       {
-        source: "/:path*", // Apply to all paths
+        source: "/", 
         headers: [
           {
             key: "Cache-Control",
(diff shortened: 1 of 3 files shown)


app/next.config.mjs
```diff
@@ -14,6 +14,19 @@ const nextConfig = {
...
     APP_VERSION: packageInfo.version,
   },
+   async headers() {
+    return [
+      {
+        source: "/", 
+        headers: [
+          {
+            key: "Cache-Control",
+            value: "no-store, no-cache, must-revalidate, proxy-revalidate",
```
lemilonkh, line 24: What does this fix? This means we're no longer caching any pages or scripts?

lemilonkh (approved):
Thanks, let's see how this performs in our edge cases on dev 👍✨

app/next.config.mjs
cephaschapa: This is an extra step meant to prevent pages from being cached I believe otherwise we could remove it if it affects our current caching mechanism . :crossed_fingers: Thanks :+1:

PR #785, opened 2024-10-08
fix: migrate user data from datasource table to activity value metadata

isaaccodekill:


lemilonkh (changes requested):
This only contains the client-side changes. We need to adjust our API routes and move the existing data from the old `DataSource` entries with `source_type === "user"`to the respective metadata properties in the associated `ActivityValue`.

lemilonkh (approved):

PR #787, opened 2024-10-08
Fix on 2602 errors loading seed data

amanda-eames:
- Added files for loading
- fixed issues with CarbonFootprint
- fixed issues with gpcprotocol


app/seed-data/emissions_factors/scripts/ghgprotocol_transportation.py
```diff
@@ -418,12 +418,12 @@ def save_to_csv(fl, data):
...
         "geographic",
-        "induced_activity_1",
-        "induced_activity_2",
-        "resident_activity",
-        "movement_driver",
+        "induced-activity-1",
+        "induced-activity-2",
+        "resident-activity",
+        "movement-driver",
         "TBD",
```
lemilonkh, line 427: Do we still need the TBD here?

lemilonkh (approved):
Thanks ✨

app/seed-data/emissions_factors/scripts/ghgprotocol_transportation.py
amanda-eames: Probably not, I think that's just a place holder let me filter it out so we don't load unnecessary data

PR #789, opened 2024-10-08
(bug) [ON-2522] show connected data on report results

thehighestprimenumber:
Querying from inventoryValue instead of ActivityValue


lemilonkh (approved):
Thanks :sparkles:

PR #795, opened 2024-10-10
pr review: [ON-2379] frontend changes

thehighestprimenumber:
pr comments for PR https://github.com/Open-Earth-Foundation/CityCatalyst/pull/786 that trigger failing playwright tests


app/src/util/helpers.ts
```diff
@@ -254,33 +254,33 @@ export const getInputMethodology = (methodologyId: string) => {
 };
 
 export function convertKgToTonnes(
-  valueInTonnes: number | bigint,
+  valueInKg: number | bigint,
```
lemilonkh, line 257: Let's make sure we're not introducing a breaking change here, since this function is probably used in other places as well?

lemilonkh (approved):
Thanks :sparkles:

app/src/util/helpers.ts
thehighestprimenumber: thanks! I rolled back the change

PR #810, opened 2024-10-14
incineration waste formula

isaaccodekill:
image


app/src/backend/formulas.ts
```diff
@@ -6,6 +6,9 @@ import type { GasValueCreationAttributes } from "@/models/GasValue";
 import type { EmissionsFactorAttributes } from "@/models/EmissionsFactor";
 import { findClosestCityPopulation } from "@/util/population";
 import type { Inventory } from "@/models/Inventory";
+import { db } from "@/models";
+import { InventoryValue } from "@/models/InventoryValue";
+import { Decimal } from "decimal.js";
```
lemilonkh, line 11: This hasn't been added to package.json yet it seems?

app/src/backend/formulas.ts
```diff
@@ -61,6 +72,189 @@ export function handleDirectMeasureFormula(
...
+      "technology-open-burning": 150,
+    },
+    "waste-composition-industrial-solid-waste": {
+      "technology-continuous-incineration": 100,
+      "technology-semi-continuous-incineration": 100,
+      "technology-batch-type-incineration": 100,
+      "technology-open-burning": 100,
+    },
+    "waste-composition-clinical-waste": {
+      "technology-continuous-incineration": 100,
```
lemilonkh, line 117: This data could be moved to the top of the file or a different file and imported

app/src/backend/formulas.ts
```diff
@@ -61,6 +72,189 @@ export function handleDirectMeasureFormula(
...
+    },
+  };
+
+  if (!data) {
+    throw new createHttpError.BadRequest(
+      "Activity has no data associated, so it can't use 'incineration-waste' formula",
+    );
+  }
+
+  const activityTitle = activityValue.metadata?.["activityTitle"];
```
lemilonkh, line 142: Is this the correct key for it? Shouldn't it be saved to `activity-title`?

app/src/backend/formulas.ts
```diff
@@ -61,6 +72,189 @@ export function handleDirectMeasureFormula(
...
+      gas: "CH4",
+      amount: BigInt(new Decimal(totalCH4Emission).trunc().toString()),
+    },
+    {
+      gas: "N2O",
+      amount: BigInt(new Decimal(totalN2OEmissions).trunc().toString()),
+    },
+    {
+      gas: "CO2",
+      amount: BigInt(new Decimal(totalCO2Emissions).trunc().toString()),
```
lemilonkh, line 253: If we are using DecimalJS now, we should return that instead of a truncated BigInt to keep the precision. Fine for now, we will do this once the other formulas are refactored to use Decimal.js as well :+1:

app/src/components/percentage-breakdown-input.tsx
```diff
@@ -113,13 +124,16 @@ const PercentageBreakdownInput: FC<FormInputProps> = ({
   const breakdownSummary = useMemo(() => {
     return Object.entries(breakDownValues)
       .map(([key, value]) => {
-        const category = breakdownCategories.find((c) => c.id === key);
-        return `${t(category?.id ?? "")} ${value}%`;
+        console.log(key, value);
```
lemilonkh, line 127: Remove logging

app/src/components/percentage-breakdown-input.tsx
```diff
@@ -113,13 +124,16 @@ const PercentageBreakdownInput: FC<FormInputProps> = ({
...
+        const category = breakdownCategories.find((c) => c === key);
+        return `${t(category ?? "")} ${value}%`;
       })
       .join(", ");
     // breakdownCategories
   }, [breakDownValues, t]);
 
+  console.log(breakdownSummary, "the summary");
```
lemilonkh, line 135: Remove logging

app/src/i18n/locales/en/data.json
```diff
@@ -579,11 +578,11 @@
...
   "gallons-per-kilometer": "Gallons per kilometer (gal/km)",
   "gallons-per-miles": "Gallons per miles (gal/mi)",
-  "incineration-waste-inboundary-boiler-type": "Boiler type",
+  "boiler-type": "Boiler type",
   "incineration-waste-inboundary-incineration-facility-address": "Incineteration facility address (Optional)",
   "incineration-waste-inboundary-incineration-facility-id": "Incineration facility ID",
-  "incineration-waste-inboundary-technology": "Technology",
-  "incineration-waste-inboundary-waste-composition": "Waste Composition",
+  "technology": "Technology",
+  "waste-composition": "Waste Composition",
```
lemilonkh, line 585: I think we need to keep these like they are unfortunately?
Or why are the strings passed to `t` shortened in this way?

app/src/i18n/locales/es/data.json
```diff
@@ -578,11 +577,11 @@
...
   "gallons-per-kilometer": "Galones por kilómetro (gal/km)",
   "gallons-per-miles": "Galones por milla (gal/mi)",
-  "incineration-waste-inboundary-boiler-type": "Tipo de caldera",
+  "boiler-type": "Tipo de caldera",
   "incineration-waste-inboundary-incineration-facility-address": "Dirección de la instalación de incineración (opcional)",
   "incineration-waste-inboundary-incineration-facility-id": "ID de la instalación de incineración",
-  "incineration-waste-inboundary-technology": "Tecnología",
-  "incineration-waste-inboundary-waste-composition": "Composición de los residuos",
+  "technology": "Tecnología",
+  "waste-composition": "Composición de los residuos",
```
lemilonkh, line 584: Let's double check

app/src/i18n/locales/pt/data.json
```diff
@@ -579,11 +579,10 @@
   "fugitive-emissions-oil-gas-type-oil": "Petróleo",
   "gallons-per-kilometer": "Galões por quilômetro (gal/km)",
   "gallons-per-miles": "Galões por milha (gal/mi)",
-  "incineration-waste-inboundary-boiler-type": "Tipo de Caldeira",
+  "boiler-type": "Tipo de Caldeira",
   "incineration-waste-inboundary-incineration-facility-address": "Endereço da Instalação de Incineração (Opcional)",
   "incineration-waste-inboundary-incineration-facility-id": "ID da Instalação de Incineração",
-  "incineration-waste-inboundary-technology": "Tecnologia",
-  "incineration-waste-inboundary-waste-composition": "Composição dos Resíduos",
+  "technology": "Tecnologia",
```
lemilonkh, line 585: Check

app/src/models/FormulaInput.ts
```diff
@@ -160,6 +152,8 @@ export class FormulaInput
           type: DataTypes.UUID,
           allowNull: false,
           field: "formulainput_id",
+          primaryKey: true,
+          defaultValue: DataTypes.UUIDV4,
```
lemilonkh, line 156: We should do this for more tables, much easier to create new instances that way :+1:

app/src/util/form-schema/manual-input-hierarchy.json
lemilonkh, line 1: It seems like you didn't change the content of the file, just the formatting?
If so, let's maybe remove this from the PR to prevent potential merge conflicts today and tomorrow.

app/src/util/form-schema/manual-input-hierarchy.json
```diff
@@ -3593,31 +3806,37 @@
...
                 "options": [
                   "technology-continuous-incineration",
                   "technology-semi-continuous-incineration",
                   "technology-batch-type-incineration"
                 ]
               },
               {
-                "id": "incineration-waste-inboundary-boiler-type",
-                "options": ["boiler-type-stoker", "boiler-type-fluidised-bed"]
+                "id": "boiler-type",
```
lemilonkh, line 3829: Sorry, I had to revert the PR with the activity input ID normalization as Evan said it would introduce too many breaking changes too late in the process.
So let's revert this to the old state (and the translation files as well) and use a name mapping in the formula to get the value depending on which activity is used.

Annoying, but nothing to be done about that for now...

lemilonkh (commented):
Thanks! :sparkles:

app/src/util/form-schema/manual-input-hierarchy.json
lemilonkh: There are some changes but we will likely need to revert them as well unfortunately.

app/src/util/form-schema/manual-input-hierarchy.json
isaaccodekill: @lemilonkh  so for the changes I made. I made them to only the methodologies I worked with.  I also cross check the translation as that the only other place it was being used

app/src/i18n/locales/en/data.json
isaaccodekill: in this case the translation strings already exists.

app/src/util/form-schema/manual-input-hierarchy.json
lemilonkh: Yes, however changing only these and not the other ones we introduce breaking changes to our data model all the time. So let's do them all at once in a separate PR if possible.

app/src/backend/formulas.ts
isaaccodekill: no that was for the activity data. We can change this to snake case in the future

app/src/i18n/locales/en/data.json
```diff
@@ -579,11 +578,14 @@
   "fugitive-emissions-oil-gas-type-oil": "Oil",
   "gallons-per-kilometer": "Gallons per kilometer (gal/km)",
   "gallons-per-miles": "Gallons per miles (gal/mi)",
-  "incineration-waste-inboundary-boiler-type": "Boiler type",
+  "boiler-type": "Boiler type",
```
lemilonkh, line 581: We need to change these keys back since the ID change for the inputs was reverted, right?

lemilonkh (approved):
Thanks :sparkle:

app/src/i18n/locales/en/data.json
lemilonkh: For all 3 affected translation files

PR #811, opened 2024-10-14
[ON-2266] fix: hover population year

thehighestprimenumber:
image
 show the year on the population hover


lemilonkh (approved):
Thanks :sparkles:

PR #812, opened 2024-10-15
Implements multiple activities per methodology

isaaccodekill:
https://github.com/user-attachments/assets/b688e5c5-b88e-470d-b5b3-76e2b99e4410

includes ui changes and api changes


app/src/backend/formulas.ts
```diff
@@ -216,9 +216,10 @@ export function handleIndustrialWasteWaterFormula(
   const degradableOrganicComponents = data["degradable-organic-components"];
   const methaneProductionCapacity =
     data["methane-production-capacity"] ?? DEFAULT_METHANE_PRODUCTION_CAPACITY; // TODO should this only be handled UI-side?
-  const removedSludge = data["removed-sludge"];
-  const methaneCorrectionFactor = data["methane-correction-factor"];
-  const methaneRecovered = data["methane-recovered"];
+  const removedSludge = data["total-organic-sludge-removed"];
+  const methaneCorrectionFactor = data["methane-correction-factor"]; // TODO @leminkoh is this correct? it doesn't exist on the ui
```
isaaccodekill, line 220: I don't if this is correct @lemilonkh  because this field doesn't exist on the UI

lemilonkh: This needs to come from the FormulaInput table (it's called MCF there
). Alternatively for now you can encode similar if/ else rules like we have for other formulas to not have to load this data until it's ready.

app/src/backend/formulas.ts
```diff
@@ -244,8 +246,7 @@ export async function handleDomesticWasteWaterFormula(
   }
 
   const methaneProductionCapacity = DEFAULT_METHANE_PRODUCTION_CAPACITY; // TODO should this only be handled UI-side?
-  const removedSludge =
-    data["wastewater-inside-domestic-calculator-total-organic-sludge-removed"];
+  const removedSludge = data["total-organic-sludge-removed"];
```
lemilonkh, line 249: I think we might need to revert this change?
And need a better mapping of input IDs to actual formula values that allows the different names from different activities to be read here.

app/src/util/form-schema/manual-input-hierarchy.json
```diff
@@ -717,20 +776,26 @@
             "prefills": [
               {
                 "key": "manufacturing-and-construction-type",
-                "value": "manufacturing-and-construction-type-institutional-buildings"
+                "value": "institutional-buildings"
```
lemilonkh, line 779: We will unfortunately have to revert the changes in this file.

lemilonkh (commented):
Thanks :sparkles:

app/src/backend/formulas.ts
isaaccodekill: this wasn't a change from the manual input. The formula was just referring to the wrong property

app/src/util/form-schema/manual-input-hierarchy.json
lemilonkh: Seems to be thorough so let's go with it

app/src/backend/formulas.ts
isaaccodekill: I'll extend this in a follow up ticket

lemilonkh: As long as it doesn't give us an error with saving for now that's good. Otherwise comment this line and replace it with a constant for now.

lemilonkh: Because the result will be NaN otherwise

lemilonkh (approved):

PR #814, opened 2024-10-15
fix(ui): removes fixed height from climate assistant that caused overflow

cephaschapa:
changes
- Removes fixed height property that caused chat components to overflow


lemilonkh (approved):
Thanks :sparkles:

PR #818, opened 2024-10-16
Feat/add unit tests for formulas

cephaschapa:
Changes
- Adds test script for windows evn
- Adds unit tests for the POST and PATCH route to test formulas based on methodology


app/tests/api/activity_value.jest.ts
```diff
@@ -142,51 +146,135 @@ describe.skip("Activity Value API", () => {
   it("should not create an activity value with invalid data", async () => {
     const req = mockRequest(invalidCreateActivity);
     const res = await createActivityValue(req, {
-      params: { inventory: inventory.inventoryId }
+      params: { inventory: inventory.inventoryId },
     });
     expect(res.status).toBe(400);
   });
 
+  it("should not create an activity value with with formulas", async () => {
```
lemilonkh, line 154: Why? Really this should return a 200 if the data is correct?

lemilonkh (approved):
Thanks ✨

PR #819, opened 2024-10-16
filter emission factors for scope 2, group activities for methodologies with >1 activities

isaaccodekill:


lemilonkh (approved):
Thanks ✨

PR #824, opened 2024-10-21
fix(ui): replace portugal flag icon with brazil flag

cephaschapa:
Changes
- Adds logic to resolve and place the portuguese flag icon with brazilian flag for the dropdown
- Adds the same logic for the selected portuguese flag

image


lemilonkh (approved):
Thanks ✨

PR #835, opened 2024-10-22
[ON-2612] migrate from bigint to decimal.js

thehighestprimenumber:


app/src/backend/ActivityService.ts
```diff
@@ -217,7 +219,7 @@ export default class ActivityService {
           BigInt(inventoryValue.co2eq as bigint) -
             BigInt(activityValue.co2eq as bigint) ?? 0n;
 
-        const calculatedCO2e = BigInt(totalCO2e); // Ensure totalCO2e is BigInt
+        const calculatedCO2e = BigInt(totalCO2e.toNumber()); // Ensure totalCO2e is BigInt
```
lemilonkh, line 222: Can you instead convert totalCO2e to a string and pass it here? Otherwise you're losing all of the precision gained by using decimal.js here by using the lossy JS Number type as an intermediary.

app/src/backend/ActivityService.ts
```diff
@@ -321,7 +323,7 @@ export default class ActivityService {
           );
 
         const currentCO2e = BigInt(inventoryValue.co2eq ?? 0n);
-        const calculatedCO2e = BigInt(totalCO2e); // Ensure totalCO2e is BigInt
+        const calculatedCO2e = BigInt(totalCO2e.toNumber()); // Ensure totalCO2e is BigInt
```
lemilonkh, line 326: Use `.toString()` instead (see above)

app/src/backend/ActivityService.ts
```diff
@@ -330,7 +332,7 @@ export default class ActivityService {
         );
 
         await inventoryValue.save({ transaction });
-        activityValue.co2eq = totalCO2e;
+        activityValue.co2eq = BigInt(totalCO2e.toNumber());
```
lemilonkh, line 335: .toString()

app/src/backend/formulas.ts
```diff
@@ -182,10 +182,14 @@ export async function handleIncinerationWasteFormula(
       );
     }
 
-    (totalCH4Emission +=
-      AmountOfWasteForWasteTypeI * CH4EmissionFactorForWasteTypeI * 10 ** -3),
-      (totalN2OEmissions +=
-        AmountOfWasteForWasteTypeI * NO2EmissionFactorForWasteTypeI * 10 ** -3);
+    totalCH4Emission = Decimal.sum(
+      totalCH4Emission,
+      AmountOfWasteForWasteTypeI * CH4EmissionFactorForWasteTypeI * 10 ** -3,
```
lemilonkh, line 187: Can we define these as Decimal as well?

app/src/backend/formulas.ts
```diff
@@ -247,29 +251,32 @@ export async function handleIncinerationWasteFormula(
...
 
-    totalPartialCO2Emissions +=
+    totalPartialCO2Emissions = Decimal.sum(
+      totalPartialCO2Emissions,
       WasteFractionI *
-      dryMatterContentI *
-      fractionOfCarbonI *
-      fractionOfFossilCarbonI *
-      oxidationFactorI;
+        dryMatterContentI *
```
lemilonkh, line 257: Do this with decimal.js instead of JS Numbers to prevent overflow

app/src/backend/formulas.ts
```diff
@@ -372,13 +379,14 @@ export function handleMethaneCommitmentFormula(
...
 
-  const ch4Emissions =
-    totalSolidWaste *
+  const ch4Emissions = Decimal.mul(
+    totalSolidWaste,
     methaneGenerationPotential *
-    (1 - recoveredMethaneFraction) *
-    (1 - oxidationFactor);
+      (1 - recoveredMethaneFraction) *
+      (1 - oxidationFactor),
```
lemilonkh, line 386:
```suggestion
      1 - recoveredMethaneFraction,
      1 - oxidationFactor,
```
To prevent number overflows

lemilonkh (approved):
Amazing, thanks 🙌✨

PR #836, opened 2024-10-22
Feat/on 2594 filter emissions factors

isaaccodekill:


app/src/app/api/v0/emissions-factor/route.ts
```diff
@@ -2,33 +2,58 @@ import { db } from "@/models";
...
-    attributes: ["regionLocode"],
-    include: [
-      {
-        model: db.models.Inventory,
-        as: "inventories",
-        attributes: [],
-        where: {
-          inventoryId: inventoryId,
+const filterMappings: Record<string, any> = {
+  "fuel-type-wood/wood-waste": "fuel-type-wood-wood-waste",
```
lemilonkh, line 9: This should probably be fixed in the data (seeder or MI JSON depending on where this comes from), rather than "painted over" in JS...

app/src/app/api/v0/emissions-factor/route.ts
```diff
@@ -39,6 +64,39 @@ export const GET = apiHandler(async (req: NextRequest, _context: {}) => {
       whereClause.methodologyName = "modeled-data";
   }
 
+  if (!!metadata) {
+    let andCondition = [];
+    for (let key in metadata) {
+      let clause = {
+        [Op.or]: [
+          {
+            [`metadata.${key}`]: "nan",
```
lemilonkh, line 73: I think you can use the literal `NaN` instead of a string here? Or does that not work with the types?

app/src/services/api.ts
```diff
@@ -421,8 +421,13 @@ export const api = createApi({
       }
     >({
       query: (params) => {
-        const queryString = new URLSearchParams(params).toString();
-        return `/emissions-factor${queryString ? `?${queryString}` : ""}`;
+        // const queryString = new URLSearchParams(params).toString();
```
lemilonkh, line 424: Remove this line

app/src/services/api.ts
```diff
@@ -421,8 +421,13 @@ export const api = createApi({
...
+        return {
+          url: `/emissions-factor`,
+          method: "POST",
+          body: params,
+        };
+        // return `/emissions-factor${queryString ? `?${queryString}` : ""}`;
```
lemilonkh, line 430: Can this also be removed?

lemilonkh (approved):
Thanks ✨

app/src/app/api/v0/emissions-factor/route.ts
isaaccodekill: For this it needs to be the actual 'nan' string. that's the default string the mau created for factors that are applicable for all options

app/src/app/api/v0/emissions-factor/route.ts
isaaccodekill: This is a good point. I think we can delay this till when we are updating the manual input json file. then we can replace the keys across board

PR #837, opened 2024-10-22
bug: [ON-2663] go back to hardcoded scopes

thehighestprimenumber:
because the db returns more scopes than those required by BASIC. We'll have to re-work the db when we start supporting more standards


lemilonkh (approved):
Thanks :sparkles:

PR #841, opened 2024-10-23
Fix/bring climate advisor to front

cephaschapa:
Changes
-  Adds modifiers to customize the popover zindex properties


lemilonkh (approved):
Thanks 👍🏼✨

PR #858, opened 2024-10-28
Missing translations

evanp:
These are a bunch of missing translation strings from the wastewater sector which were never added when we added the manual input definitions.

There are also a few translation strings that were added to the manual input document that are no longer in the design document.

Finally, there was a bug in the "number" type extra field where the label wasn't getting translated. That is also fixed.


evanp:
I also added a script to find untranslated strings in the manual input file.

cephaschapa (approved):
Great catch 👍, we could use copilot to catch typos for translation strings.

PR #861, opened 2024-10-29
Filter Emission Factor Value

isaaccodekill:
- Convert all units to default-units in calculations 
- Display Emission factor units based on selected extra field values


app/src/app/api/v0/emissions-factor/route.ts
```diff
@@ -101,6 +92,17 @@ export const POST = apiHandler(async (req: NextRequest, _context: {}) => {
...
+  if (methodologyId) {
+    if (methodologyId.includes("fuel-combustion"))
+      whereClause.methodologyName = "fuel-combustion-consumption";
+    if (methodologyId.includes("scaled"))
+      whereClause.methodologyName = "sampling-scaled-data";
+    if (methodologyId.includes("modeled-data"))
+      whereClause.methodologyName = "modeled-data";
+  }
+
+  console.log(whereClause, "where");
```
lemilonkh: Remove logging

app/src/backend/UnitConversionService.ts
```diff
@@ -0,0 +1,73 @@
+// Purpose: Service to convert units between different systems.
+
+import createHttpError from "http-errors";
+
+export default class UnitConversionService {
+  private static conversionTable: { [key: string]: { [key: string]: number } } =
+    {
+      "units-cubic-meters": {
+        "units-gallons": 0.00378541, // 1 gallon = 0.00378541 cubic meters
```
lemilonkh, line 9: Maybe we should use decimal.js for conversions here at some point, so we don't run into floating point inaccuracy issues... This many decimals are probably handle-able by JS Numbers, but it would be nice to check if the resulting precision is sufficient in a later QA step.

app/src/backend/formulas.ts
```diff
@@ -115,6 +117,80 @@ const DEFAULT_INCOME_GROUP_FRACTIONS: Record<string, number> = {
...
+  if (activity?.["default-units"]) {
+    const val = data[activity?.["activity-title"] as string];
+    const fuelTypeKey = Object.keys(data).find((key) =>
+      key.includes("fuel-type"),
+    );
+    const fuelType = data[fuelTypeKey as string];
+    data[activity?.["activity-title"] as string] = new Decimal(
+      UnitConversionService.convertUnits(
+        val,
+        data[`${activity?.["activity-title"]}-unit`],
```
lemilonkh, line 168: Could you maybe extra this key into a variable beforehand instead of doing string interpolation in the object access in the function call?

app/src/backend/formulas.ts
```diff
@@ -321,8 +408,14 @@ export function handleVkt1Formula(
 
 export function handleMethaneCommitmentFormula(
   activityValue: ActivityValue,
+  inventoryValue: InventoryValue,
 ): Gas[] {
-  const data = activityValue.activityData;
+  const data = convertDataToDefaultUnit(
+    // use convert all the values
+    activityValue,
+    inventoryValue.inputMethodology as string,
```
lemilonkh, line 416: Instead of converting the type it might be good to handle null cases in some place beforehand?

app/src/backend/formulas.ts
```diff
@@ -466,11 +571,18 @@ export function handleIndustrialWasteWaterFormula(
   return [{ gas: "CH4", amount }];
 }
 
+// TODO how correct is this formula ?
```
lemilonkh, line 574: Yes I think Mau's calculations confirmed we have some work to do on it. We will have tickets for the upcoming sprint for it.

app/src/components/Modals/activity-modal/activity-form-modal.tsx
```diff
@@ -85,6 +84,15 @@ const AddActivityModal: FC<AddActivityModalProps> = ({
     methodology: methodology,
   });
 
+  const { emissionsFactorTypes, emissionsFactorsLoading } = useEmissionFactors({
```
lemilonkh, line 87: Calling it `areEmissionsFactorsLoading` might make it more clear that it's a boolean.

app/src/components/Modals/activity-modal/activity-form-modal.tsx
```diff
@@ -348,6 +306,7 @@ const AddActivityModal: FC<AddActivityModalProps> = ({
           <ModalCloseButton marginTop="10px" />
           <ActivityModalBody
             emissionsFactorTypes={emissionsFactorTypes}
+            loadingEmissionFactors={emissionsFactorsLoading}
```
lemilonkh, line 309: Two different names for the same field are a bit confusing

app/src/components/Modals/activity-modal/activity-modal-body.tsx
```diff
@@ -113,12 +119,18 @@ const ActivityModalBody = ({
...
 
   const unitValue = useWatch({
     control,
-    name: `activity.${title}Unit` as any,
+    name: `activity.${title}-unit` as any,
+  });
+
+  const emissionFactorTypeValue = useWatch({
+    control,
+    name: `activity.emissionFactorType` as any,
```
lemilonkh, line 133: As any should not be required here? Same as back tick string.

app/src/components/formatted-number-input.tsx
```diff
@@ -57,6 +59,12 @@ function FormattedNumberInput({
   return (
     <Controller
       control={control}
+      rules={{
+        required: t("value-required"),
+        validate: (value) => {
+          return (value !== "" && !isNaN(value)) || t("value-required");
```
lemilonkh, line 65: Remove debug logging

app/src/hooks/activity-value-form/use-emission-factors.ts
```diff
@@ -0,0 +1,155 @@
...
+    ...source,
+    gasValuesByGas: Object.keys(source.gasValuesByGas).reduce(
+      (acc, currentValue) => {
+        return {
+          ...acc,
+          [currentValue]: {
+            gasValues: uniqBy(
+              source.gasValuesByGas[currentValue].gasValues,
+              "emissionsPerActivity",
+            ).filter((factor) => ["kg/m3", "kg/kWh"].includes(factor.units)), // filter only emissions that have kg/m3 or kg/kWh as the unit
```
lemilonkh, line 67: Why only these?

app/src/hooks/activity-value-form/use-emission-factors.ts
```diff
@@ -0,0 +1,155 @@
+// take the methodology selected,
```
lemilonkh, line 1: What is this comment for?

app/src/hooks/activity-value-form/use-emission-factors.ts
```diff
@@ -0,0 +1,155 @@
...
+      return [];
+    }
+
+    // now that we have three or more emission factors, we want to reduce it down to a collection of gases per dataset
+    return reduceEmissionsToUniqueSourcesAndUnits(emissionsFactors);
+  }, [emissionsFactors]);
+
+  // when the data changes we wanna set the emissions factor types to the first selection that exists.
+  useEffect(() => {
+    console.log(emissionsFactorTypes, "they changed");
```
lemilonkh, line 143: Remove debug logging

app/src/hooks/activity-value-form/use-emission-factors.ts
```diff
@@ -0,0 +1,155 @@
...
+    // if (emissionsFactorTypes.length > 0) {
+    //   setValue("activity.emissionFactorType", emissionsFactorTypes[0].id);
+    // }
```
lemilonkh, line 146: Does this need to be re-enabled?

app/src/services/api.ts
```diff
@@ -23,636 +23,637 @@ import type {
   UserInfoResponse,
   UserInviteResponse,
 } from "@/util/types";
-import type { GeoJSON } from "geojson";
-import { createApi, fetchBaseQuery } from "@reduxjs/toolkit/query/react";
+import type {GeoJSON} from "geojson";
```
lemilonkh, line 26: Can you run Prettier on this file to remove the unnecessary formatting changes?

lemilonkh (changes requested):

app/src/backend/UnitConversionService.ts
isaaccodekill: Noted

app/src/hooks/activity-value-form/use-emission-factors.ts
isaaccodekill: Not really

app/src/hooks/activity-value-form/use-emission-factors.ts
isaaccodekill: These are the units that every emission factor should be in. I and @mfonsecaOEF  agreed that the emissions factors should all be in the same units. 

@mfonsecaOEF  is still working on converting all EFs to these units. But until she's done,  i'm filtering out EFs that are not these units so that the calculations will be accurate

app/src/backend/formulas.ts
isaaccodekill: the input methodology will always be required for conversions because we need to access the methodology in the manual input json file

lemilonkh: Yes, I meant doing a null check and throwing/ logging an error here instead of effectively using the `!` Typescript postfix operator by using `as string`

lemilonkh: Checking it and returning/ throwing in that case means it is no longer considered nullable in the current block by the TS compiler, meaning you can leave out `as string` in all of those places.

lemilonkh (approved):
Thanks ✨

app/src/backend/formulas.ts
isaaccodekill: noted and corrected. THanks

PR #882, opened 2024-11-06
[ON-2770] fix public dashboard URLs

thehighestprimenumber:
use the window.location to build the URLs


app/src/components/HomePage/DownloadAndShareModals/PublishedView.tsx
```diff
@@ -46,16 +46,9 @@ export function PublishedView({
...
-  } catch (e) {
-  } finally {
-    console.log(
-      "process.env.NEXT_PUBLIC_HOST",
-      JSON.stringify(process.env.NEXT_PUBLIC_HOST),
-    ); // TODO NINA
-  }
-  const URL = `${process.env.NEXT_PUBLIC_HOST}/${lng}/public/${inventoryId}`;
+
+  const URL = `${window.location.protocol}//${window.location.host}/${lng}/public/${inventoryId}`;
```
lemilonkh, line 50: I think you need the port after a : after the host in case there is a port number other than 80 in use

lemilonkh (approved):
Great, thanks 👍✨

PR #883, opened 2024-11-06
basic plus extra fields

mfonsecaOEF:


app/src/components/Modals/activity-modal/activity-modal-body.tsx
```diff
@@ -233,7 +222,7 @@ const ActivityModalBody = ({
                 {f.options && (
                   <FormControl key={idx} className="w-full">
                     <BuildingTypeSelectInput
-                      options={f.options}
+                      options={f.options as string[]}
```
lemilonkh, line 225: I think we need to map or change the data type of this property here if the data is in a different format now?

app/src/util/form-schema/index.ts
```diff
@@ -25,7 +25,7 @@ export function findMethodology(
 export interface ExtraField {
   id: string;
   type?: string;
-  options?: string[];
+  options?: string[] | { type: string; names: string[] }[];
```
lemilonkh, line 28: Ideally we only have one of these and rework the existing data to also use this schema. Otherwise we need to have an if/ else block for both cases in every place this is used...

lemilonkh: We could potentially have a helper function that normalizes it, but at that point let's just use my automatic rewriting script for the JSON and fix all of the data types to be consistent.

lemilonkh (approved):
Amazing, thanks :raised_hands: :sparkles:

thehighestprimenumber:
let's not merge this until we meet with Evan

lemilonkh:
@evanp is this safe to merge now?

evanp:
Gang, I'm not going to be able to review a PR in the hour before demo. I'm not sure what went wrong here. Let's plan on reviewing it tomorrow. I'm really disappointed that this ticket won't be finished.

lemilonkh:
This PR will be put on hold until we have implemented the logic for the dependent choices

PR #885, opened 2024-11-07
bug: [ON-2771] fix csv download

thehighestprimenumber:


app/src/app/api/v0/inventory/[inventory]/download/route.ts
```diff
@@ -53,9 +53,6 @@ export const GET = apiHandler(async (req, { params, session }) => {
           {
             model: db.models.GasValue,
             as: "gasValues",
-            include: [
-              { model: db.models.EmissionsFactor, as: "emissionsFactor" },
-            ],
```
Comment (lemilonkh) [inline app/src/app/api/v0/inventory/[inventory]/download/route.ts:58]: Sorry this was a bit of miscommunication.

We need to change this whole GasValue block to be inside of a (to be added) block for ActivityValue. So we can get the actual gas values associated with the activity here, and also their emissions factors.

lemilonkh (changes requested):

app/src/app/api/v0/inventory/[inventory]/download/route.ts
```diff
@@ -51,12 +51,22 @@ export const GET = apiHandler(async (req, { params, session }) => {
...
+                as: "gasValues",
+                include: [
+                  { model: db.models.EmissionsFactor, as: "emissionsFactor" },
+                ],
+              },
             ],
           },
+          {
+            model: db.models.GasValue,
+            as: "gasValues",
```
Comment (lemilonkh) [inline app/src/app/api/v0/inventory/[inventory]/download/route.ts:68]: These will not contain any data with our current way to save the manual input data. So we can either fix that later or remove/ comment out this block for now

lemilonkh (approved):
Thanks :sparkles:

PR #886, opened 2024-11-07
[ON-2327] fix displayed scopes in add data

thehighestprimenumber:


app/src/app/[lng]/[inventory]/data/[step]/[subsector]/page.tsx
```diff
@@ -129,14 +134,17 @@ function SubSectorPage({
...
     return Object.entries(MANUAL_INPUT_HIERARCHY)
       .filter(([key]) => key.startsWith(subSectorData?.referenceNumber!))
-      .map(([k, v]) => ({
-        ...v,
-        referenceNumber: k,
-      }));
+      .map(([k, v]) => ({ ...v, referenceNumber: k }))
+      .filter((scope) =>
+        getScopesForInventoryAndSector(
+          inventoryData.inventoryType!,
```
Comment (lemilonkh) [inline app/src/app/[lng]/[inventory]/data/[step]/[subsector]/page.tsx:143]: Can we do a null check beforehand and throw a more useful error here in case it is null-ish?

app/src/app/[lng]/[inventory]/data/[step]/[subsector]/page.tsx
```diff
@@ -129,14 +134,17 @@ function SubSectorPage({
...
+          scope.referenceNumber[0],
+        )!.includes(scope.scope),
```
Comment (lemilonkh) [inline app/src/app/[lng]/[inventory]/data/[step]/[subsector]/page.tsx:145]: Should we handle this null error instead of ignoring it with `!`? Just a default value of [] would be sufficient here (or throwing an error in case that's better).

lemilonkh (approved):
Thanks 🙌✨

app/src/app/[lng]/[inventory]/data/[step]/[subsector]/page.tsx
Comment (thehighestprimenumber) [reply app/src/app/[lng]/[inventory]/data/[step]/[subsector]/page.tsx:143]: there should never be a case where it's null because it has a default setting in the DB. and also I'm checking that inventoryData is there

Comment (lemilonkh) [reply app/src/app/[lng]/[inventory]/data/[step]/[subsector]/page.tsx:143]: Ok great, then we can adjust the type definition in the model file so it's clear that it's not null? We can do it in a follow-up tech debt ticket next sprint if necessary.

Comment (thehighestprimenumber) [reply app/src/app/[lng]/[inventory]/data/[step]/[subsector]/page.tsx:143]: https://openearth.atlassian.net/browse/ON-2851

PR #902, opened 2024-11-17
[ON-2906] fix: remove "Add Data" button on public page

thehighestprimenumber:
before:

image
image

after:
image
image

### DIFF
--- a/app/src/components/HomePage/DownloadAndShareModals/PublishedView.tsx
+++ b/app/src/components/HomePage/DownloadAndShareModals/PublishedView.tsx
@@ -47,7 +47,7 @@ export function PublishedView({
       locale: getLocale(lng),
     });
 
-  const URL = `${window.location.protocol}//${window.location.host}${window.location.port ? `:${window.location.port}` : ""}/${lng}/public/${inventoryId}`;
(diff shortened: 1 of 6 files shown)


app/src/components/HomePage/DownloadAndShareModals/PublishedView.tsx
```diff
@@ -47,7 +47,7 @@ export function PublishedView({
       locale: getLocale(lng),
     });
 
-  const URL = `${window.location.protocol}//${window.location.host}${window.location.port ? `:${window.location.port}` : ""}/${lng}/public/${inventoryId}`;
+  const URL = `${window.location.protocol}//${window.location.host}/${lng}/public/${inventoryId}`;
```
lemilonkh, line 50: Why was the port removed here?

lemilonkh (approved):
Thanks :sparkles:

app/src/components/HomePage/DownloadAndShareModals/PublishedView.tsx
thehighestprimenumber: because it was causing the url to be localhost:3000:3000. I tested with other ports and it still works.

PR #905, opened 2024-11-17
[ON-2911] add translations for direct measure

thehighestprimenumber:


lemilonkh (approved):
Thanks :raised_hands:

PR #906, opened 2024-11-17
fix: [ON-2903] remove hardcoded size for Ask Ai button

thehighestprimenumber:
to adapt to different text lengths in different languages


lemilonkh (approved):
Thanks :sparkles:

PR #913, opened 2024-11-18
[ON-2892] fix: issue with loading IPPU and AFOLU

thehighestprimenumber:
The error was caused because these sectors don't have a subcategories. Changed the logic to look for the subcategory and if not found, look for the subsector directly.

### DIFF
--- a/app/src/backend/GPCService.ts
+++ b/app/src/backend/GPCService.ts
@@ -5,7 +5,7 @@ import { QueryTypes } from "sequelize";
 type IDsFromReferenceNumberResult = {
   sectorId: string;
   subSectorId: string;
-  subCategoryId: string;
+  subCategoryId: string | null;
 };
 
 export default class GPCService {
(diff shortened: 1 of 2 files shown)


app/src/backend/GPCService.ts
```diff
@@ -30,18 +30,34 @@ export default class GPCService {
...
-
-    if (!sectorId || !subSectorId || !subCategoryId) {
-      throw new createHttpError.BadRequest(
-        "Couldn't find sector/ subsector/ subcategory for given GPC reference number",
-      );
+    if (subcategory) {
+      return {
+        sectorId: subcategory?.subsector?.sector?.sectorId,
+        subSectorId: subcategory?.subsector?.subsectorId,
+        subCategoryId: subcategory?.subsector?.sector?.sectorId,
```
lemilonkh, line 37: You can leave out the question mark after subcategory in these lines as you have just checked that is truthy with the if statement you're in.

lemilonkh (approved):
Amazing, thanks 🙌✨

isaaccodekill (approved):

PR #916, opened 2024-11-20
Backend Endpoints To Support Multi Inventory

isaaccodekill:
- Contains City Breakdown endpoints
- Contains year on year emissions statistics endpoint


app/src/components/HomePage/HomePage.tsx
```diff
@@ -150,7 +162,15 @@ export default function HomePage({
...
-            <Box className="flex mx-auto max-w-full w-[1090px] css-0">
+            <Box className="mx-auto max-w-full w-[1090px] css-0">
+              <Text
+                color="content.primary"
+                fontWeight="bold"
+                lineHeight="24px"
+                fontSize="headline.sm"
+                fontFamily="heading"
+                fontStyle="normal"
+              ></Text>
```
lemilonkh, line 173: What is this empty Text component for?

app/src/models/Inventory.ts
```diff
@@ -163,6 +166,11 @@ export class Inventory
...
           field: "global_warming_potential_type",
         },
+        lastUpdatedAt: {
+          type: DataTypes.DATE,
+          allowNull: true,
+          field: "last_updated_at",
+        },
       },
       {
         sequelize,
```
lemilonkh, line 176: We have the `last_updated` timestamp defined below and we can manually overwrite it in other queries/ endpoints if necessary. Why do we need 2 last update timestamp fields?

app/src/models/InventoryValue.ts
```diff
@@ -238,6 +238,24 @@ export class InventoryValue
...
+              if (inventory) {
+                await inventory.update({ lastUpdatedAt: new Date() });
+              }
+            }
+          },
+          afterUpdate: async (inventoryValue: InventoryValue) => {
+            if (inventoryValue.inventoryId) {
+              const inventory = await inventoryValue.getInventory();
+              if (inventory) {
+                await inventory.update({ lastUpdatedAt: new Date() });
```
lemilonkh, line 254: Can we just set lastUpdated here and add it to the model instead of lastUpdateAt if necessary, as the DB column exists already...

lemilonkh (approved):
Thanks 🙌

PR #918, opened 2024-11-22
bug: [ON-2881] deleteUserCityCascade

thehighestprimenumber:

Implement a cascading delete for the `city_id` foreign key constraint in the `Inventory` table and update the application logic to handle related user default inventory settings on city deletion.

These changes address a bug where deleting a city could leave orphaned records in the database and potentially break user-related inventory associations. By ensuring cascading deletes on the `Inventory` table and updating user default inventory preferences, data integrity is maintained and user experience is improved. This solution was chosen as it is the most efficient way to maintain relational integrity and handle user preferences upon city deletion.


app/migrations/20241121140052-inventory__alter__inventory_city__cascade.cjs
lemilonkh, line 1: Do we need the double underscores in the file name here?

lemilonkh (approved):
Thanks ✨

PR #922, opened 2024-11-23
fix: [ON-2929] add translations

thehighestprimenumber:


lemilonkh (approved):
Great German translations as well, thanks 👍🏼✨

PR #930, opened 2024-11-25
fix: [ON-2930] show sectors IV and V in report results

thehighestprimenumber:


app/src/i18n/locales/en/dashboard.json
```diff
@@ -123,5 +123,7 @@
   "ghg-emissions-inventory-in-year": "GHG emissions inventory in {{ year }}",
   "set-inventory-preferences": "Set inventory preferences",
   "discover-relevant-subsectors": "Discover the GPC sub-sectors relevant to your city and tailor your inventory to fit your city's specific needs.",
+  "agriculture,-forestry,-and-other-land-use-(afolu)": "AFOLU",
+  "industrial-processes-and-product-uses-(ippu)": "IPPU",
```
lemilonkh, line 127: Do we need these for the other languages as well? Doesn't have to be in this PR of course.

lemilonkh (approved):
Thanks 👍🏼

PR #931, opened 2024-11-25
[ON-2785] use default inventory without redirecting

thehighestprimenumber:


lemilonkh (approved):
Thanks, this was long overdue! :sparkles:

PR #933, opened 2024-11-26
update(ui): changes layout and text for email template, removes used props

cephaschapa:
Changes
- Changes and styles text in email message body
- Removes unused props

### DIFF
--- a/app/src/lib/emails/InviteUserTemplate.tsx
+++ b/app/src/lib/emails/InviteUserTemplate.tsx
@@ -1,5 +1,7 @@
+import { useCopyToClipboard } from "@/hooks/use-copy-to-clipboard";
 import { CityAttributes } from "@/models/City";
 import { UserAttributes } from "@/models/User";
+import React, { useState } from "react";
 import {
   Body,
   Button,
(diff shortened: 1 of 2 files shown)


app/src/lib/emails/InviteUserTemplate.tsx
```diff
@@ -15,19 +17,13 @@ import {
...
 }: {
   url?: string;
-  user?: { name: string; email: string; cityId?: string };
-  city?: CityAttributes;
-  invitingUser?: { name: string; email: string };
-  members: UserAttributes[];
+  user?: { name: string };
 }) {
-  const countryCode = city?.locode?.split(" ")!;
+  const generatedCode = "2KGuWexgH@!";
```
lemilonkh, line 26: We should probably generate this using base64 or another approach to generate random codes. And we should do that outside of the email template and save it.

app/src/lib/emails/InviteUserTemplate.tsx
lemilonkh, line 1: I think we should probably keep both templates instead of changing this one?

Since a user inviting another user to a city is still a valid use case, this new template is just for giving people access to CityCatalyst itself with an invite code.

Which btw is something we probably could do manually for now...

lemilonkh (changes requested):
Not sure about this one, let's discuss

PR #936, opened 2024-11-28
[ON-2997] fix redirect when user has no defaultInventoryId (new user)

thehighestprimenumber:


evanp (approved):
Lgtm! Thanks for the quick work.

PR #944, opened 2024-12-02
fix(ui): adds translation for new password label

cephaschapa:
Changes
- Adds translation string to update password - new password label
- Also adds translations for en, es, pt

image


lemilonkh (approved):
Thanks 👍🏼✨

PR #947, opened 2024-12-02
[ON-2997] navBar on first load

thehighestprimenumber:
Navbar wasn't loading when we were loading the default inventory id without redirecting to the /[inventory] page.


lemilonkh (approved):
Thanks :sparkles:

PR #950, opened 2024-12-03
feat: adds circle flag component and  locode extraction function

cephaschapa:
Changes
- Replaces hardcoded flag icon with CircleFlag component
- Adds helper function to extract the country locode from `cityLocode` string

### DIFF
--- a/app/src/components/Tabs/my-inventories-tab.tsx
+++ b/app/src/components/Tabs/my-inventories-tab.tsx
@@ -45,7 +45,6 @@ import type { CityAttributes } from "@/models/City";
 import { api } from "@/services/api";
 import type { InventoryAttributes } from "@/models/Inventory";
 import { CircleFlag } from "react-circle-flags";
-import { extractCountryLocode } from "@/util/helpers";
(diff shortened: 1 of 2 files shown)


app/src/components/Tabs/my-inventories-tab.tsx
```diff
@@ -178,10 +180,12 @@ const MyInventoriesTab: FC<MyInventoriesTabProps> = ({
                   >
                     <Box>
                       <Box display="flex" gap="8px" alignItems="center">
-                        <Avatar
-                          className="h-[32px] w-[32px]"
-                          name="Argentina"
-                          src="https://upload.wikimedia.org/wikipedia/commons/1/1a/Flag_of_Argentina.svg"
+                        <CircleFlag
+                          countryCode={extractCountryLocode(
+                            city.countryLocode!,
```
lemilonkh, line 185: You are passing the country locode, not the city locode to this function. So I think the call to extractCountryLocode is not required or you need to pass the city locode into it...

PR #951, opened 2024-12-03
Fix/add padding to ask ai btn

cephaschapa:
Changes
- Adds padding to ask AI button, (prevously we had hard coded dimensions hence the overflow) instead a padding is added to evenly adjust the button proportions.


lemilonkh (approved):
Thanks

PR #953, opened 2024-12-04
fix(ui): resolves checkmark bug for all population inputs

cephaschapa:
Changes
- Adds uses watch to track population and year values to each input
- Uses tracked values to set the check icon component

### DIFF
--- a/app/src/components/steps/add-population-data-step.tsx
+++ b/app/src/components/steps/add-population-data-step.tsx
@@ -130,6 +130,15 @@ export default function SetPopulationDataStep({
     }
   }, [countryData, year, setValue]);
 
+  const cityPopulation = watch("cityPopulation");
+  const cityPopulationYear = watch("cityPopulationYear");
+
+  const regionPopulation = watch("regionPopulation");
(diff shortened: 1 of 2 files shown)


app/src/components/steps/add-population-data-step.tsx
```diff
@@ -243,7 +252,9 @@ export default function SetPopulationDataStep({
                   mt={5}
                   mr={6}
                 >
-                  {!!year && <CheckIcon color="semantic.success" boxSize={4} />}
+                  {!!countryPopulationYear && countryPopulation && (
+                    <CheckIcon color="semantic.success" boxSize={4} />
+                  )}
```
lemilonkh, line 257: I wonder if this behavior could be extracted into a reusable component instead of copied a bunch of times for all inputs...

Same with the watch calls.

cephaschapa: That possible aswell

lemilonkh: I meant the whole input but this is a good start :+1:

lemilonkh (approved):
Thanks :sparkles:

PR #967, opened 2024-12-09
fix(api): count all InventoryValues with a source as third party

lemilonkh:

Update the logic in `ECRFDownloadService.ts` and `InventoryProgressService.ts` to count inventory values as third party if a data source exists, removing the explicit check for source type as "third_party".

The change simplifies the condition by assuming any existing data source is third party, removing the need to check the specific source type. This streamlines the code and accommodates various scenarios where data could be categorized under different source types not strictly labeled as "third_party".


isaaccodekill (approved):
Amazing, thanks!

PR #970, opened 2024-12-10
fix: add ON DELETE CASCADE to foreign key of GasValue

evanp:
We have an error in the deletion of ActivityValue entries because of the foreign key to GasValue.

This migration adds an ON DELETE CASCADE and ON UPDATE CASCADE to the foreign key.

Add ON DELETE CASCADE action to the foreign key constraint on the `activity_value_id` column of the `GasValue` table in the database migration script.

These changes ensure that when a related record in the `ActivityValue` table is deleted, the corresponding entries in the `GasValue` table will automatically be removed, maintaining referential integrity and preventing orphaned records. This decision mitigates potential data inconsistencies and adheres to best practices in relational database management.

### DIFF
--- a/app/migrations/20241210173228-gasvalue-activityvalue-on-delete-cascade.cjs
+++ b/app/migrations/20241210173228-gasvalue-activityvalue-on-delete-cascade.cjs
@@ -1,0 +1,36 @@
+
+"use strict";
+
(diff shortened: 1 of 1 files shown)


lemilonkh (approved):
Thanks 👍✨

PR #971, opened 2024-12-11
fix(api): load InventoryValues in emissions by scope

lemilonkh:

Modify the emissions calculation to include third-party emissions by loading `InventoryValues` scoped to each inventory in the results service and update handling in the API route layer.

This change is necessary to ensure accurate emissions reporting by including emissions data from third-party sources when internal activity values are absent. By refactoring the API route and results service, the system can now calculate emissions comprehensively, addressing potential data gaps and improving reporting accuracy.

### DIFF
--- a/app/src/backend/ResultsService.ts
+++ b/app/src/backend/ResultsService.ts
@@ -457,11 +457,9 @@ async function calculateThirdPartyEmissionsByScope(
 
     const co2eq = value.co2eq ?? 0n;
     const totalEmissions = bigIntToDecimal(co2eq);
+
     return {
-      // TODO use SubCategory name instead?
-      // TODO i18n
-      activityTitle:
-        value.dataSource?.datasetName?.en ?? "Unknown third-party source",
+      activityTitle: "mixed-activities",
       scopes: {
         [scopeName]: bigIntToDecimal(co2eq),
       },
--- a/app/src/backend/ResultsService.ts
+++ b/app/src/backend/ResultsService.ts
@@ -1,5 +1,5 @@
 import { db } from "@/models";
-import { QueryTypes } from "sequelize";
+import { Op, QueryTypes } from "sequelize";
 import { MANUAL_INPUT_HIERARCHY } from "@/util/form-schema";
 import groupBy from "lodash/groupBy";
 import mapValues from "lodash/mapValues";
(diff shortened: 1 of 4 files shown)


app/src/backend/ResultsService.ts
```diff
@@ -426,6 +433,38 @@ function convertEmissionsToStrings(
...
+    const co2eq = value.co2eq ?? 0n;
+    const totalEmissions = bigIntToDecimal(co2eq);
+    return {
+      activityTitle: "third-party", // TODO use data source name or SubCategory name?
+      scopes: {
+        [scopeName]: bigIntToDecimal(co2eq),
+      },
+      totalEmissions,
+      percentage: 100,
+    };
```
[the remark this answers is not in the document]
lemilonkh: @korbit-ai there is only one activity for the whole inventory value in the case of third party data associated with a whole GPC reference number.

app/src/backend/ResultsService.ts
```diff
@@ -426,6 +436,36 @@ function convertEmissionsToStrings(
...
```
isaaccodekill, line 463: Shouldn't this be limited to inventory values with a data source id? Especially if the goal is to get a scope breakdown of thirdparty emissions

isaaccodekill (approved):
Amazing work

app/src/backend/ResultsService.ts
```diff
@@ -426,6 +436,36 @@ function convertEmissionsToStrings(
...
+    if (!scopeName) {
+      throw new createHttpError.InternalServerError(
+        "Scope name not found for inventory value " + value.id,
+      );
+    }
+
+    const co2eq = value.co2eq ?? 0n;
+    const totalEmissions = bigIntToDecimal(co2eq);
+    return {
+      activityTitle: "third-party", // TODO use data source name or SubCategory name?
```
thehighestprimenumber, line 457: Dino wanted to call this "Mixed activities"

thehighestprimenumber (approved):
just a comment about naming

app/src/backend/ResultsService.ts
lemilonkh: Good point, done :+1: I'm also using the dataset name instead of `third-party` now.

PR #972, opened 2024-12-11
[ON-2881] fix deletion of cities with invites

thehighestprimenumber:

Implement cascading delete for the "CityInvite" table by updating the foreign key constraint to allow invites to be automatically deleted when a city is removed.

To ensure referential integrity and prevent orphaned records in the "CityInvite" table when a city is deleted, which resolves issues where invites remained without an associated city. This migration adjusts the foreign key constraint to cascade deletions, providing a cleaner and more automated maintenance process.


lemilonkh (approved):
Good spot, thanks :sparkles:

PR #978, opened 2024-12-11
ON-3073 sector percentages fix for basic+

isaaccodekill:
Now that scope 3 has been added to basic+ this pr includes changes to accommodate that

Refactor the calculation of subcategory counts for sectors to accurately account for inventory types, and mark the `datasourceId` field in the `InventoryValue` model as deprecated.

The previous implementation of subcategory counting did not properly consider the specific scopes applicable to an inventory type, which could lead to incorrect sector percentages for GPC_BASIC_PLUS. The refactoring ensures that only applicable subcategories are counted based on inventory type, thus improving accuracy. Additionally, `datasourceId` has been moved to `ActivityValue`, and the field in the `InventoryValue` model is marked deprecated for eventual removal, reflecting data model changes for better organization.

### DIFF
--- a/app/src/models/InventoryValue.ts
+++ b/app/src/models/InventoryValue.ts
@@ -28,7 +28,8 @@ export interface InventoryValueAttributes {
   subSectorId?: string;
   subCategoryId?: string;
   inventoryId?: string;
-  datasourceId?: string | null;
+  /** @deprecated moved to ActivityValue */
+  datasourceId?: string | null; // TODO remove
   created?: Date;
   lastUpdated?: Date;
 }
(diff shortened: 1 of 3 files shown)


app/src/models/InventoryValue.ts
```diff
@@ -28,7 +28,8 @@ export interface InventoryValueAttributes {
   subSectorId?: string;
   subCategoryId?: string;
   inventoryId?: string;
-  datasourceId?: string | null;
+  /** @deprecated moved to ActivityValue */
+  datasourceId?: string | null; // TODO remove
```
lemilonkh, line 32: This change can be reverted as the important change was already merged and it is no longer deprecated.

lemilonkh (approved):
Amazing work, thanks 😊👍

app/src/models/InventoryValue.ts
evanp: Are we blocking on this or can we merge?

PR #982, opened 2024-12-12
fix(ui): adds condition to show na for zero values

cephaschapa:
Changes
- adds condition to show N/A instead of zero values
- uses same condition to render correct styling
image

Add a condition to display "N/A" for percentage values that are zero in the EmissionBySectorTable component.

To improve the user interface by clearly indicating when a sector's emission percentage is zero, avoiding potential confusion caused by displaying zero with an upward or downward icon. The approach ensures that zero values are displayed as "N/A", which more accurately reflects the data context.


lemilonkh (approved):
Thanks 👍✨
The bot comment about renaming the variable seems good, otherwise good to merge imo 👍

PR #986, opened 2024-12-12
WIP: feat: add a custom server for logging ON-3130

evanp:
The only way I could find to get logging of all our HTTP requests was to create a custom server that loads NextJS itself. 

https://nextjs.org/docs/pages/building-your-application/configuring/custom-server

I don't think this is the path we want to take; it means we lose a lot of NextJS features.

The next best solution I've seen is patching the NextJS library and adding your logging code directly to it! Which is definitely a lot worse.

If someone else gets a good idea of how to do detailed server-side logging on the app, please let me know. Right now I'm stumped.

Add a custom server for logging HTTP requests using `pino-http` in the `server.ts` file and update dependencies accordingly in `package.json` and `package-lock.json`.

These changes introduce structured logging to our server for better monitoring and debugging capabilities, using `pino-http` for efficient HTTP request logs. This enhancement aligns with ON-3130 user story requirements, ensuring that any errors and significant events can be logged at appropriate levels for improved observability.

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -67,6 +67,7 @@
         "pg-hstore": "^2.3.4",
         "pigeon-maps": "^0.21.6",
         "pino": "^9.3.2",
+        "pino-http": "^10.3.0",
         "postcss": "8.4.31",
         "react": "18.3.1",
         "react-circle-flags": "^0.0.20",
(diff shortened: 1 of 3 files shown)


evanp:
I think this is a bad solution to a small problem, so I'm closing it.

PR #990, opened 2024-12-13
fix(api): don't return a 404 error when no emissions factors are found

lemilonkh:
The API route `/api/v0/emissions-factor` now returns an empty array instead of a 404 error when no emissions factors are found.

Remove the 404 error when no emissions factors are found and ensure the API returns a JSON response with an empty data set instead.

Previously, the API would return a 404 error if no emissions factors were found, which implied client error, though it is valid for some queries to yield no results. Returning an empty JSON makes the API more robust and offers an accurate representation of a search that yields no data, facilitating better client handling of such cases.


isaaccodekill (approved):

thehighestprimenumber (approved):
great!

PR #993, opened 2024-12-13
[ON-3137] fix bigint conversion

thehighestprimenumber:

Correct the conversion of a BigInt from a floating-point number by using the `toFixed(0)` method to ensure no fractional part remains before conversion.

The changes address a bug where fractional values could inadvertently affect conversion to BigInt, potentially causing errors or incorrect gas amount calculations. Using `toFixed(0)` ensures the number is properly formatted as an integer string, eliminating fractions before `BigInt` conversion, and thus enhancing reliability and correctness in gas processing logic.


lemilonkh (approved):
Good spot, thanks 👍✨

thehighestprimenumber:
I think it's the wrong solution

PR #994, opened 2024-12-13
[ON-3137] fix bigint conversion

thehighestprimenumber:

Fix the conversion of the `amount` to a number before converting to BigInt in `ActivityService.ts` by using `toFixed(0)`.

The previous conversion method might have led to precision issues or conversion errors since `toNumber()` alone might not provide the desired integer format for BigInt conversion. Using `toFixed(0)` ensures the `amount` is properly formatted as an integer string before conversion, thus preventing potential runtime errors or incorrect values.


lemilonkh (approved):
Thanks 👍

PR #995, opened 2024-12-13
fix: correct units difference between country and city total emissions

evanp:
The units we pull from OpenClimate for total country emissions are in tonnes, and we use kilograms in CityCatalyst. I updated the display code to correct it.

Correct the units difference between country total emissions and city total emissions by converting the country total emissions from tonnes to kilograms for accurate percentage calculation.

Previously, there was a unit mismatch where country emissions were in tonnes, and city emissions were in kilograms, causing incorrect percentage calculations. By multiplying the country emissions by 1000, both values are now in kilograms, ensuring accurate comparison and calculation.


lemilonkh (approved):
Thanks good spot. This should eventually be fixed during onboarding, good for now though 👍

PR #998, opened 2024-12-13
[ON-3144] fix save multiple direct measures

thehighestprimenumber:
Modify the query to save multiple direct measures by adding `inventoryId` as a condition in InventoryValue lookup.

Previously, saving multiple direct measures failed because the system only checked for `gpcReferenceNumber`, which could be associated with multiple `inventoryId`s, leading to incorrect conflict detection. The inclusion of `inventoryId` ensures the query accurately identifies unique records, preventing false positives in conflict checks.


lemilonkh (approved):
Amazing, thanks 👍✨

PR #999, opened 2024-12-13
[ON-3037] fix emissions breakdown when there's an NA activity

thehighestprimenumber:

Fix the emissions breakdown by excluding inventory values where CO2eq is NULL.

An issue was identified where activities with a NULL CO2eq value were incorrectly being included in emissions breakdown results, leading to "cannot convert null to bigint". The condition `iv.co2eq IS NOT NULL` ensures only relevant data is considered in the calculations, thereby improving the accuracy of the emissions breakdown.


evanp (approved):
This looks right.

PR #1002, opened 2024-12-16
CSV output updates

isaaccodekill:
- Add new functionality for exporting inventory data to CSV by introducing `CSVDownloadService`, updating dependencies to include `csv-stringify`, and refactoring existing types and services to support CSV formatting.

- Include new columns for csv output
```
[
"Inventory Reference",
"GPC Reference Number",
"Subsector name",
"Notation Key",
"Total Emissions",
"Activity Value",
"Activity Units",
"Emission Factor - CO2",
"Emission Factor - CH4",
"Emission Factor - N2O",
"Emission Factor - Unit",
"Datasource ID",
"DataSource name",
]
```

These changes enable users to download inventory data in CSV format, a common and widely supported file type that simplifies data manipulation and visualization. The addition of `CSVDownloadService` centralizes the CSV generation logic, keeping the code well-organized and extensible.

### DIFF
--- a/app/src/backend/CSVDownloadService.ts
+++ b/app/src/backend/CSVDownloadService.ts
@@ -1,0 +1,186 @@
+import { InventoryWithInventoryValuesAndActivityValues } from "@/util/types";
+import { sortGpcReferenceNumbers, toDecimal } from "@/util/helpers";
+import Decimal from "decimal.js";
+import { translationFunc } from "@/i18n/server";
+import { stringify } from "csv-stringify/sync";
+
+export default class CSVDownloadService {
+  public static async downloadCSV(
+    output: InventoryWithInventoryValuesAndActivityValues,
+    lng: string,
+  ) {
+    const headers = [
+      "Inventory Reference",
+      "GPC Reference Number",
+      "Subsector name",
+      "Notation Key",
+      "Total Emissions",
+      "Activity Value",
+      "Activity Units",
+      "Emission Factor - CO2",
+      "Emission Factor - CH4",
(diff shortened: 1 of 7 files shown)


app/src/backend/CSVDownloadService.ts
```diff
@@ -0,0 +1,186 @@
...
+            }
+
+            return {
+              emission_factor_unit: efUnit,
+              emission_co2,
+              emission_ch4,
+              emission_n2o,
+              activity_amount: activityAmount,
+              activity_unit: activityUnit,
+              data_source_id: "N/A",
```
lemilonkh, line 170: Leaving the field empty in this case might make sense, as if we parse it automatically having N/A strings might complicate things?

lemilonkh (approved):
Amazing thanks 👍✨

app/src/backend/CSVDownloadService.ts
isaaccodekill: Good point!

app/src/backend/CSVDownloadService.ts
```diff
@@ -0,0 +1,186 @@
...
+                ?.div(new Decimal("1e3"))
+                .toNumber(),
+            };
+          }),
+        };
+      }
+
+      if (dataDictionary[gpcRefNo as string].notation_key) {
+        dataDictionary[gpcRefNo as string].activityValues = [{}];
+      }
```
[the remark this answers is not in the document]
isaaccodekill: in this case, it's not important to care about previously set data

PR #1014, opened 2024-12-17
fix(ui): sets population validation input to explicitly accept numbers

cephaschapa:
Changes
- Adds more validation rules to population inputs
- During testing there was a special case while adding and inventory for Vicente Lopez that resulted in a valiation error response from the server.

image


lemilonkh (approved):
Great, thanks 👍

PR #1022, opened 2024-12-18
[ON-2881] Fix delete inventory

thehighestprimenumber:
Add functionality to delete an inventory through the DeleteInventoryModal component with password confirmation.

The feature update resolves the need for secure inventory deletion within the application by integrating API requests for password verification and deletion, and by providing user feedback upon successful or unsuccessful attempts. This enhances the user experience by ensuring necessary security protocols are followed and confirmations are communicated effectively.

### DIFF
--- a/app/src/services/api.ts
+++ b/app/src/services/api.ts
@@ -12,6 +12,7 @@ import {
   ConnectDataSourceResponse,
   EmissionsFactorResponse,
   GetDataSourcesResult,
+  InventoryDeleteQuery,
   InventoryProgressResponse,
   InventoryResponse,
   InventoryUpdateQuery,
@@ -48,6 +49,7 @@ export const api = createApi({
     "SectorBreakdown",
     "Inventory",
     "CitiesAndInventories",
+    "Inventories",
   ],
   baseQuery: fetchBaseQuery({ baseUrl: "/api/v0/", credentials: "include" }),
   endpoints: (builder) => ({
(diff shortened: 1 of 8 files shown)


app/src/services/api.ts
```diff
@@ -48,6 +49,7 @@ export const api = createApi({
     "SectorBreakdown",
     "Inventory",
     "CitiesAndInventories",
+    "Inventories",
```
lemilonkh, line 52: Can we use the Inventory tag we have already here? Normally we don't distinguish between singular/ plural in these tags...

app/src/services/api.ts
```diff
@@ -289,6 +291,18 @@ export const api = createApi({
...
+      InventoryAttributes,
+      InventoryDeleteQuery
+    >({
+      query: (data) => ({
+        url: `/inventory/${data.inventoryId}`,
+        method: "DELETE",
+      }),
+      transformResponse: (response: { data: InventoryAttributes }) =>
+        response.data,
+      invalidatesTags: ["InventoryProgress", "InventoryValue", "Inventories"],
```
lemilonkh, line 304: Technically this also invalidates Inventory, ActivityValue, InventoryValue, ReportResults, YearlyReportResults...

Probably won't cause issues for now as you will be on another inventory page or onboarding afterwards but still 😅

lemilonkh (approved):
Thanks 👍

app/src/services/api.ts
thehighestprimenumber: I had to add the Inventories so that the list of inventories gets updated without the inventory you just deleted

PR #1024, opened 2024-12-19
[ON-3176] invalidate cache for report results when disconnecting 3rd party data

thehighestprimenumber:
Invalidate the cache for "ReportResults" when disconnecting third-party data sources.

This change ensures that any cached report results are cleared and refreshed when a third-party data source is disconnected, preventing users from viewing stale data that may no longer be relevant or accurate.


lemilonkh (approved):
Thanks 👍✨

PR #1030, opened 2024-12-30
fix(ui): adds formcontrol to prevent default form action

cephaschapa:
Changes
- Adds `FormControl `component to handle default form actions gracefully. In this case each time the enter/return key is pressed the form gets submitted and pushes the password query param shows up in the URL

Replace the default HTML `<form>` tag with a Chakra UI `<FormControl>` component in the `DeleteInventoryModal` component to prevent default form actions.

The default form behavior was unintentionally submitting or redirecting the page, causing disruptions in the user experience. Switching to `FormControl` aligns with the UI library's practices and prevents these actions, providing a more consistent and reliable modal behavior.

### DIFF
--- a/app/src/components/Modals/delete-inventory-modal.tsx
+++ b/app/src/components/Modals/delete-inventory-modal.tsx
@@ -4,6 +4,7 @@ import {
   Badge,
   Box,
   Button,
+  FormControl,
   Modal,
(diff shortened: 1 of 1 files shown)


thehighestprimenumber (approved):
lgtm!

PR #1050, opened 2025-01-09
bug: [ON-2864] handle missing data from globalAPI &  rename module

thehighestprimenumber:

Handle missing data from GlobalAPI by returning `undefined` when response status is not 200, and rename the module from `OpenClimateService` to `GlobalAPIService`.

These changes address a bug where the application crash if the GlobalAPI does not return data successfully, improving robustness by returning `undefined` to handle non-200 responses. The module is renamed to better reflect its purpose and promote clearer codebase organization and maintainability.

### DIFF
--- a/app/src/backend/GlobalAPIService.ts
+++ b/app/src/backend/GlobalAPIService.ts
@@ -13,6 +13,9 @@ export const getGrowthRatesFromOC = async (
     const response = await fetch(URL);
     logger.info(`${URL} Response Status: ${response.status}`);
     const data = await response.json();
+    if (response.status !== 200) {
+      return undefined;
+    }
     return {
       ...data,
       growthRates: data.growth_rates,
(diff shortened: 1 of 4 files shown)


app/src/backend/GlobalAPIService.ts
```diff
@@ -13,6 +13,9 @@ export const getGrowthRatesFromOC = async (
     const response = await fetch(URL);
     logger.info(`${URL} Response Status: ${response.status}`);
     const data = await response.json();
+    if (response.status !== 200) {
```
lemilonkh, line 16: Would be good to also rename the method name or remove the "fromOC" suffix from it (since we are apparently loading the data from the global API and not OpenClimate).

lemilonkh (approved):

app/src/backend/GlobalAPIService.ts
lemilonkh: Will handle this on a separate branch

PR #1059, opened 2025-01-15
fix: on-2950, on-2830, on-3113 fix methane commitment formula issue

isaaccodekill:

Fix methane commitment formula issue by correcting logic in `ManualInputValidationService`, adjusting comparison logic in `FormattedNumberInput`, updating translations for methane-related phrases across multiple locales, and modifying default units in the form schema.

These changes are necessary to address issues with the validation of input fields where numerical types were incorrectly identified as missing, ensure value boundary conditions are properly checked, clarify translations to better represent the concept of fractional methane collection, and correct unit specification from tonnes to kilograms in the form schema to align with system requirements.


app/src/util/form-schema/manual-input-hierarchy.json
```diff
@@ -3811,7 +3811,7 @@
               "methane-commitment-solid-waste-inboundary-landfill-id"
             ],
             "activity-title": "total-municipal-solid-waste-disposed",
-            "default-units": "units-tonnes",
+            "default-units": "units-kilograms",
```
lemilonkh, line 3814: Was this change confirmed by the data team (that the formula expects the inputs in kg)?

lemilonkh (approved):
Great work, thanks 👍✨

app/src/util/form-schema/manual-input-hierarchy.json
isaaccodekill: yeah, this was gotten from the csv file that mau used to make hand calculations

PR #1071, opened 2025-01-20
fix: adds better token verification error handling

cephaschapa:
Changes
- adds error handling for password reset token this way we do get a user not found error when it expires

Improve token verification error handling in the password reset route by adding specific error checks for token expiration and invalid tokens, and providing more informative error messages.

The previous implementation did not differentiate between expired and invalid tokens, leading to generic error messages that could confuse users. By specifically identifying the type of token error, we can return more user-friendly messages and improve the overall user experience during password resets.


isaaccodekill (approved):
Great thanks!

PR #1072, opened 2025-01-20
Fix/on 2950 decimal entry foreign language

isaaccodekill:
- Fix the issue with entering decimal separators in Number Inputs  using non-english locale
- Localize emission values on all data tables

Refactor the code to support locale-specific decimal formatting and number parsing for foreign language entries within the `Activity` components by passing the appropriate locale to number formatting functions and updating the `formatted-number-input` component to use locale-specific decimal separators.

To address issue ON-2950, where decimal entries were not correctly handled in foreign language contexts, potentially leading to misinterpretation of numerical data. By incorporating locale-specific formatting and parsing, users can accurately enter and view numbers according to regional conventions, improving user experience and reducing errors. This solution leverages an accentuated focus on determining locale with `regionalLocales` constants and `useParams` for enhanced adaptability, though it might require further refinement for dynamic locale support in future iterations.

### DIFF
--- a/app/src/components/formatted-number-input.tsx
+++ b/app/src/components/formatted-number-input.tsx
@@ -8,6 +8,7 @@ import {
   NumberInputProps,
 } from "@chakra-ui/react";
 import { useParams } from "next/navigation";
+import { regionalLocales } from "@/util/constants";
 
 interface FormattedNumberInputProps extends NumberInputProps {
   control: Control<any, any>;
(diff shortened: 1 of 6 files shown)


app/src/util/constants.ts
```diff
@@ -136,3 +136,10 @@ export const getSectorByName = (name: string) =>
 
 export const getReferenceNumberByName = (name: keyof ISector) =>
   findBy("name", name)?.referenceNumber;
+
+export const regionalLocales: Record<string, string> = {
```
thehighestprimenumber, line 140: can we rename this to be in ALL_CAPS, since it's a constant?

app/src/components/formatted-number-input.tsx
```diff
@@ -48,22 +50,41 @@ function FormattedNumberInput({
     name,
   });
 
-  // Format the number according to the locale
   const format = (nval: number | string) => {
```
thehighestprimenumber, line 53: I think this function is worth having a unit test for

thehighestprimenumber (approved):

app/src/components/formatted-number-input.tsx
isaaccodekill: will do this in a follow up ticket

PR #1075, opened 2025-01-21
Nina/bug/on 3294/adapt forecast for sector V

thehighestprimenumber:
adjusting the emissions forecast to some changes we got from iCare (sector V comes separated into its subsectors)

Adapt the emissions forecast functionality to include subsector support by modifying the forecast data structure, adding subsector colors/configuration, and updating database queries.

The existing emissions forecast model didn't accommodate subsectors, especially for sector V, leading to inaccurate data representation. By enhancing the forecast logic with subsector handling, this change ensures more accurate emissions projections and visualization, addressing client feedback and aligning with updated business requirements. However, the approach introduces added complexity in mapping subsector identifiers, which could be further streamlined in future iterations.


app/src/app/[lng]/[inventory]/InventoryResultTab/EmissionsForecast/EmissionsForecastChart.tsx
```diff
@@ -38,29 +47,33 @@ export const EmissionsForecastChart = ({
   const convertToLineChartData = (
     forecastData: EmissionsForecastData,
   ): LineChartData[] => {
-    const sectors = Object.keys(
+    let sectors = Object.keys(
```
Comment (lemilonkh) [inline app/src/app/[lng]/[inventory]/InventoryResultTab/EmissionsForecast/EmissionsForecastChart.tsx:50]: Why the change to `let` here? It seems like it's not being changed afterwards...

app/src/app/[lng]/[inventory]/InventoryResultTab/EmissionsForecast/GrowthRatesExplanationModalTable.tsx
```diff
@@ -18,6 +18,12 @@ export const GrowthRatesExplanationModalTable = ({
   growthRates: ProjectionData;
   t: TFunction;
 }) => {
+  function getNameTranslationString(
+    sector: ISector | { name: string; color: string; referenceNumber: string },
+  ) {
+    return sector.name === "ippu" ? sector.name + "-short" : sector.name;
```
Comment (lemilonkh) [inline app/src/app/[lng]/[inventory]/InventoryResultTab/EmissionsForecast/GrowthRatesExplanationModalTable.tsx:24]: Should we change the default string for this sector to be the short one? So we don't have to make this special case just for this one sector...

app/src/app/[lng]/[inventory]/InventoryResultTab/EmissionsForecast/GrowthRatesExplanationModalTable.tsx
```diff
@@ -34,20 +40,25 @@ export const GrowthRatesExplanationModalTable = ({
...
-                  <Td key={year}>
-                    {growthRates[year][sector.referenceNumber]}
-                  </Td>
-                ))}
-              <Td>{growthRates["2030"][sector.referenceNumber]}</Td>
-              <Td>{growthRates["2050"][sector.referenceNumber]}</Td>
-            </Tr>
-          ))}
+          {[
+            ...SECTORS.slice(0, 4),
```
Comment (lemilonkh) [inline app/src/app/[lng]/[inventory]/InventoryResultTab/EmissionsForecast/GrowthRatesExplanationModalTable.tsx:44]: I think we store if the sectors are BASIC or BASIC+ on them now, so there might be better ways to filter them than just the index.

This doesn't necessarily apply to the SECTORS constant, but could be added here too for more flexibility.

Not explicitly necessary for this PR though.

lemilonkh (approved):
Great, thanks :sparkles:

app/src/app/[lng]/[inventory]/InventoryResultTab/EmissionsForecast/GrowthRatesExplanationModalTable.tsx
Comment (thehighestprimenumber) [reply app/src/app/[lng]/[inventory]/InventoryResultTab/EmissionsForecast/GrowthRatesExplanationModalTable.tsx:24]: we use the short name in only one place, while we use the long name everywhere else, I think making the change would require a lot of code changes and possible bugs

app/src/app/[lng]/[inventory]/InventoryResultTab/EmissionsForecast/GrowthRatesExplanationModalTable.tsx
Comment (thehighestprimenumber) [reply app/src/app/[lng]/[inventory]/InventoryResultTab/EmissionsForecast/GrowthRatesExplanationModalTable.tsx:44]: we're not separating basic from basic+, only treats sector V differently because it uses the subsectors

PR #1080, opened 2025-01-23
bug: [ON-3320] fix report results when inventory has no data

thehighestprimenumber:

Fix the handling of report results when the inventory has no data by ensuring that `totalEmissions` returns default values of an empty array for sectors and a "0" string for the total, and add test cases to verify this behavior.

This change addresses a bug where previous implementations failed to handle inventories without data gracefully, leading to potential errors or misleading responses. By providing default values, we ensure consistent and predictable API responses even when an inventory lacks data, thereby improving the robustness of the system. The addition of test cases ensures this behavior is validated and prevents future regression.


evanp (approved):
Looks good!

PR #1081, opened 2025-01-23
bug: [ON-3242] fix routing when invalid invite

thehighestprimenumber:

Fix the routing logic to navigate users to the home page (`/`) only when an invite is valid by adding a condition to check for errors before performing the routing.

The existing implementation incorrectly routed users to the home page even when an invite was invalid. This change ensures that users are only redirected to the home page when there is no error, maintaining the correct application flow and improving user experience. The commented-out code indicates the previously incorrect behavior was intentionally deactivated.


cephaschapa (approved):
good catch

PR #1096, opened 2025-02-03
Feat/on 2688 load testing

isaaccodekill:
Load test setup with k6

Add a new load testing script for the eCRF download endpoint using k6, including virtual users, duration, error thresholds, and checks for response status and content type.

These changes are being made to ensure the performance and reliability of the eCRF download endpoint under load conditions, simulating up to 50 concurrent users over a 10-second period. This approach is essential for verifying that the endpoint can handle expected traffic efficiently, identifying any performance bottlenecks, and ensuring that the response meets the defined criteria for success.


thehighestprimenumber (approved):
maybe remove the hardcoded inventoryId and collect all the lines into a single log call?

PR #1106, opened 2025-02-04
chore: move react-table types to regular dependencies

thehighestprimenumber:

Move `@types/react-table` from devDependencies to regular dependencies in `package.json` for consistency.

The `@types/react-table` package is required at runtime, not just during development or testing phases, and therefore should be a regular dependency. This ensures its availability in production environments where it is needed.


thehighestprimenumber:
not needed after all

PR #1121, opened 2025-02-12
Fix/settings page style changes

cephaschapa:
Changes
This PR fixes styling and some other breaking changes since the UI upgrate, also adds a skeleton componet to handle loading state

### DIFF
--- a/app/src/components/Modals/delete-city-modal.tsx
+++ b/app/src/components/Modals/delete-city-modal.tsx
@@ -71,7 +71,7 @@ const DeleteCityDialog: FC<DeleteCityDialogProps> = ({
 
   return (
     <>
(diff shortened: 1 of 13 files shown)


app/src/components/Modals/delete-city-modal.tsx
```diff
@@ -71,7 +71,7 @@ const DeleteCityDialog: FC<DeleteCityDialogProps> = ({
 
   return (
     <>
-      <DialogRoot preventScroll open={isOpen} onExitComplete={onClose}>
+      <DialogRoot preventScroll open={isOpen} onOpenChange={onClose}>
```
lemilonkh, line 74: Let's check if it's actually being checked by using the argument to determine the opening/ closing state (same for the other onOpenChange params below)

lemilonkh (approved):
Thanks :+1:

PR #1133, opened 2025-02-18
Fix/user details not showing

cephaschapa:
Changes
- Adds setValue method to set field user profile values

Update `AccountDetailsTabPanel` component to automatically populate user details (name, email, and role) into the form fields using the `setValue` method when `userInfo` is available.

These changes are intended to address an issue where user details were not being displayed within the form by ensuring that the form fields are correctly and automatically populated when editing profile details. This change leverages the `setValue` function provided by the form management library to enhance user experience by reducing manual input and potential errors.


lemilonkh (approved):
Let's make sure we only set this on the first run, not in every rendering pass afterwards.
Otherwise great :+1:

PR #1135, opened 2025-02-19
fix(ui): tabs on InventoryResult missing content 

lemilonkh:
Happened after Chakra v3 migration - tab trigger values weren't properly set, same with the defaultValue


cephaschapa (approved):
Great work 👍

PR #1136, opened 2025-02-19
fix: first scope tab not selected by default on subsector page

lemilonkh:
Also simplify default tab handling by not using full reference numbers as tab values

Refactor the subsector page to set the first scope tab as the default selection and update the `Tabs.Content` component positioning within the subsector page and activity tab components.

The changes address an issue where the first scope tab was not being selected by default on the subsector page by initializing `Tabs.Root` with a "0" default value. The refactor improves the tab navigation and ensures consistency by standardizing the use of index-based values for tab selection rather than relying on reference numbers, thus providing a more robust and maintainable structure for handling tabs across different components.


cephaschapa (approved):
🚀🚀🚀

PR #1138, opened 2025-02-20
fix(ui): don't redirect to invites page without a token in query params

lemilonkh:
Fixes 404 error after login when navigating directly to login page

Modify the login page to prevent redirection to the invites page if query parameters lack a token.

Previously, users could be redirected to the invites page even when query parameters didn't include a token, causing unnecessary navigational logic that could be confusing. This change ensures redirection only occurs when a token is present, improving navigation accuracy.


cephaschapa (approved):
Thanks 👍

PR #1146, opened 2025-02-21
fix(ui): redirect to /onboarding when inventory is missing

lemilonkh:
Fixes users getting a 404 page after first signup and then login, so they couldn't use the platform

Update routing logic in the NotFound and MissingInventory components to redirect users to the root or /onboarding path when the inventory is missing, and replace hardcoded support emails with an environment variable.

The prior logic erroneously attempted to push a nonexistent or missing inventory ID, which could lead to user errors. By redirecting users to the /onboarding path or root, the UX is streamlined, guiding users appropriately when inventory data is unavailable. Additionally, using an environment variable for support emails increases maintainability and prevents the need for code changes if email addresses are updated.

### DIFF
--- a/app/src/app/[...not_found]/page.tsx
+++ b/app/src/app/[...not_found]/page.tsx
@@ -1,7 +1,6 @@
 "use client";
 import { Button } from "@/components/ui/button";
 import { useTranslation } from "@/i18n/client";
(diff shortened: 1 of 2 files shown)


cephaschapa (approved):
Nice thanks 👍

PR #1152, opened 2025-02-24
Fix/email redirect url

cephaschapa:
Changes
- Adds logic to check if invited user exsists in the database
- Adds doesUserExist param in the invitation url
- Implements routing login to login front end auth page

Redirect users based on the existence of an account linked to the invitation in the login page and modified the invitation email to include a query parameter indicating if an invited user exists in the system.

The changes improve user flow by directing users who do not have an existing account to the signup page upon login attempt, while those who do are directed to the invites page. This adjustment ensures users are appropriately guided based on their account status, aligning with usability improvements. The addition of this logic in the email route helps establish the condition used in redirects.

### DIFF
--- a/app/src/app/[lng]/auth/login/page.tsx
+++ b/app/src/app/[lng]/auth/login/page.tsx
@@ -65,6 +65,22 @@ export default function Login({
     }
   }
 
+  // Extract doesInvitedUserExist from callback params\
+  // If it is true, redirect to /user/invites page
+  // If it is false, redirect to /auth/signup page
+  const [path, queryString] = callbackUrl.split("?");
+  const callbackUrlParams = new URLSearchParams(queryString);
+  const doesInvitedUserExist = callbackUrlParams.get("doesInvitedUserExist");
+
+  if (doesInvitedUserExist && doesInvitedUserExist !== "true") {
+    // remove the doesInvitedUserExist param from the callbackUrl
+    callbackUrlParams.delete("doesInvitedUserExist");
+    const updatedCallbackUrl = `${path}?${callbackUrlParams.toString()}`;
+    return router.push(
+      "/auth/signup/?callbackUrl=" + encodeURIComponent(updatedCallbackUrl),
+    );
+  }
+
   const { showLoginSuccessToast } = useAuthToast(t);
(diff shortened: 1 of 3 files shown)


app/src/app/[lng]/auth/login/page.tsx
```diff
@@ -65,6 +65,22 @@ export default function Login({
     }
   }
 
+  // Extract doesInvitedUserExist from callback params\
+  // If it is true, redirect to /user/invites page
+  // If it is false, redirect to /auth/signup page
+  const [path, queryString] = callbackUrl.split("?");
```
Comment (lemilonkh) [inline app/src/app/[lng]/auth/login/page.tsx:71]: I think we need to check here if there is a `?` in the string beforehand? Since people can navigate directly to the login page (we had errors about that previously).

app/src/app/api/v0/user/invites/route.ts
```diff
@@ -123,8 +123,13 @@ export const POST = apiHandler(async (req, { params, session }) => {
...
           }),
         );
+        // Check if invited user exists then flag that user if they don't exist in the database
+
+        const doesInvitedUserExist = await db.models.User.findOne({
+          where: { email },
+        });
         const host = process.env.HOST ?? "http://localhost:3000";
-        const url = `${host}/user/invites?cityIds=${encodeURIComponent(invites.map((i) => i.cityId).join(","))}&token=${encodeURIComponent(invitationCode)}&email=${encodeURIComponent(email)}`;
+        const url = `${host}/user/invites?cityIds=${encodeURIComponent(invites.map((i) => i.cityId).join(","))}&token=${encodeURIComponent(invitationCode)}&email=${encodeURIComponent(email)}&doesInvitedUserExist=${doesInvitedUserExist ? "true" : "false"}`;
```
lemilonkh, line 132: Is the new parameter used somewhere in the codebase? It seems it's only being removed on the login page, but the invite route doesn't check it yet?

lemilonkh, line 132: Might make sense to use URLSearchParams here too to construct the search string rather than doing it manually?

cephaschapa: Its only being used in this route

lemilonkh (approved):

PR #1153, opened 2025-02-25
Fix/remove redux persistance middleware

cephaschapa:
Changes
- Removes redux persist library
- Removes imports and provider components

Remove Redux Persist middleware and its related dependencies and configuration to simplify the Redux store setup.

Redux Persist was being used to support persistent state, which is no longer needed for the current application requirements and is adding unnecessary complexity. This change streamlines the store configuration and eliminates potential issues related to persistent state synchronization across sessions.

### DIFF
--- a/app/src/app/providers.tsx
+++ b/app/src/app/providers.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { persistor, store } from "@/lib/store";
+import { store } from "@/lib/store";
 // import { CacheProvider } from "@chakra-ui/next-js";
 // TODO enable again when dark mode is supported
 // import { Provider as ChakraProvider } from "@/components/ui/provider";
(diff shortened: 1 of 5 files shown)


app/src/app/providers.tsx
```diff
@@ -1,14 +1,13 @@
 "use client";
 
-import { persistor, store } from "@/lib/store";
+import { store } from "@/lib/store";
 // import { CacheProvider } from "@chakra-ui/next-js";
```
lemilonkh, line 4: Can we remove this comment? The line exists again below (uncommented)

lemilonkh (approved):
Thanks 👍✨

app/src/app/providers.tsx
cephaschapa: Sure

PR #1156, opened 2025-02-26
bug: [ON-3443] use our Toast hooks when possible, add <Toaster/> everywhere else

thehighestprimenumber:


app/src/components/Tabs/MyProfileTab/AccountDetailsTabPanel.tsx
```diff
@@ -41,12 +46,7 @@ const AccountDetailsTabPanel: FC<AccountDetailsFormProps> = ({
       name: data.name,
       email: data.email,
       role: data.role,
-    }).then(() =>
-      toaster.success({
-        description: t("user-details-updated"),
-        duration: 5000,
-      }),
-    );
+    }).then(() => showSuccessToast());
```
lemilonkh, line 49:
```suggestion
    }).then(showSuccessToast);
```

app/src/hooks/Toasts.tsx
```diff
@@ -1,36 +1,58 @@
...
   title: string;
-  description: string;
-  text: string;
+  description?: string;
+  duration?: number;
+  t?: TFunction;
 }) {
-  const { title, description, text } = props;
+  const currentLanguage = i18next.language;
+  const { t: tFeedback } = useTranslation(currentLanguage, "feedback");
```
lemilonkh, line 17: Hooks need to be used on the top level of React components usually (this function is called from inside of if statements in a bunch of places though). Let's make sure we're not causing errors because of the hook usages in places where these toasts are emitted and check the changed call sites for errors.

lemilonkh (commented):
Thanks 👍

lemilonkh (approved):

PR #1158, opened 2025-02-27
fix added the datasource for emissions factor in route

amanda-eames:
Add the `emissionfactor_datasource` field to the emissions factor data retrieval in the `db_query` method within the `ghgi_emissions.py` route.

This change is made to ensure that the data source of the emission factors is explicitly included in query results, enhancing data transparency and allowing consumers of the API to trace the origin of emission factor values. It aligns with the requirement for clearer data provenance in emissions reporting.


thehighestprimenumber (approved):

evanp:
@thehighestprimenumber I just merged this based on your approval. Is there a reason you didn't merge? I think the person who reviews and approves should do the merge. Am I missing something?

PR #1184, opened 2025-03-06
isaac/bug/on-3523/wrong-year-over-year-calculations

isaaccodekill:

Fix the calculation of year-over-year percentage change by correctly dividing by the previous year's data, and handle cases where last year's data equals zero.

Users reported incorrect year-over-year emission percentage changes due to division by the wrong denominator, which has been adjusted to use last year's data; additionally, a condition was added to gracefully handle cases where last year's emission data is zero to prevent division errors.


lemilonkh (approved):
Thanks :) Let's remove the logging, then this can be merged :+1:

PR #1185, opened 2025-03-06
fix(ui): ON-3502 adds positioning prop to popover component

cephaschapa:
Changes
- Adds positioning prop to popover component to fix overflowing


lemilonkh (approved):
Thanks :+1:

PR #1197, opened 2025-03-10
fix(e2e): log out before performing invalid data test in login.spec.ts

lemilonkh:

Ensure the user is logged out before performing the invalid data test in `login.spec.ts` by navigating to the signout endpoint prior to the test execution.

Logging out before the test ensures the test environment is in a consistent and known state, preventing potential order of execution issues that might arise if the user is already logged in. This change helps to maintain the reliability and accuracy of end-to-end tests by avoiding false positives or negatives due to session-related issues.

### DIFF
--- a/app/e2e/helpers.ts
+++ b/app/e2e/helpers.ts
@@ -27,17 +27,6 @@ export async function signup(
   return await result.json();
 }
 
-export async function logOut(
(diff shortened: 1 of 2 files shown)


cephaschapa (approved):
Nice 👍

PR #1207, opened 2025-03-11
feat(gtm):[on-3493] fixes client side crash in percentage-breakdown-i…

isaaccodekill:
…nput.tsx


lemilonkh (approved):
Great spot, thanks :+1:

PR #1228, opened 2025-03-19
 chore: [ON-2851]. Update model definitions to allow setting not null values in the interface

thehighestprimenumber:

Update model definitions to allow implementing partial attribute interfaces for flexible attribute initialization.

The change supports setting partial attributes for model classes, providing greater flexibility during object initialization by allowing omitted non-required attributes. This approach aligns with scenarios where only specific fields need to be instantiated, improving code maintainability and adhering to TypeScript's strict type checking capabilities.

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -28098,6 +28098,21 @@
         "type": "github",
         "url": "https://github.com/sponsors/wooorm"
       }
+    },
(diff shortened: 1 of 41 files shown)


lemilonkh (approved):
Great, thanks :+1: :sparkles:

PR #1231, opened 2025-03-20
feat: update Spanish translations

evanp:
These are automatically generated translations of strings into Spanish. I'm not going to fix the wording of individual translations; I just need to know that these files work correctly. If you think one or more of the translation strings needs to be changed, you can fix it yourself and make a PR!


lemilonkh (approved):
Great, thanks :+1:

PR #1236, opened 2025-03-21
fix: adds condition to render admin menu item

cephaschapa:
Changes
- Adds condition to render admin menu item

Add a condition to render the admin menu item based on the user's role in the navigation bar component.

Without this condition, the admin menu item is rendered for all users indiscriminately. This change ensures that only users with an admin role will see the admin menu item, enhancing both security and user experience by preventing unauthorized access to admin features.


lemilonkh (approved):
Great, thanks :+1: :sparkles:

PR #1247, opened 2025-03-25
fix(api): make notation key manager GET API use scopes

lemilonkh:
Return subsector + scope list instead of just subsectors, ignoring the scopes.


cephaschapa (approved):
Thanks 👍

PR #1248, opened 2025-03-25
feat(ui): show error toast when saving notation keys fails

lemilonkh:

Add an error toast notification when saving notation keys fails in the `SectorTabs` component.

This change provides immediate user feedback when an error occurs during the save operation, enhancing the user experience by clearly indicating when the action was unsuccessful. Prior to this, the error was only logged to the console without a visual indication to the user.


cephaschapa (approved):
👍

PR #1249, opened 2025-03-26
UI: on_3608 side bar cc dashboard

isaaccodekill:
Uploading cc_sidebar.mov…

Integrate project-based organization and access management capabilities, update user interface elements and add new models for role management in projects and organizations.

These changes address the need for enhanced project navigation and access control by connecting user roles with project and organization affiliations. The update improves the application’s scalability for multi-organization use and enhances the user interface for better project navigation and selection. The underlying changes include incorporating `zustand` for state management, ensuring that navigation and state management across components are efficient and scalable.


app/src/stores/project-drawer-store.ts
```diff
@@ -0,0 +1,14 @@
+import { create } from "zustand";
```
lemilonkh: I personally like `zustand`, however since we are already using Redux Toolkit for state management having both as dependencies might be a bit overkill?

isaaccodekill: This has been removed in newer versions of the pr

PR #1280, opened 2025-04-09
chore: created API endpoint for gettig climate actions

mircorudolph:
Added an API endpoint to the global-api to get the list of climate actions.

Endpoint /api/v0/climate_actions

### DIFF
--- a/global-api/routes/get_climate_actions.py
+++ b/global-api/routes/get_climate_actions.py
@@ -1,0 +1,103 @@
+from fastapi import APIRouter, HTTPException
+from sqlalchemy import text
+from db.database import SessionLocal
+from typing import List, Dict, Any
+
+api_router = APIRouter(prefix="/api/v0")
+
+
+def db_climate_actions() -> List[Dict[str, Any]]:
+    with SessionLocal() as session:
+        query = text(
+            """
+            SELECT 
+                action_id as "ActionID",
+                action_name as "ActionName",
+                action_type as "ActionType",
+                hazard as "Hazard",
+                sector as "Sector",
+                subsector as "Subsector",
+                primary_purpose as "PrimaryPurpose",
+                description as "Description",
+                -- CoBenefits columns
+                co_benefits_air_quality as "CoBenefits_air_quality",
+                co_benefits_water_quality as "CoBenefits_water_quality",
+                co_benefits_habitat as "CoBenefits_habitat",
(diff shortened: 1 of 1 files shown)


global-api/routes/get_climate_actions.py
```diff
@@ -0,0 +1,125 @@
+from fastapi import APIRouter, HTTPException
+from sqlalchemy import text
+from db.database import SessionLocal
+from typing import List, Dict, Any
+
+api_router = APIRouter(prefix="/api/v0")
```
evanp, line 6: Are we using v0 or v1 now?

global-api/routes/get_climate_actions.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def db_climate_actions() -> List[Dict[str, Any]]:
+    with SessionLocal() as session:
+        query = text(
+            """
+            SELECT 
+                ActionID,
```
evanp, line 14: is this better than select * somehow?

global-api/routes/get_climate_actions.py
```diff
@@ -0,0 +1,125 @@
...
+        result = session.execute(query).mappings().all()
+
+        # Process the results to construct dictionaries
+        actions = []
+        for row in result:
+            action_dict = dict(row)
+
+            # Construct CoBenefits dictionary
+            co_benefits = {
+                "air_quality": action_dict.pop("CoBenefitsAirQuality"),
```
evanp, line 63: Why are you popping these?

evanp: You could use get() instead of pop()?

global-api/routes/get_climate_actions.py
```diff
@@ -0,0 +1,125 @@
...
+                adaptation_effectiveness_per_hazard
+            )
+
+            actions.append(action_dict)
+
+        return actions
+
+
+@api_router.get(
+    "/climate_actions", summary="Get all climate actions as a list of dictionaries"
```
evanp, line 112: Should this just be /actions ? Do we need the climate_ prefix?

evanp (approved):
Looks good!

global-api/routes/get_climate_actions.py
mircorudolph: we want to get rid of the otherwise double entries

PR #1283, opened 2025-04-09
fix: sql query

mircorudolph:

Revise the SQL query in `get_climate_actions.py` by adding double quotes around column names.

The change ensures compatibility with PostgreSQL where case sensitivity issues can occur without quoted column names. This approach preserves the intended lowercase column names and avoids future syntax errors, facilitating seamless interaction with the database schema.


evanp (approved):
Nicely done.

PR #1289, opened 2025-04-14
Fix: ON-3576 emission factor type not being validated

cephaschapa:
Changes
- Adds invalid prop to trigger input errors messages
- Adds register function with 'required' config
- Correctly sets the onchange function to the right component
- cosmetic changes


lemilonkh (approved):
Great, thanks :+1:

PR #1305, opened 2025-04-22
Adds conversion rules for new fuel Units

isaaccodekill:
Add conversion rules for new fuel units in the `UnitConversionService` and use `Decimal.js` for precise calculation of inventory percentage in `InventoryResultTab`.

These changes enhance the system by supporting new unit conversion capabilities for different fuel measurements, which is necessary due to expanded fuel unit options. The adjustment to `Decimal.js` ensures more accurate calculations, addressing precision issues seen with floating-point arithmetic in JavaScript.


cephaschapa (approved):
👍👍

PR #1311, opened 2025-04-24
bug (project overview): [ON-3823] hide Project overview link behind FF

thehighestprimenumber:
also, store feature flags in enum.

Hide the "Project overview" link feature behind the `PROJECT_OVERVIEW_ENABLED` feature flag and refactor feature flag usage to utilize an enumerated type for better consistency.

This change improves feature management by allowing the "Project overview" link to be conditionally displayed based on configuration while ensuring that the feature flag handling is more structured and less error-prone through the use of TypeScript enums, which reduces the risk of typos and enhances code readability.


evanp (changes requested):
So, for a hot patch fix to main, I'd really prefer that we don't change how things like feature flags work. As we can see in this patch, it means making changes to unrelated code. For hot patches, we really want to keep things minimal.

Could you just make the change to check hasFeatureFlag for project overview?

PR #1319, opened 2025-04-28
fix: deleted duplicate EF

amanda-eames:


lemilonkh (approved):
To make sure these are removed from the database we need to clear the EmissionsFactor table in the DB before merging this (or before running the import again). Or alternatively adjust the seeder script so it removes existing values that are no longer in the file.

amanda-eames:
I already did a delete of these emissions factors in the DB, was causing an issue with manual import.

PR #1324, opened 2025-04-29
adds condition to render manage subsectors sector button

cephaschapa:
Changes
- Adds condition using isPublic variable to hide manage subsectors button on public dashboard

Add a condition to render the "manage subsectors" button only when `isPublic` is `false`.

The button should be visible only to certain users, presumably those with permission to manage subsectors, ensuring it doesn't appear in public views. This change enhances security and user experience by displaying relevant options based on user access levels.


lemilonkh (approved):
Good spot 👍

PR #1330, opened 2025-05-05
feat: [on-3694] implements white label color-system

isaaccodekill:
Implement a white label color system enabling customizable color themes across the application while updating the design components to use the new color variables.

These changes are made to provide a flexible theming framework that supports different color themes for white labeling purposes, allowing the app to adapt to different brand requirements efficiently. This approach centralizes color definitions for the application, simplifying future updates and maintenance.

image

image

image


app/src/app/[lng]/account-settings/account/account-settings-tab.tsx
```diff
@@ -0,0 +1,233 @@
...
+            letterSpacing="wide"
+            marginTop="8px"
+          >
+            {t("brand-settings-description")}
+          </Text>
+          <Box mt={9}>
+            <Field className="w-full" label={t("logo")}>
+              <FileUploadRoot accept={{ "image/*": [] }} maxFiles={1}>
+                <LogoUploadCard />
+              </FileUploadRoot>
```
Comment (lemilonkh) [inline app/src/app/[lng]/account-settings/account/account-settings-tab.tsx:142]: It looks like the FileUploadRoot could be moved into LogoUploadCard perhaps?

app/src/lib/theme/recipes/app-theme.ts
```diff
@@ -0,0 +1,656 @@
...
+              _light_brown_theme:
+                "{colors.light_brown_theme.content.alternative}",
+              _dark_orange_theme:
+                "{colors.dark_orange_theme.content.alternative}",
+              _green_theme: "{colors.green_theme.content.alternative}",
+              _light_blue_theme:
+                "{colors.light_blue_theme.content.alternative}",
+              _violet_theme: "{colors.violet_theme.content.alternative}",
+            },
+          },
```
lemilonkh, line 51: These are very verbose and need to be created for every color name and theme we use it seems. I wonder if they could be created by a function beforehand that uses map statements or for loops to generate these objects for each color name and theme in advance, before passing the result into `createSystem`...

lemilonkh (approved):
Great work!

I would like to see the creation of new color names and color themes be simplified a bit if possible.

app/src/app/[lng]/account-settings/account/account-settings-tab.tsx
Comment (isaaccodekill) [reply app/src/app/[lng]/account-settings/account/account-settings-tab.tsx:142]: It needs to be outside inorder to use the context api within

app/src/lib/theme/recipes/app-theme.ts
isaaccodekill: That's a great idea we can explore this in a follow up implementation

PR #1370, opened 2025-05-20
fix(ui):ON-3866 adds logic to render default project and adds translations

cephaschapa:
Changes
- Adds logic to render default project instead of `cc_project_default` on dashboard and drawer
- Adds translations for other languages

Add logic to render a default project name in the UI and update translations to include a new key for "default-project" in multiple languages.

These changes address the need to properly display a default project name in the UI when a specific condition is met and to ensure translation consistency across supported languages so that users experience a localized interface. This approach aids in improving user experience by providing clarity and ensuring that the application UI accommodates multiple languages effectively.


lemilonkh (approved):
Great, thanks 👍

PR #1404, opened 2025-05-27
fix: capitalises data quality values on activity accordion

cephaschapa:
Changes
- Adds `textTransformation ` to capitalize data quality values on on the activity accordion component


lemilonkh (approved):
👍

PR #1418, opened 2025-05-30
console functions -> logger functions

evanp:
This is a pretty huge logging change for the service. It takes all the random console.log(), console.error(), and console.whatever() calls and uses the Pino logger for them instead. There is definitely additional work to be done, but this first pass should get us to timestamped, structured log entries.

I was able to build this branch and run it through most of the interface. Unfortunately, because of the nature of this change, a lot of the code is hard to test -- it's in error handler branches that are rarely triggered.

Refactor all instances of `console` logging statements to use a centralized `logger` service for more structured and maintainable logging.

The changes are aimed at enhancing the logging strategy by replacing unstructured `console` logs with a robust `logger` service. This will improve log management across different environments (e.g., development, production) and provide better context and structure to log outputs for easier debugging and monitoring.


app/src/app/[lng]/[inventory]/CapTab/ActionDrawer.tsx
```diff
@@ -41,7 +42,7 @@ export function ActionDrawer({
   onClose: () => void;
   t: TFunction;
 }) {
-  console.log(action)
+  logger.info(action)
```
Comment (lemilonkh) [inline app/src/app/[lng]/[inventory]/CapTab/ActionDrawer.tsx:45]: Should be removed since it's at the top level of the component so it will print on every render.

app/src/app/api/v0/inventory/[inventory]/download/route.ts
```diff
@@ -223,12 +223,12 @@ async function inventoryXLS(inventory: Inventory): Promise<Buffer> {
...
   }
-  console.timeEnd("edit_ciris");
+  logger.debug({ timeEnd: Date.now() }, "edit_ciris");
 
   // await workbook.xlsx.writeFile("test.xlsx");
-  console.time("save_ciris");
+  logger.debug({ time: Date.now() }, "save_ciris");
   const buffer = Buffer.from(await workbook.xlsx.writeBuffer());
-  console.timeEnd("save_ciris");
+  logger.debug({ timeEnd: Date.now() }, "save_ciris");
```
Comment (lemilonkh) [inline app/src/app/api/v0/inventory/[inventory]/download/route.ts:231]: Might be better to log the duration rather than the two timestamps? (which is the behaviour of the console.time function that was used here originally).

app/src/util/api.ts
```diff
@@ -69,7 +69,7 @@ export function apiHandler(handler: NextHandler) {
 
 function errorHandler(err: unknown, _req: NextRequest) {
   // TODO log structured request info like route here
-  console.error(err);
+  logger.error(err);
```
lemilonkh, line 72: The TODO comment can perhaps be removed as the logger is automatically logging more request context?

lemilonkh (approved):
Let's double-check that the logger works properly in the frontend and the log messages there aren't too confusing to read because of the `pino` formatting, then we can merge this.

app/src/app/api/v0/inventory/[inventory]/download/route.ts
Comment (lemilonkh) [reply app/src/app/api/v0/inventory/[inventory]/download/route.ts:231]: Same for the other instance of this in this file

PR #1419, opened 2025-05-30
fix(ui): ON-4024 prevent null string from making it into API calls

lemilonkh:
Use default instead of `"null"` in inventory-related API calls on HomePage, NavigationBar, root inventory layout

Update the UI logic to prevent the string "null" from being used as a valid inventory identifier in API calls.

APIs were receiving "null" as a string from certain components, leading to incorrect behavior or errors; this fix ensures only valid inventory IDs or default values are used, improving the robustness and predictability of API interactions.

### DIFF
--- a/app/src/app/[lng]/[inventory]/layout.tsx
+++ b/app/src/app/[lng]/[inventory]/layout.tsx
@@ -24,10 +24,15 @@ export default function DataLayout({
   const { data: userInfo, isLoading: isUserInfoLoading } =
     api.useGetUserInfoQuery();
 
-  const inventoryId = userInfo?.defaultInventoryId || inventory;
+  let inventoryId: string | null;
+  if (inventory && inventory !== "null") {
+    inventoryId = inventory;
+  } else {
+    inventoryId = userInfo?.defaultInventoryId ?? null;
+  }
 
   const { data: inventoryOrgData, isLoading: isInventoryOrgDataLoading } =
-    useGetOrganizationForInventoryQuery(inventoryId, {
+    useGetOrganizationForInventoryQuery(inventoryId!, {
       skip: !inventoryId,
     });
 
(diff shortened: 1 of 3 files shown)


app/src/app/[lng]/[inventory]/layout.tsx
```diff
@@ -24,10 +24,15 @@ export default function DataLayout({
...
+  let inventoryId: string | null;
+  if (inventory && inventory !== "null") {
+    inventoryId = inventory;
+  } else {
+    inventoryId = userInfo?.defaultInventoryId ?? null;
+  }
 
   const { data: inventoryOrgData, isLoading: isInventoryOrgDataLoading } =
-    useGetOrganizationForInventoryQuery(inventoryId, {
+    useGetOrganizationForInventoryQuery(inventoryId!, {
```
Comment (thehighestprimenumber) [inline app/src/app/[lng]/[inventory]/layout.tsx:35]: by the logic on line 31, inventoryId can be null :confused:

app/src/components/HomePage/HomePage.tsx
```diff
@@ -115,15 +123,13 @@ export default function HomePage({
   }, [cityYears]);
 
   const { data: inventoryOrgData, isLoading: isInventoryOrgDataLoading } =
-    useGetOrganizationForInventoryQuery(inventoryIdFromParam as string, {
+    useGetOrganizationForInventoryQuery(inventoryIdFromParam!, {
```
thehighestprimenumber, line 126: why do we need the `!` here?

thehighestprimenumber (approved):

cephaschapa (approved):
👍👍👍

app/src/app/[lng]/[inventory]/layout.tsx
Comment (lemilonkh) [reply app/src/app/[lng]/[inventory]/layout.tsx:35]: Yes, this gives a type error without the `!`. It will never be passed into the API as when it's null the skip condition will be active.

app/src/components/HomePage/HomePage.tsx
lemilonkh: Yes, this gives a type error otherwise. It will never be passed into the API as when it's null the skip condition will be active.

lemilonkh:
Closing in favor of #1420 so it can be merged into dev and tested first.

PR #1420, opened 2025-05-30
fix(ui): ON-4024 prevent null string from making it into API calls

lemilonkh:
Use default instead of "null" in inventory-related API calls on HomePage, NavigationBar, root inventory layout


lemilonkh:
@thehighestprimenumber this will be resolved in a follow up PR. For now this prevents the invalid `"null"` string from making it into the API requests.

evanp:
@thehighestprimenumber does this still happen on develop?

isaaccodekill (approved):
Amazing work, thanks!

PR #1456, opened 2025-06-06
fix(db): ON-3955 remove deprecated emissions factors

lemilonkh:
- Adds `deprecated` column to `EmissionsFactor` (default false)
- Sets it to true for all existing entries at the start of the EF seeder
- Newly written/ updated entries will have it set to false
- Query at the end of all seed folder imports removed EF that are still deprecated and not used in any inventories

Add a `deprecated` column to the `EmissionsFactor` model and update seeders to set existing emissions factors as deprecated, subsequently removing unused deprecated factors.

These changes manage emissions factors more effectively by deprecating factors no longer in use, maintaining necessary factors for historical data consistency, and cleaning the database of unused, deprecated emissions factors. The approach ensures that outdated emissions factors do not interfere with current calculations, yet preserves them for reference in existing inventories.


amanda-eames (approved):
Looks good

PR #1513, opened 2025-06-20
fix: db constaints for city polygon

amanda-eames:

Update the database migration script to change `country_code`, `region_code`, and `city_id` columns to `VARCHAR` type, and refactor the city boundary endpoint logic to use direct SQL queries for more precise location data retrieval, eliminating complex geometry transformations.

The migration to `VARCHAR` ensures consistent data type across relevant fields, which offers better interoperability with other systems. Refactoring the endpoint logic enhances performance by utilizing SQL's native spatial functions, which are optimized for geographical data operations, and simplifies previous complex calculations within the application code. These updates streamline operations, reduce error potential, and optimize code maintainability.


global-api/migrations/versions/3fcc7f5218ed_merge_formular_values_and_country_.py
```diff
@@ -0,0 +1,26 @@
...
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    pass
+
+
+def downgrade() -> None:
+    pass
```
lemilonkh, line 26: This seems to be an empty migration?

global-api/migrations/versions/d84e47e54741_change_country_region_city_columns_to_.py
```diff
@@ -0,0 +1,45 @@
...
+revision: str = 'd84e47e54741'
+down_revision: Union[str, None] = 'e3c866a57c19'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # Create city_polygon table
+    op.create_table('city_polygon',
+        sa.Column('city_id', sa.VARCHAR(), nullable=True),
```
lemilonkh, line 24: Should be or type UUID? Or is this a locode?

global-api/routes/city_boundaries_endpoint.py
```diff
@@ -1,109 +1,53 @@
...
-        Query the database to retrieve OSM data for a given locode.
+        Query the database to retrieve city polygon data for a given locode.
 
         Args:
             locode (str): The location code to filter by.
 
         Returns:
-            Osm: The queried row from the OSM table matching the locode.
+            dict: The queried row from the modelled.city_polygon table matching the locode.
+            None: If no city is found with the given locode.
```
lemilonkh, line 18: Since the output schema was apparently changed, do we need to create a new version of this endpoint for compatibility?

global-api/routes/city_boundaries_endpoint.py
```diff
@@ -171,30 +111,23 @@ def get_locode(lat: Decimal, lon: Decimal):
...
-    # Extract the locodes from the Osm objects
-
-    locodes = list(map(lambda x: x.locode, matches))
+        query = text("""
+            SELECT locode
+            FROM modelled.city_polygon
+            WHERE bbox_north >= :lat
+                AND bbox_south <= :lat
+                AND bbox_east >= :lon
+                AND bbox_west <= :lon
```
lemilonkh, line 120: I think this might break when going across certain meridians/ the equator?

Is there a native data type from the PostGIS to handle coordinate bounding boxes or coordinate rectangles?

lemilonkh (approved):
Great, thanks 👍🏼✨

global-api/migrations/versions/3fcc7f5218ed_merge_formular_values_and_country_.py
amanda-eames: This is just to resolve conflicting heads issue with the migration.

global-api/routes/city_boundaries_endpoint.py
amanda-eames: I have remove the lat lon conditions and we are checking the point is in the polygon, this shouldn't have any issues around meridians.

PR #1555, opened 2025-07-02
fix(ui): ON-4174 remove wrong alias for EmissionsFactor table in ranking query

lemilonkh:

Remove the incorrect alias for the `EmissionsFactor` table in the ranking query within `route.ts`.

The wrong alias was causing an issue by not correctly matching column names during query execution, leading to errors in data retrieval. This change corrects the alias usage to prevent query mismatches and ensure the desired data is accurately ranked and retrieved.


isaaccodekill (approved):
Great 🙌🏽

PR #1565, opened 2025-07-07
fix: on-4139-hydration-issue

isaaccodekill:

Add `suppressHydrationWarning` to HTML elements in layout files and re-enable a previously skipped test suite for the DataSource API.

To resolve hydration issues during server rendering by suppressing related warnings, ensuring consistent user experience across different locales in the application. Re-enabling the DataSource API test suite helps in maintaining code quality by ensuring that all functionalities continue to work as expected post-changes.

### DIFF
--- a/app/src/app/[...not_found]/layout.tsx
+++ b/app/src/app/[...not_found]/layout.tsx
@@ -21,7 +21,7 @@ export default function RootLayout({
   children: React.ReactNode;
 }) {
   return (
-    <html lang="">
+    <html lang="" suppressHydrationWarning>
       <head>
         <link rel="icon" type="image/svg+xml" href="/assets/icon.svg" />
         <link rel="icon" type="image/png" href="/assets/icon.png" />
--- a/app/src/app/[lng]/layout.tsx
+++ b/app/src/app/[lng]/layout.tsx
@@ -24,7 +24,7 @@ export default function RootLayout(props: {
   const { lng } = use(props.params);
 
   return (
-    <html lang={lng} dir={dir(lng)}>
+    <html lang={lng} dir={dir(lng)} suppressHydrationWarning>
       <head>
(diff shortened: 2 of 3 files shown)


app/src/app/[...not_found]/layout.tsx
```diff
@@ -21,7 +21,7 @@ export default function RootLayout({
   children: React.ReactNode;
 }) {
   return (
-    <html lang="">
+    <html lang="" suppressHydrationWarning>
```
Comment (lemilonkh) [inline app/src/app/[...not_found]/layout.tsx:24]: I'm not sure if this is the best way to handle it. Ideally we fix the root cause, which is that the data-theme property only renders on the client, not the server (which causes performance issues on the client when rendering as it needs to re-render everything the server sent).

app/src/app/[lng]/layout.tsx
```diff
@@ -24,7 +24,7 @@ export default function RootLayout(props: {
   const { lng } = use(props.params);
 
   return (
-    <html lang={lng} dir={dir(lng)}>
+    <html lang={lng} dir={dir(lng)} suppressHydrationWarning>
```
Comment (lemilonkh) [inline app/src/app/[lng]/layout.tsx:27]: Let's fix the hydration issue by making sure the server renders the same markup as the client or we use a suspense (but ideally not at the root of the page as it would prevent full server rendering)

app/tests/api/datasource.jest.ts
```diff
@@ -81,7 +81,7 @@ async function cleanupDatabase() {
   await db.models.Sector.destroy({ where: { sectorName } });
 }
 
-describe.skip("DataSource API", () => {
+describe("DataSource API", () => {
```
lemilonkh, line 84: How were these tests fixed? They were not working before, and this is part of a different ticket I think?

isaaccodekill: @evanp  had previously disabled them

PR #1571, opened 2025-07-10
fix: on-4242 wrong project list on invite modal

isaaccodekill:

Remove the `isAdmin` property and refactor project fetching logic in the AddCollaboratorsModal to show appropriate project lists based on organization ID or user role.

The invite modal was incorrectly listing projects due to the reliance on the `isAdmin` flag, which led to project visibility issues. By leveraging organization IDs and ensuring proper access checks via `UserService`, these changes ensure that the project list is tailored both to organizational context and user roles, improving accuracy and administrative control.

### DIFF
--- a/app/src/components/HomePage/AddCollaboratorModal/AddCollaboratorsModal.tsx
+++ b/app/src/components/HomePage/AddCollaboratorModal/AddCollaboratorsModal.tsx
@@ -54,14 +54,12 @@ const AddCollaboratorsDialog = ({
   onClose,
   onOpen,
   organizationId,
-  isAdmin,
 }: {
   lng: string;
   isOpen: boolean;
   onClose: () => void;
   onOpen?: () => void;
   organizationId?: string;
-  isAdmin?: boolean;
 }) => {
   const { t } = useTranslation(lng, "dashboard");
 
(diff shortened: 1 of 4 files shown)


app/src/components/HomePage/AddCollaboratorModal/AddCollaboratorsModal.tsx
```diff
@@ -220,7 +233,7 @@ const AddCollaboratorsDialog = ({
                 }}
               >
                 <option value="collaborator">{t("collaborator")}</option>
-                {isAdmin && <option value="admin">{t("admin")}</option>}
+                {organizationId && <option value="admin">{t("admin")}</option>}
```
lemilonkh, line 236: Is the organization ID only set if the user is an admin? This seems off.

app/src/components/HomePage/AddCollaboratorModal/AddCollaboratorsModal.tsx
```diff
@@ -301,7 +314,7 @@ const AddCollaboratorsDialog = ({
               <BodyLarge>{t("select-cities-to-share-description")}</BodyLarge>
               {cityData?.length === 0 && (
                 <Text fontSize="body.md" mt={3} color="content.tertiary">
-                  {isAdmin && !selectedProject
+                  {organizationId && !selectedProject
```
lemilonkh, line 317: Same question as before, why use the organization ID here instead of is admin?

lemilonkh (approved):
Great, thanks 👍🏼✨

app/src/components/HomePage/AddCollaboratorModal/AddCollaboratorsModal.tsx
isaaccodekill: no, the organizationID is set only on pages that have an organizational context.

PR #1579, opened 2025-07-14
fix(ui): ON-4170 make all go-back links more specific

lemilonkh:
To language and inventory ID (where possible)

Update all go-back links to use language-specific routes by adding language segments to URLs for consistency.

The previous implementation lacked specificity in navigation, which could result in incorrect page routing, especially for multilingual support. Including the language in the URL ensures that the routing logic aligns with the user's language context, which enhances the user experience by always taking the user back to the correct localized page.


thehighestprimenumber (approved):
nice!

PR #1604, opened 2025-07-22
fix: [on-4231] restrict add new inventory button to city limit count

isaaccodekill:


app/src/app/api/v0/city/route.ts
```diff
@@ -49,7 +49,8 @@ export const POST = apiHandler(async (req, { session }) => {
     throw new createHttpError.Forbidden("Organization is not active");
   }
 
-  if (project.cities.length === project.cityCountLimit) {
+  if (Number(project.cities.length) >= Number(project.cityCountLimit)) {
+    logger.error("City count limit reached for project");
```
cephaschapa, line 53: Would it also make sense to show a toast to alert the user of this limit message?

cephaschapa (approved):
Looks great 👍

app/src/app/api/v0/city/route.ts
isaaccodekill: Yeah we do, as this error is passed to the frontend

PR #1633, opened 2025-07-30
fix [ON-4430]: fix report results breakdown

thehighestprimenumber:
use inventory value instead of activities


lemilonkh (approved):
Great spot, thanks 👍🏼✨

thehighestprimenumber:
closing since the issue was in the database not in the code

PR #1638, opened 2025-07-31
fix: ON-4406 restores back button on dashboard

cephaschapa:
Changes
- Adds text color to back button
- Adds container size to page wrapper


lemilonkh (approved):
Thanks 👍🏼

PR #1647, opened 2025-08-06
Feat/on 4307 sidebar navigation

cephaschapa:
Changes
- Adds Journey navigator Navigation Drawer (Only shows on JN Page)
- Adds project and city filtering via search, and dropdowns
- Adds some nav links
- Adds dynamic module navigation with accordions

> [!NOTE]
> I'm currently writing a description for your pull request. I should be done shortly (<1 minute). Please don't edit the description field until I'm finished, or we may overwrite each other. If I find nothing to write about, I'll delete this message.


thehighestprimenumber:
Thank you for your work Cephas! Let's make sure we're keeping module logic separate from JN logic, and that we reuse existing code without duplicating it :pray:

PR #1648, opened 2025-08-06
fix/ON-4299-button-icon-overflowing

cephaschapa:
changes
- Removes margins and uses padding instead, this fixes the overflowing


app/src/components/GHGIHomePage/DownloadAndShareModals/PublishButtons.tsx
```diff
@@ -43,7 +43,8 @@ const PublishButtons = ({
         <Button
           key={title}
           my="16px"
-          mx="16px"
+          pr="16px"
+          pl="35px"
```
lemilonkh, line 47: Why does the padding need to be different on the left and right side of the button?

lemilonkh (approved):
Thanks 👍🏼

PR #1730, opened 2025-08-20
Feat/on 4418 fe re run button and section

cephaschapa:
Changes
- Adds citycatalyst actions section
- Adds re-run button
- Adds translations for all lngs
image


app/src/app/[lng]/cities/[cityId]/HIAP/page.tsx
```diff
@@ -51,8 +54,81 @@ export default function HIAPPage(props: {
...
+          <Box
+            display="flex"
+            mx="auto"
+            py="56px"
+            w="full"
+            maxW="1090px"
+            flexDirection="column"
+            gap="24px"
+          >
+            {/* citycatalyst actions section */}
```
Comment (lemilonkh) [inline app/src/app/[lng]/cities/[cityId]/HIAP/page.tsx:66]: This could be a separate component instead of a section with a comment in a larger component

lemilonkh (approved):
Thanks, looks good but let's refactor into separate components for sections where appropriate and use the components in src/components/Texts instead of manual Text styling where possible for consistency.

PR #1787, opened 2025-08-26
fix: e2e [4991] third party data sources e2e

isaaccodekill:


lemilonkh (approved):
Amazing work, thanks 👍🏼✨
I would like to see more isolation between the different modules in the future, let's keep that in mind for follow up PRs.

PR #1806, opened 2025-08-29
fix(api, ui): 0% of sector emissions on report sector table

lemilonkh:
Also fixes that some of the rows in the table weren't full width by using React.Fragment instead of <> or <Box>, so the key attribute can be supplied to it but the rows are direct children of the table body.

Replace `Box` components with `React.Fragment` in `ByScopeView.tsx` and refactor emission calculations to improve accuracy and replace `bigIntToDecimal` conversions with `Decimal` in `ResultsService.ts` for calculating subsector emissions percentage accurately in reports.

The `React.Fragment` usage instead of `Box` is intended to eliminate redundant DOM nodes and improve rendering performance. The refactoring in `ResultsService.ts` is aimed at fixing issues where emission percentages were inaccurately calculated due to improper usage of `bigIntToDecimal`, allowing for more precise financial calculations using `Decimal`. This approach is favored to ensure data integrity and consistency across reports, presenting the right emissions data for decision-making.

### DIFF
--- a/app/src/app/[lng]/[inventory]/InventoryResultTab/ByScopeView.tsx
+++ b/app/src/app/[lng]/[inventory]/InventoryResultTab/ByScopeView.tsx
@@ -95,19 +95,23 @@ const ByScopeView: React.FC<ByScopeViewProps> = ({
         </Table.Cell>
       ))}
       <Table.Cell>
-        <BodyMedium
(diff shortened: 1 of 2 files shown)


cephaschapa (approved):
Great, thanks 👍👍👍

PR #1813, opened 2025-09-03
feat: (JN) [on-4571] enables hiap in global dashboard

isaaccodekill:


thehighestprimenumber (approved):
Nice work! Let's make sure to check permissions to module in the routes, and to avoid duplicated code in the frontend

PR #1830, opened 2025-09-10
feat: adds total emissions component and translations

cephaschapa:
Changes
- Adds total emissions component on JN dashboard
- Adds translations

image


app/src/components/HomePageJN/Hero.tsx
```diff
@@ -48,6 +55,10 @@ export function Hero({
     skip: !city?.locode,
   });
 
+  const formattedEmissions = ghgiCityData?.totalEmissions
+    ? formatEmissions(ghgiCityData.totalEmissions)
+    : { value: t("N/A"), unit: "" };
```
lemilonkh, line 60: I think we are using the translation key `n-a` elsewhere, might make sense to use it here as well.

lemilonkh (approved):
Great, thanks 👍🏼

PR #1831, opened 2025-09-10
chore: remove unused importers and datasets from repo

evanp:
In reviewing the repository today to add the climate-advisor microservice, I realized we still had a tonne of importer and seed data directories in the repo. This PR is to remove that unused code. Please take a look if we still need any of this!


lemilonkh (approved):
Great, thanks 👍🏼

evanp:
@amanda-eames can you confirm that these files are unused?

PR #1832, opened 2025-09-10
fix: remove unused schema directory

evanp:
We have a top-level schema dir that we don't use; this PR removes it.


schema/reset.sql
lemilonkh, line 1: Can we move this file into the app folder maybe? At least I use it when I need to re-create my local DB for testing purposes and configure the DB permissions correctly.

lemilonkh (approved):
Thanks 👍🏼 Would like to keep the reset.sql in some subfolder of the app folder at least, otherwise good ✨

schema/reset.sql
evanp: OK! Let's move it to "scripts".

PR #1833, opened 2025-09-10
chore: remove unused Jupyter notebook

evanp:
We have an unused Jupyter notebook in our root directory. I think Mau made it two years ago. I don't think it's needed anymore, so let's get rid of it.


evanp:
@mfonsecaOEF can you check that it's OK to remove this?

PR #1844, opened 2025-09-15
fix: migration filename

thehighestprimenumber:
solves issue:

image (2)

Rename the OAuth migration file to use the .cjs extension and convert the export from ES module default to CommonJS module.exports.

This ensures the migration loads properly in the runtime environment that expects CommonJS exports and the .cjs extension.


lemilonkh (approved):
Good spot, thanks 👍🏼

PR #1846, opened 2025-09-16
feat: (General maintenance) ensure unique names for orgs

isaaccodekill:


app/src/app/[lng]/admin/CreateOrganizationModal.tsx
```diff
@@ -98,6 +101,27 @@ const CreateOrganizationModal: FC<CreateOrganizationModalProps> = ({
...
+        email.replace(/[<>"'&]/g, ""),
+      );
+      description += " " + safeEmails.join(", ");
+    }
+
+    showErrorToast({
+      title: t(fallbackTitle),
+      description,
+    });
+  };
```
Comment (lemilonkh) [inline app/src/app/[lng]/admin/CreateOrganizationModal.tsx:123]: Can this helper function be extracted so it can be reused in other places?

app/src/util/api.ts
```diff
@@ -299,6 +301,8 @@ function errorHandler(err: unknown, _req: NextRequest) {
       },
       { status: 400 },
     );
+  } else if (err instanceof CustomOrganizationError || err instanceof CustomInviteError) {
+    return NextResponse.json(err.data, { status: 409 });
```
lemilonkh, line 305: Do these error types always result in an error 409 Conflict?
Or should we have that as a parameter on the custom error types and it just defaults to 409 but can be adjusted in the constructor/ property?

And then reading the HTTP status code value from the error here.
But fine for now if that's all we use these error types for.

lemilonkh (approved):
Amazing work, thanks 👍 ✨

PR #1847, opened 2025-09-16
feat: create endpoint for  extracting json for open API endpoint

isaaccodekill:


app/src/lib/swagger.ts
```diff
@@ -2,7 +2,7 @@ import { createSwaggerSpec } from "next-swagger-doc";
 
 export const getApiDocs = async () => {
   const spec = createSwaggerSpec({
-    apiFolder: "src/app/api",
+    apiFolder: "src/app/api/v0", // Point to v0 folder to avoid the openapi.json endpoint itself
```
evanp, line 5: We have some API endpoints that aren't in v0, though! Specifically, most of the NextAuth ones. I'm not sure if this is the right move.

app/src/app/api/openapi/json/route.ts
```diff
@@ -0,0 +1,44 @@
+/**
+ * @swagger
+ * /api/openapi/json:
```
evanp, line 3: I think there's a common practice of providing the endpoint at /openapi.json and/or /swagger.json. I think we can add redirects, though.

evanp: Don't we always use JSON, though? Maybe just /api/openapi ?

evanp (approved):
This looks good. Let's merge it for now. I'm going to add rewrites for /openapi.json and /swagger.json.

PR #1871, opened 2025-09-19
feat(tools): add AGENTS.md file

lemilonkh:
Contains rules for how agentic coding agents should interact with the codebase (e.g. [opencode](https://github.com/opencode-ai/opencode)).

Please provide PR feedback or direct change suggestions on this PR for rules you think these tools should follow when interacting with our codebase.
We can also add a similar file called `.github/copilot-instructions.md` to make Copilot use these rules as well.

### DIFF
--- a/app/AGENTS.md
+++ b/app/AGENTS.md
@@ -1,0 +1,30 @@
+# AGENTS.md
+
+## Build, Lint, and Test Commands
+
+- Build: `npm run build`
+- Lint: `npm run lint`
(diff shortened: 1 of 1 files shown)


mircorudolph:
We should probably create separate AGENTS.md files per package (CC, Global API, HIAP, Climate Advisor). This one is in app/ so that is already great. Just wanting to point out that we do not try to incorporate different typescript, python, ... instructions into one root level AGENTS.md

PR #1880, opened 2025-09-22
Climate Advisor microservice setup

piotrnowakowski:
Added all endpoints according to the architecture outlined. 
Functions to save user interaction as thread and messages ( only to test with local db for now)
There is one static html website to test current functionalities as connection to CC is a focus of next sprint 
We are using openrouter as LLM provider with tested


evanp:
Around here, these are called mircoservices

PR #1881, opened 2025-09-22
Isaac/feat/on-4365-enable-test-coverage-checks

isaaccodekill:
Enable test coverage checks


app/package.json
```diff
@@ -145,6 +148,7 @@
     "jest": "^29.7.0",
     "jest-environment-jsdom": "^29.7.0",
     "k6": "^0.0.0",
+    "open-cli": "^8.0.0",
```
lemilonkh, line 151: Can we do something else than depend on this package?
I think just having a command to open the file for each of the 3 major OS's could be sufficient?
`open` on OSX and Windows (different arguments though), `xdg-open` on Linux.

lemilonkh (approved):
Great work, thanks 🙌 ✨

PR #1890, opened 2025-09-23
refactor(db): create module seeder to replace module migrations

lemilonkh:
The new setup allows us to just add a new entry in the modules.json file instead of creating a new migration each time we add a new module to journey navigator or update an existing one.

The previous migrations were removed because they would do the same thing the seeder does. They were not un-done with a separate migration as the seeder uses upserts.


evanp (changes requested):
Can we replace the deleted migration files with null migrations (empty up and down functions) and then make sure the seeder uses upserts so it is idempotent?

And let's also make sure we avoid inserting data in migrations in the future!

PR #1918, opened 2025-09-29
Package refactor

thehighestprimenumber:
* move shared components to /package dir.
* create Button components


app/src/app/[lng]/auth/forgot-password/page.tsx
```diff
@@ -71,9 +71,8 @@ export default function ForgotPassword(props: {
             {t("reset-password")}
           </Button>
           <Button
-            type="reset"
```
Comment (lemilonkh) [inline app/src/app/[lng]/auth/forgot-password/page.tsx:74]: Why did you remove the type property? I think that might prevent this button from working correctly?

app/src/app/[lng]/auth/update-password/page.tsx
```diff
@@ -108,9 +108,8 @@ export default function UpdatePassword(props: {
             {t("reset-button")}
           </Button>
           <Button
-            type="reset"
```
Comment (lemilonkh) [inline app/src/app/[lng]/auth/update-password/page.tsx:111]: Not sure here

lemilonkh (approved):
Great, thanks 🙌

app/src/app/[lng]/auth/forgot-password/page.tsx
Comment (thehighestprimenumber) [reply app/src/app/[lng]/auth/forgot-password/page.tsx:74]: because the property was duplicated

PR #1920, opened 2025-09-29
feat HIAP: bulk prioritization & action migration [ON-4682]

thehighestprimenumber:


app/src/app/api/v0/admin/bulk-hiap-prioritization/route.ts
```diff
@@ -0,0 +1,197 @@
...
+      );
+
+      results.push({
+        cityId: city.cityId,
+        cityName,
+        inventoryId: inventory.inventoryId,
+        status: HighImpactActionRankingStatus.PENDING,
+        taskId: ranking.jobId!,
+      });
+      startedCount++;
```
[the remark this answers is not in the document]
lemilonkh: Seems to be resolved already

app/src/app/api/v0/admin/project/[projectId]/hiap-jobs/route.ts
```diff
@@ -0,0 +1,193 @@
...
+      taskId: ranking.jobId,
+      actionType: ranking.type,
+      status: ranking.status,
+      createdAt: ranking.created?.toISOString(),
+    };
+  });
+
+  logger.info(
+    `HIAP Jobs API called for project ${validatedProjectId}, year: ${validatedYear}, actionType: ${validatedActionType}, found ${hiapJobs.length} jobs`,
+  );
```
Comment (lemilonkh) [inline app/src/app/api/v0/admin/project/[projectId]/hiap-jobs/route.ts:188]: It might make sense to pass this as a data object into the logger as it's useful for structured logging. See here for more details about logging structured data with Pino (which is the tool we're using for logging): https://getpino.io/#/docs/api?id=mergingobject

app/src/backend/hiap/HiapService.ts
```diff
@@ -553,3 +551,277 @@ export const fetchRanking = async (
...
+// ============================================================================
+// HIAP Action Selections Migration Functions
+// We should delete these after the migration is complete
+// ============================================================================
+
+function getSelectedActionsFileName(locode: string, type: ACTION_TYPES) {
+  return `data/selected/${type}/${locode}.json`;
+}
+
+const streamToString = async (stream: any) => {
```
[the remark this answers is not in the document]
lemilonkh: That would be better to prevent type issues

app/src/models/HighImpactActionRanking.ts
```diff
@@ -60,7 +60,7 @@ export class HighImpactActionRanking
         },
         jobId: {
           type: DataTypes.TEXT,
-          allowNull: true,
```
lemilonkh, line 63: We might need to create a migration to add this non-null constraint in the DB as well? Or is this already the case for the `job_id` column?

app/src/services/api.ts
```diff
@@ -1141,6 +1145,37 @@ export const api = createApi({
...
+        }) => response,
+      }),
+      startBulkHiapPrioritization: builder.mutation({
+        query: (data: {
+          projectId: string;
+          year: number;
+          actionType: ACTION_TYPES;
+          language?: string;
+        }) => ({
+          url: `/admin/bulk-hiap-prioritization`,
```
lemilonkh, line 1166: Does it make sense to have the hiap routes in the admin path if we plan to enable self-service HIAP in the future? Or will we make a new route to accomodate this then?

lemilonkh (approved):
Amazing work, thanks 👍 ✨

PR #1922, opened 2025-09-30
Feat/on 1986 implement climate advisor disclaimer

cephaschapa:
Changes
- Adds user disclaimer dialog prompt
- Restores clima ai
- Adds login to persist disclaimer state

image
image


app/src/components/icons.tsx
```diff
@@ -1815,3 +1815,81 @@ export const AllProjectsIcon = () => {
...
+    <svg
+      xmlns="http://www.w3.org/2000/svg"
+      width="19"
+      height="23"
+      viewBox="0 0 19 23"
+      fill="none"
+    >
+      <path
+        fillRule="evenodd"
+        clip-rule="evenodd"
```
lemilonkh, line 1830: I think this needs to be changed to clipRule to work and not give a client-side error

lemilonkh (commented):
Great work, thanks 🙌 ✨

lemilonkh (approved):

PR #1928, opened 2025-10-01
feat: ON-4672 add database, k8s, pgvector embeddings for CA

piotrnowakowski:
Proper database implementation in production 
Setting up k8s files
Setting up pg vector and embeddings for it


climate-advisor/service/app/services/openrouter_client.py
```diff
@@ -39,9 +39,11 @@ def __init__(
 
     def _headers(self, request_id: Optional[str] = None) -> Dict[str, str]:
         h = {
-            "Authorization": f"Bearer {self.api_key}",
             "Content-Type": "application/json",
         }
+        # Only add Authorization header if API key is provided
+        if self.api_key and self.api_key.strip():
+            h["Authorization"] = f"Bearer {self.api_key}"
```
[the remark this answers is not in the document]
piotrnowakowski: very unlikely situation that i will pass wrong key with different format. 
most common is empty and expired which will pass the format validation but still fail in live app. Dont see any benefits of doing it

climate-advisor/service/requirements.txt
```diff
@@ -1,10 +1,29 @@
...
+# AI and embeddings (Latest: 1.35.x)
+openai>=1.0.0
+numpy>=1.24.0
+
+# Document processing (Latest: 3.0.1)
+PyPDF2>=3.0.0
+langchain-text-splitters>=0.2.0,<0.3.0
+
+# Vector database (Latest: 0.3.x)
+pgvector>=0.2.0
```
lemilonkh, line 29: Do we need these comments with latest version numbers here or can you try to upgrade all the dependencies to the latest version instead (maybe in a separate PR?)

climate-advisor/vector_db/files/GHGP_GPC_0.pdf
lemilonkh, line 1: Do we need to keep this as a raw file in the repository or can it be pre-processed/ extracted?

climate-advisor/vector_db/services/embedding_service.py
```diff
@@ -0,0 +1,273 @@
...
+                    embedding=[],
+                    model=self.model,
+                    tokens_used=0,
+                    success=False,
+                    error="Empty text provided"
+                )
+
+            # Truncate text if too long (OpenAI has token limits)
+            if len(text) > 8000:  # Conservative limit for text-embedding-3-small
+                text = text[:8000] + "..."
```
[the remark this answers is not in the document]
lemilonkh: Add EMBEDDING_LENGTH_LIMIT constant at the top of the file so this can be adjusted more easily (and use it in both places).

climate-advisor/vector_db/utils/text_processing.py
```diff
@@ -0,0 +1,242 @@
...
+        try:
+            reader = PdfReader(file_path)
+            text_content = []
+
+            for page in reader.pages:
+                text = page.extract_text()
+                if text.strip():  # Only add non-empty pages
+                    text_content.append(text.strip())
+
+            return "\n\n".join(text_content)
```
lemilonkh, line 49: Can we do this once and output the text content to a text file that is added to the repository instead of having the raw (and quite large) text file and processing it each time?

climate-advisor/llm_config.yaml
```diff
@@ -50,18 +54,14 @@ api:
     max_retries: 3
     # Delay between retries (exponential backoff)
     retry_delay_ms: 1000
-    # Maximum tokens per request (safety limit)
-    max_tokens_per_request: 92000
```
lemilonkh, line 54: Should we keep this here for safety (in case the embedding length limit set in another file isn't applied to some queries on accident).

climate-advisor/setup.sh
lemilonkh, line 1: This file is really cool, would be neat to have a script like this for the CC web app as well that checks your local setup and configures the DB etc. 👍

k8s/ca-deploy.yml
lemilonkh, line 1: It might make sense to name this file (and maybe the others) the same as the service name for simplicity.
So having it as `climate-advisor-deploy.yml`.

climate-advisor/service/tests/test_thread_service_uuid.py
```diff
@@ -0,0 +1,22 @@
...
+from unittest.mock import AsyncMock
+
+from app.exceptions import ThreadNotFoundException
+from app.services.thread_service import ThreadService
+
+
+class ThreadServiceUUIDValidationTests(unittest.IsolatedAsyncioTestCase):
+    async def test_invalid_uuid_raises_not_found(self) -> None:
+        session = AsyncMock()
+        service = ThreadService(session)
```
lemilonkh, line 13: Might be good to add more tests for the happy path of ThreadService in the future. Can be done in a follow-up PR/ ticket though imo.

.cursorignore
```diff
@@ -0,0 +1 @@
+# Add directories or file patterns to ignore during indexing (e.g. foo/ or *.csv)
```
lemilonkh, line 1: Do we need to add some directories here? Like the raw data or installed packages in the `venv` folder?

lemilonkh (approved):
Amazing work, thanks 🙌 🍝

climate-advisor/service/tests/test_thread_service_uuid.py
piotrnowakowski: Good i will include more test for threads in future PRs

climate-advisor/llm_config.yaml
piotrnowakowski: This was used for llm calls as the embeddings have their own limits 
more over it was used to limit max tokens output but it is handled by opentouter and other llms provider we dont need to limit it unless we would want a very short answer which is not intended

k8s/ca-deploy.yml
piotrnowakowski: good point readjusted

climate-advisor/service/migrations/versions/20250118_120000_initial_schema.py
```diff
@@ -74,17 +78,51 @@ def upgrade() -> None:
     
     # Create indexes for messages table
     op.create_index('ix_messages_thread_id', 'messages', ['thread_id'])
+    
+    # ==================== VECTOR DATABASE TABLES ====================
+    # Create document_embeddings table with pgvector support
+    # Note: Using raw SQL because SQLAlchemy can't properly serialize VECTOR type
+    op.execute("""
+        CREATE TABLE document_embeddings (
+            embedding_id VARCHAR(36) PRIMARY KEY,
```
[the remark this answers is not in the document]
piotrnowakowski: done thanks

climate-advisor/llm_config.yaml
```diff
@@ -41,7 +38,14 @@ api:
...
     retry_delay_ms: 1000
-    
+
+  # OpenAI configuration for embeddings
+  openai:
+    base_url: "https://api.openai.com/v1"
+    # API key should be set via environment variable OPENAI_API_KEY
+    timeout_ms: 30000
+    embedding_model: "text-embedding-3-large"
+
```
mircorudolph, line 48: ok we are using embedding large. You can ignore my comment further down where I ask about which model

climate-advisor/vector_db/services/embedding_service.py
```diff
@@ -0,0 +1,273 @@
...
+                return EmbeddingResult(
+                    text=text,
+                    embedding=[],
+                    model=self.model,
+                    tokens_used=0,
+                    success=False,
+                    error="Empty text provided"
+                )
+
+            # Truncate text if too long (OpenAI has token limits)
```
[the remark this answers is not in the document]
piotrnowakowski: yeah we are using large this is a mistake done by codex for comments 
So this is like very unlikely as we don't use 8000 split which is embedding model token limit we use 3k so this is fallback that prob would never happend
But you are right i switched to tokenized split using tiktoken and will readjust this code to not have truncate option at all

climate-advisor/README.md
```diff
@@ -15,9 +15,11 @@ Prerequisites: Python 3.11+, pip, and Docker (for local Postgres).
```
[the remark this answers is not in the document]
piotrnowakowski: yeah my typo thanks for pointing it

mircorudolph (approved):

PR #1943, opened 2025-10-06
fix: adds login hook to manage login logic

cephaschapa:
Changes
- Adds useLogin hook to handle login logic
- Updates login page


lemilonkh (approved):
Great work 🙌 ✨ Thanks 👍

PR #1946, opened 2025-10-06
fix: adds fit-content to prevent view site button from overflowing

cephaschapa:
Changes
- Adds fit content to prevent view-site button from overflowing


lemilonkh (approved):
Great spot, thanks 👍 ✨

PR #1949, opened 2025-10-07
fix: add validation to activity type select input

cephaschapa:
changes
- Adds input validation for activity type


lemilonkh (approved):
Great, thanks 👍 👾

PR #1957, opened 2025-10-07
Move api/v0 to api/v1

evanp:
This is a big change! Our 1.0.0 version aliased api/v1 onto the api/v0. This PR does two things:

1. It moves all the files from api/v0 to api/v1, and aliases the other way (api/v0 -> api/v1)
2. It changes all the references to api/v0 in the code to api/v1 (except references to the global API).

All our tests ran well, but I'm almost certain there are at least some problems here that aren't covered by tests yet. I'd like to merge this PR early in the sprint, so we can shake out and fix any issues during the dev cycle.


app/package-lock.json
lemilonkh, line 1: These changes broke our CI before as the `npm ci` command failed because of inconsistencies between package.json and package-lock.json.

Can you confirm `npm ci` works locally before merging the PR?

lemilonkh (approved):
Great, thanks 👍🏼 
Are we changing to API v1 because of any breaking changes we're introducing or mainly because we're moving to version 1?

app/package-lock.json
evanp: Yes, it works!

PR #1985, opened 2025-10-15
fix: [on-4631] translating manual input keys in csv file

isaaccodekill:
Translates keys in csv

- Activity type
- Activity Units


lemilonkh (approved):
Great, thanks 👍 ✨

PR #2056, opened 2025-10-27
feat:HIAP: Differentiate bulk vs single prioritizations [ON-4828]

thehighestprimenumber:
since they require different endpoints, keep track of whether a ranking is part of a batch or not. Use that field do decide which API method to call.


cephaschapa (approved):
👍👍

PR #2057, opened 2025-10-28
On 4745

piotrnowakowski:
Implement remaining db tool
Implement testing step for CA 
Add tests for CA

### DIFF
--- a/app/src/app/[lng]/[inventory]/InventoryResultTab/TopEmissionsWidget.tsx
+++ b/app/src/app/[lng]/[inventory]/InventoryResultTab/TopEmissionsWidget.tsx
@@ -122,7 +122,7 @@ const TopEmissionsWidget = ({
         </Card.Body>
       </Card.Root>
     );
-  } else if (results!?.totalEmissions.total <= 0) {
+  } else if (!results || !results.totalEmissions || results.totalEmissions.total === null || results.totalEmissions.total === undefined || !results.totalEmissions.bySector || !Array.isArray(results.totalEmissions.bySector) || results.totalEmissions.bySector.length === 0) {
     return (
       <Card.Root width={"713px"} height={"448px"}>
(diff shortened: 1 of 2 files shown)


climate-advisor/service/app/models/db/thread.py
```diff
@@ -7,9 +7,10 @@
 from sqlalchemy import DateTime, String, func
 from sqlalchemy.orm import Mapped, mapped_column, relationship
 
-from sqlalchemy.dialects.postgresql import JSONB, UUID as PGUUID
+from sqlalchemy.dialects.postgresql import UUID as PGUUID
```
mircorudolph: why are we importing this as PGUUID? Is this different then below where we use UUID?

mircorudolph (approved):
Thanks!

mircorudolph:
@Codex

climate-advisor/service/app/models/db/thread.py
piotrnowakowski: yeah so we have two different one's and I was getting confused. there is a UUID from postgresql ( that's why PGUID) and the one from python. We use PGUID for db connection and uuid for the clear python type handling

.github/workflows/climate-advisor-test.yml
```diff
@@ -0,0 +1,183 @@
...
+  contents: read
+  packages: write
+  checks: write
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - develop
+    paths:
```
mircorudolph: You would probably also want to include 
on:
  workflow_dispatch:
   pull_request:
   
   and not just push

mircorudolph:
I see you added the tests to the github workflow -test.

You would also need to add this to the -dev and -tag workflow,
We want to run the pytests on all our deployments

PR #2109, opened 2025-11-04
dashboard: removed double footer - HIAP: run ranking at same time for both types [ON-4971]

thehighestprimenumber:


evanp (approved):
Works in dev

PR #2116, opened 2025-11-06
On 5025 input validation

mircorudolph:
Added stricter input validation on plan creator and plan translation API.

Before merge, needs to be checked on CC backend side


lemilonkh (approved):
This is great to have to make the API easier to use, thanks 👍 ✨

PR #2128, opened 2025-11-12
disables client side logging for test and prod

cephaschapa:
Changes
- Disables client side logging on test and prod


lemilonkh (approved):
Thanks 👍🏼

PR #2135, opened 2025-11-17
feat: add key impacts column to the actions list

amanda-eames:
- Added key_impacts (VARCHAR[]) to modelled.cap_climate_action via new Alembic migration.
- Extended GET /api/v0/climate_actions to select and return the new KeyImpacts field so can see the data alongside existing attributes.


lemilonkh (approved):
Great work, thanks 👍🏼✨

PR #2136, opened 2025-11-17
Fix: [on-5049], [on-4938] include Unvailable reasons into progress bar

isaaccodekill:


app/src/components/SegmentedProgress.tsx
```diff
@@ -101,7 +101,14 @@ export function SegmentedProgress({
         {shownValues.map((value, i) => (
           <Box
             key={i}
-            backgroundColor={value.color}
+            backgroundColor={
+              colors[i] === "striped" ? "transparent" : value.color
+            }
+            backgroundImage={
+              colors[i] === "striped"
+                ? "repeating-linear-gradient(45deg, #C5CBF5, #C5CBF5 4px, transparent 4px, transparent 8px)"
```
lemilonkh: Should we use theme colors for these color codes or is that overkill?

lemilonkh (approved):
Great work! 💪 ✨

PR #2141, opened 2025-11-17
feat(api): implement API key verification for check-hiap-jobs cron job route

lemilonkh:


k8s/cc-check-hiap-jobs.yml
```diff
@@ -27,12 +27,11 @@ spec:
                 - -c
                 - |
                   echo "Checking HIAP jobs at $(date)"
-                  curl -f -L -X GET http://cc-web:3000/api/v1/cron/check-hiap-jobs
+                  curl -f -L --header "Authorization: Bearer $(CC_CRON_JOB_API_KEY)" -X GET http://cc-web:3000/api/v1/cron/check-hiap-jobs
```
lemilonkh: How do we get this env var into the job securely from Github Secrets?

PR #2142, opened 2025-11-17
feat: [on-4929] add NE notation keys for subsectors without source data

isaaccodekill:
image

### DIFF
--- a/app/src/app/[lng]/[inventory]/data/[step]/page.tsx
+++ b/app/src/app/[lng]/[inventory]/data/[step]/page.tsx
@@ -305,7 +305,12 @@ export default function AddDataSteps() {
   }, [activeStep]);
 
   const totalStepCompletion = currentStep
-    ? clamp(currentStep.connectedProgress + currentStep.addedProgress + currentStep.reasonNEProgress + currentStep.reasonNOProgress)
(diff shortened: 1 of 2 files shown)


app/src/app/[lng]/[inventory]/data/[step]/page.tsx
```diff
@@ -775,15 +785,17 @@ export default function AddDataSteps() {
...
                     <Badge w="auto">
                       <Box
-                        boxSize={6}
+                        boxSize={3}
                         borderRadius="full"
                         backgroundImage="repeating-linear-gradient(45deg, #D1D5DB, #D1D5DB 2px, transparent 2px, transparent 4px)"
                       />
-                      {formatPercentage(currentStep.reasonNOProgress)}% <Trans t={t}>not-occurring</Trans>
+                      {formatPercentage(currentStep.reasonNOProgress)}%{" "}
+                      <Trans t={t}>not-occurring</Trans>
```
Comment (lemilonkh) [inline app/src/app/[lng]/[inventory]/data/[step]/page.tsx]: Why was the first translation key changed to include the `-progress` suffix but the other one wasn't?
Also make sure to update the respective translation files in other languages to remove the old key (if necessary), as I already merged the translation PR that added them.

PR #2156, opened 2025-11-24
fix: tooltip layout

cephaschapa:
Changes
- Adds correct layout structure to tooltip
image


lemilonkh (approved):
Amazing work, thanks 🤩✨

PR #2162, opened 2025-11-24
fix(deps): removes unused packages from CC web dependencies

cephaschapa:
Changes
- Removes unused packages from cc web package files


lemilonkh (approved):
Awesome, good to have these cleaned up a bit 🙌 ✨

PR #2167, opened 2025-11-25
Feat/on 4068 add rate limiting and throttling for apis

cephaschapa:
Changes
- Adds RateLimiter class
- Adds RaleLimiter middleware to api handler
- Adds unit tests
image


app/src/util/api.ts
```diff
@@ -5,6 +5,7 @@ import createHttpError from "http-errors";
 import { NextRequest, NextResponse } from "next/server";
 import { ZodError } from "zod";
 import { H } from "@/lib/highlight";
+import rateLimit from "express-rate-limit";
```
lemilonkh: Are you still using this? It seems it's only using the RateLimiter class that's in our codebase, not the rateLimit function from the external package?

If not, let's remove the dependency and import.

PR #2170, opened 2025-11-25
chore: upgrade NodeJS to v22

evanp:
This upgrades our NodeJS version from v20 (ancient LTS) to v22 (somewhat less ancient LTS). It seems to fix some problems with our package-lock file, probably caused by people using different versions of Node!


lemilonkh (approved):
Great, thanks 👍🏼

PR #2204, opened 2025-12-08
fix(ON-5074): Add CityId and InventoryId to action plan schema

cephaschapa:
changes
- Adds migration to include cityId and inventoryId to database
- Adds Ids to action plan schema
- Updates services
image

### DIFF
--- a/app/src/backend/hiap/ActionPlanService.ts
+++ b/app/src/backend/hiap/ActionPlanService.ts
@@ -11,6 +11,8 @@ export interface CreateActionPlanInput {
   actionId: string;
   highImpactActionRankedId?: string;
   cityLocode: string;
+  cityId?: string;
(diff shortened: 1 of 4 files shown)


app/src/backend/hiap/ActionPlanService.ts
```diff
@@ -11,6 +11,8 @@ export interface CreateActionPlanInput {
   actionId: string;
   highImpactActionRankedId?: string;
   cityLocode: string;
+  cityId?: string;
+  inventoryId?: string;
```
lemilonkh: How do we create the values for the existing actions in e.g. the Brazil inventories?
Should we have an admin route or a script to assign them based on the cityName?
Or do we need to re-generate all of the actions?

PR #2210, opened 2025-12-08
fix: [on-5092] resolve issue with tooltip

isaaccodekill:


lemilonkh (approved):
Amazing work, thanks 👍 ✨

PR #2251, opened 2025-12-16
feat: on-5075 mcp server implementation

isaaccodekill:
image
image

### DIFF
--- a/app/src/app/api/v1/mcp/route.ts
+++ b/app/src/app/api/v1/mcp/route.ts
@@ -58,17 +58,12 @@
 import { apiHandler } from "@/util/api";
 import { NextRequest, NextResponse } from "next/server";
 import { logger } from "@/services/logger";
-import { 
-  McpError, 
-  ErrorCode,
-  Tool,
-} from "@modelcontextprotocol/sdk/types.js";
+import { McpError, ErrorCode, Tool } from "@modelcontextprotocol/sdk/types.js";
 import { AppSession } from "@/lib/auth";
 
 // Import tool implementations
 import * as inventoriesTools from "@/lib/mcp/tools/inventories";
 import * as emissionsTools from "@/lib/mcp/tools/emissions";
-import * as forecastTools from "@/lib/mcp/tools/forecast";
 import * as citiesTools from "@/lib/mcp/tools/cities";
 import * as cityProfileTools from "@/lib/mcp/tools/city-profile";
 import * as actionPlansTools from "@/lib/mcp/tools/action-plans";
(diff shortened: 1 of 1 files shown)


app/src/app/.well-known/oauth-protected-resource/route.ts
```diff
@@ -0,0 +1,73 @@
+/**
+ * @swagger
+ * /.well-known/oauth-protected-resource:
+ *   get:
```
lemilonkh: Would be good to move this comment before the method definition

app/src/app/.well-known/oauth-protected-resource/route.ts
```diff
@@ -0,0 +1,73 @@
...
+  const origin = process.env.HOST || url.origin;
+  
+  logger.debug({ origin }, "OAuth protected resource discovery requested");
+
+  return NextResponse.json({
+    resource: origin,
+    authorization_servers: [origin],
+    scopes_supported: ["read", "write"],
+    bearer_methods_supported: ["header"],
+    resource_documentation: "https://github.com/Open-Earth-Foundation/CityCatalyst/wiki/CityCatalyst-Backend-API",
```
lemilonkh: This isn't quite up do date, might make sense to point to the Swagger docs page instead? Using the current HOST value (`origin` here) and appending `/docs`.

app/src/app/api/v1/.well-known/mcp-server/route.ts
```diff
@@ -0,0 +1,62 @@
+/**
+ * @swagger
+ * /api/v1/.well-known/mcp-server:
+ *   get:
```
lemilonkh: Would be good to move this before the method definition as it documents the GET method (and there can be multiple in case of multiple HTTP method for the same path).

app/src/app/api/v1/mcp/route.ts
```diff
@@ -0,0 +1,383 @@
...
+  });
+
+  toolRegistry.set("get_climate_risk_assessment", {
+    definition: riskAssessmentTools.getClimateRiskAssessmentTool,
+    handler: riskAssessmentTools,
+  });
+}
+
+// Initialize tools on module load
+registerTools();
```
lemilonkh: Would be nicer to have an init method (so this is only called when used, and maybe as a one-time init effect that is ignored for the next call), but not super urgent.

lemilonkh (approved):
Amazing work, thanks 👍 ✨

app/src/app/api/v1/mcp/route.ts
```diff
@@ -0,0 +1,383 @@
...
+  const { protocolVersion, capabilities, clientInfo } = request.params || {};
+
+  logger.info({ clientInfo, protocolVersion }, "MCP client initializing");
+
+  return NextResponse.json({
+    jsonrpc: "2.0",
+    result: {
+      protocolVersion: "2024-11-05",
+      capabilities: {
+        tools: {},
```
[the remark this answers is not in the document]
isaaccodekill: The tools: {} capability is empty in the handleInitialize function because this is following the MCP (Model Context Protocol)
 specification.

  The empty object indicates that the server supports tools, but it doesn't list them in the capabilities. The actual tool listing happens
  through a separate tools/list method call.

PR #2253, opened 2025-12-16
On 5084 ca is not answering

piotrnowakowski:


climate-advisor/service/app/tools/climate_vector_tool.py
```diff
@@ -96,15 +96,18 @@ def __init__(
...
         if settings and hasattr(settings, 'llm') and hasattr(settings.llm, 'tools'):
             config = settings.llm.tools.climate_vector_search
-            self.top_k = config.get("top_k", 5)
-            self.min_score = config.get("min_score", 0.6)
         else:
-            # Fallback defaults
-            self.top_k = 5
-            self.min_score = 0.6
+            # Load defaults from llm_config.yaml when settings not provided
+            from ..config.settings import _load_llm_config
```
mircorudolph: Lets try to avoid those relative imports

climate-advisor/service/app/tools/payload_trimmers.py
```diff
@@ -0,0 +1,464 @@
...
+    Args:
+        inventory: Dictionary representing a detailed inventory object
+    
+    Returns:
+        Trimmed inventory dictionary with city context. 
+        Returns empty dict if input is invalid.
+    """
+    try:
+        # Validate input type
+        if inventory is None:
```
mircorudolph: both here the same (see above comment)

climate-advisor/service/app/tools/payload_trimmers.py
```diff
@@ -0,0 +1,464 @@
...
+    
+    Args:
+        source: Dictionary representing a datasource object
+    
+    Returns:
+        Trimmed datasource dictionary. Returns empty dict if input is invalid.
+    """
+    try:
+        # Validate input type
+        if source is None:
```
mircorudolph: same here

climate-advisor/service/app/tools/payload_trimmers.py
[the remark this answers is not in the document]
piotrnowakowski: Isn't it clearer to have those two in one place when they are so connected ? 
Open to both tbh just thought it makes more sense

climate-advisor/service/app/tools/payload_trimmers.py
```diff
@@ -0,0 +1,464 @@
...
+    Args:
+        inventory: Dictionary representing an inventory object
+    
+    Returns:
+        Trimmed inventory dictionary with only essential fields. 
+        Returns empty dict if input is invalid.
+    """
+    try:
+        # Validate input type
+        if inventory is None:
```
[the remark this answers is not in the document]
piotrnowakowski: Hmm yeah you are right i implemented those as a habit to catch errors

PR #2255, opened 2025-12-17
fix(ci):ON-5194-e2e-ci-error

cephaschapa:
Summary
- Adds filtering to get correct elememt under strict rule
- Adds timeout for async elements


lemilonkh (approved):
Amazing, thanks 👍

PR #2420, opened 2026-03-10
adds missing sector/subsector combinations

cephaschapa:


lemilonkh (approved):
Great, thanks 👍 I suppose in the future we could make the keys lowercase and do the same with the looked up keys, so we don't have to have all combinations of upper/ lowercase in the mapping?

PR #2455, opened 2026-04-08
fix: resolve three i18n cookie bugs in middleware

pablo-ibco:
## Summary

- Fix three interrelated bugs in `app/src/middleware.ts` that prevented
  the i18n cookie (`i18next`) from being set correctly

### 1. Cookie set on undefined response
The `Set-Cookie` header was being set on `response` before it was
assigned a value. The optional chaining (`response?.headers.set(...)`)
silently no-oped, so first-time visitors never received the language cookie.

### 2. Shadowed variable in referer branch
`const response = next(req)` created a new local variable that shadowed
the outer `let response`, causing the middleware to return `undefined`
for any request with a `Referer` header.

### 3. Missing `await` on async call
`next(req)` is async but was called without `await` in the referer branch,
so `response instanceof NextResponse` always evaluated to `false`
(comparing a Promise, not a NextResponse).

## Changes

- Moved cookie-setting logic to after `response` is determined, guarded
  by `response instanceof NextResponse`
- Changed `const response` to `response =` (outer variable) in the
  referer branch
- Added missing `await` on `next(req)`
- Used `response.cookies.set()` API with proper options instead of
  raw `Set-Cookie` header

## Test plan

- [ ] Visit the app without an `i18next` cookie — verify the cookie is
      set in the response
- [ ] Visit with `Accept-Language: es` — verify redirect to `/es/` and
      cookie set to `es`
- [ ] Navigate between pages (triggers referer branch) — verify language
      persists and response is not lost
- [ ] Run existing E2E tests (`npm run e2e:test`)


lemilonkh (approved):
Good spot, thanks 👍 ✨

PR #2456, opened 2026-04-08
fix: sanitize error responses to prevent internal detail leakage

pablo-ibco:
## Summary

- Sanitize the fallback 500 error response in `apiHandler` to stop
  leaking internal error details to API clients
- Remove leaked `headers` from OpenAI API error responses

## Motivation

The `errorHandler` in `app/src/util/api.ts` has a fallback branch for unrecognized errors that was extracting `err.message` and returning it in the JSON body alongside `"Internal server error"`. This could expose:

- Database/driver error messages (e.g. connection strings, table names)
- Stack trace fragments
- Internal service details

The OpenAI error branch was also forwarding `headers` from the upstream OpenAI API response, which could contain rate limit details or other internal metadata that clients should not see.

## Changes

- **Fallback 500**: return only `{ error: { message: "Internal server error" } }`
  without the raw error string (the full error is still logged server-side
  via `logger.error(err)` on line 422)
- **OpenAI errors**: return `{ error: { name, message } }` without `headers`

## Test plan

- [ ] Trigger a 500 error and verify the response body contains only
      the generic message, not internal details
- [ ] Verify server logs still contain the full error for debugging
- [ ] Run existing API tests (`npm run ci:test`)

### DIFF
--- a/app/src/util/api.ts
+++ b/app/src/util/api.ts
@@ -466,17 +466,14 @@ function errorHandler(err: unknown, _req: NextRequest) {
       { status: 400 },
     );
   } else if (err instanceof OpenAI.APIError) {
-    const { name, status, headers, message } = err;
-    return NextResponse.json({ name, status, headers, message }, { status });
-  } else {
-    let errorMessage = "Unknown error";
-    if ((err as Object).hasOwnProperty("message")) {
-      errorMessage = (err as Error).message;
-    } else if (err instanceof Error) {
-      errorMessage = (err as Object).toString();
-    }
+    const { name, status, message } = err;
     return NextResponse.json(
-      { error: { message: "Internal server error", error: errorMessage } },
+      { error: { name, message } },
+      { status: status ?? 500 },
+    );
+  } else {
+    return NextResponse.json(
+      { error: { message: "Internal server error" } },
       { status: 500 },
     );
   }


app/src/util/api.ts
```diff
@@ -466,17 +466,14 @@ function errorHandler(err: unknown, _req: NextRequest) {
...
+    );
   } else {
-    let errorMessage = "Unknown error";
-    if ((err as Object).hasOwnProperty("message")) {
-      errorMessage = (err as Error).message;
-    } else if (err instanceof Error) {
-      errorMessage = (err as Object).toString();
-    }
     return NextResponse.json(
-      { error: { message: "Internal server error", error: errorMessage } },
```
lemilonkh, line 479: I get the idea here of not accidentally leaking database details (hoewever all of our DB structure etc. is available publicly in this repository), for now I would hold off on merging this though.
This information can be crucial for debugging and until we have Grafana fully set up on our prod environment we kind of rely on this to debug it.

pablo-ibco: @lemilonkh nice! Do we have on our roadmap when we plan to have Grafana fully set up or is it something being discussed yet?

lemilonkh:
Closing this because of the feedback. Can be reopened in the future when it becomes relevant again.

PR #2457, opened 2026-04-08
fix: align react-dom and @next/env versions with their counterparts

pablo-ibco:
## Summary

- Align `react-dom` version with `react` (18.2.0 -> 18.3.1)
- Align `@next/env` version with `next` (^14.2.5 -> ^15.5.2)

### react-dom mismatch
`react` was at 18.3.1 but `react-dom` was at 18.2.0. React expects both packages to be on the same version — mismatches can cause subtle bugs in hydration, event handling, and reconciliation.

### @next/env mismatch
`@next/env` was pinned to ^14.2.5 (resolving to 14.2.35) while `next` is at ^15.5.2 (which bundles its own @next/env at 15.5.14). This meant tests and scripts (`tests/helpers.ts`, `scripts/catalogue-sync.ts`, `e2e/global-setup.ts`, etc.) were loading env files with a different version than the app itself, which could lead to inconsistent behavior between dev/test and production.

## Changes

- `app/package.json`: `react-dom` 18.2.0 -> 18.3.1
- `app/package.json`: `@next/env` ^14.2.5 -> ^15.5.2

## Test plan

- [ ] Run `npm install` and verify no peer dependency warnings
- [ ] Run `npm run build` successfully
- [ ] Run `npm run ci:test` — tests use `@next/env` for env loading
- [ ] Run `npm run e2e:test` — e2e setup/teardown uses `@next/env`
- [ ] Verify app starts and renders correctly (`npm run dev`)

> **Note:** `package-lock.json` will need to be regenerated via
> `npm install` after merging this change.


lemilonkh (approved):
Good catch, thanks 👍 ✨

PR #2460, opened 2026-04-08
fix: replace any types in bearer token and session helpers

pablo-ibco:
## Summary

- Add a `BearerTokenPayload` interface for decoded JWT tokens
- Replace `any` return/parameter types in `getBearerToken`, `makeOAuthUserSession`, and `makeServiceUserSession`

## Motivation

All three functions in `app/src/util/api.ts` used `any` for the JWT token, which is a security-sensitive code path. This meant:

- No compile-time validation of property access (`token.sub`, `token.aud`, `token.scope`, etc.)
- Easy to introduce typos or access non-existent fields without any compiler warning
- IDE autocompletion unavailable for token properties

## Changes
```typescript
interface BearerTokenPayload {
     sub: string;
     aud: string;
     iat: number;
     exp?: number;
     client_id?: string;
     scope?: string;
}
```
- `getBearerToken`: return type `any` -> `BearerTokenPayload`
- `makeOAuthUserSession`: parameter `token: any` -> `token: BearerTokenPayload`
- `makeServiceUserSession`: parameter `token: any` -> `token: BearerTokenPayload`

## Test plan

- [ ] Run `npm run build` — TypeScript will validate all property access against the new interface
- [ ] Run `npm run ci:test` — existing auth tests should pass
- [ ] Verify OAuth login flow still works end-to-end

### DIFF
--- a/app/src/util/api.ts
+++ b/app/src/util/api.ts
@@ -197,7 +197,7 @@ async function makeOAuthUserSession(token: BearerTokenPayload): Promise<AppSessi
     throw new createHttpError.BadRequest(`Malformed Authorization header`);
   }
   return {
-    expires: token.iat,
+    expires: new Date(token.iat * 1000).toISOString(),
     user: {
       id: user.userId,
       name: user.name,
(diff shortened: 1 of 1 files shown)


lemilonkh (approved):
Nice, thanks 👍 ✨

PR #2462, opened 2026-04-08
feat: add docker-compose.yml for local development

pablo-ibco:
## Summary

- Add `docker-compose.yml` at the repo root for local development
- Add `scripts/init-databases.sql` to bootstrap both databases in a single Postgres instance

## Motivation

Setting up the local environment currently requires multiple manual steps spread across different READMEs:

1. Run `app/scripts/start-db.sh` for the app database
2. Manually create the `ccglobal` user and database for the global-api
3. Manually configure each service to connect to the right database

There is no unified way to bring up the full local stack. This makes onboarding slow and error-prone — especially for new team members.

### `docker-compose.yml`
- **db**: Postgres 16 Alpine with persistent volume, healthcheck, and an init script that creates both `citycatalyst` (primary) and `ccglobal` (global-api) databases
- **global-api**: Builds from `./global-api/Dockerfile`, connects to the `db` service. Behind the `full` profile so it only starts when explicitly requested

### `scripts/init-databases.sql`
- Creates the `ccglobal` user and database on first Postgres startup (the `citycatalyst` database is created automatically via `POSTGRES_DB`)

### Database only (most common for app development)
`docker compose up -d`

### Then run the app locally with hot reload
`cd app && npm run dev`

### Full stack (database + global-api)
`docker compose --profile full up -d`

## Test plan
- [ ] `docker compose up -d` starts Postgres and both databases are created
- [ ] `docker compose --profile full up -d` additionally starts the `global-api service`
- [ ] App connects to the database and migrations run successfully
- [ ] `docker compose down -v` cleans up volumes
- [ ] Existing `app/scripts/start-db.sh` still works independently

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,36 @@
+services:
+  db:
+    image: postgres:16-alpine
+    environment:
+      POSTGRES_USER: citycatalyst
+      POSTGRES_PASSWORD: development
+      POSTGRES_DB: citycatalyst
+    ports:
+      - "5432:5432"
+    volumes:
+      - pgdata:/var/lib/postgresql/data
+      - ./scripts/init-databases.sql:/docker-entrypoint-initdb.d/init-databases.sql
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U citycatalyst"]
+      interval: 5s
+      timeout: 5s
+      retries: 5
+
+  global-api:
+    build: ./global-api
+    ports:
+      - "8000:8000"
+    depends_on:
+      db:
+        condition: service_healthy
+    environment:
+      DB_HOST: db
+      DB_PORT: "5432"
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,36 @@
+services:
```
lemilonkh, line 1: Currently this file only starts the different database services, not the individual frontends and backend microservices. While that's fine and can certainly be helpful for local development, should we add citycatalyst, hiap, global-api etc. here for a full local deployment?

Previously we did this with e.g. minikube and our k8s config files (the idea was to not have separate configs for deployment and local development), but I think a docker-compose setup is certainly easier to get going with...

lemilonkh (approved):
Thanks, let's see if we can expand it in the future 👍

PR #2466, opened 2026-04-09
feat(api): ON-5473 order inventories by year

lemilonkh:
Also adds missing access control to city years route


cephaschapa (approved):
Looks great 👍

PR #2468, opened 2026-04-09
fix(ui): ON-5520 select correct city in journey navigator drawer

lemilonkh:
Use current city instead of current inventory because the current inventory defaults to the user's stored default inventory, which is often a different one when e.g. on the journey navigator home page/ city dashboard.


cephaschapa (approved):
Nice work

PR #2482, opened 2026-04-13
fix: adds correct url path for inventory year cards

cephaschapa:
Changes
- Fixes url path that pointed inventories to old pages.


lemilonkh (approved):
Good spot, thanks 👍

PR #2489, opened 2026-04-14
Feat/on 5392 persist user mapping overrides

cephaschapa:


app/src/app/api/v1/city/[city]/inventory/[inventory]/import/approve/route.ts
```diff
@@ -104,6 +106,99 @@ function applyPdfFieldOverrides(
...
+        exampleRows,
+      });
+    }
+
+    logger.info(
+      { cityId, headerKey, adapterType, columnCount: Object.keys(columnMapping).length },
+      "ImportMappingFeedback upserted",
+    );
+  } catch (err) {
+    logger.warn({ err, cityId, headerKey }, "Failed to persist mapping feedback (non-fatal)");
```
Comment (lemilonkh) [inline app/src/app/api/v1/city/[city]/inventory/[inventory]/import/approve/route.ts:198]: Do we need to throw an error in this case to cancel the operation or is the persisting of the mapping an optional step?

app/src/models/init-models.ts
```diff
@@ -1162,6 +1173,19 @@ export function initModels(sequelize: Sequelize) {
...
+    foreignKey: "cityId",
+    onDelete: "CASCADE",
+    onUpdate: "CASCADE",
+  });
+  City.hasMany(ImportMappingFeedback, {
+    as: "importMappingFeedbacks",
+    foreignKey: "cityId",
+    onDelete: "CASCADE",
+    onUpdate: "CASCADE",
+  });
```
lemilonkh, line 1187: Do we have a way to differentiate these between multiple import inventories processes for the same city?
Or are they deleted when it finishes/ is canceled?

lemilonkh (approved):
Great, thanks 👍

app/src/app/api/v1/city/[city]/inventory/[inventory]/import/approve/route.ts
Comment (cephaschapa) [reply app/src/app/api/v1/city/[city]/inventory/[inventory]/import/approve/route.ts:198]: Persisting the mapping is intentionally non-fatal — the import rows are already committed before this runs. A failure here only means the feedback loop won't have a prior mapping to inject next time; it doesn't affect the current import. We log a warning so the issue is visible without interrupting the user's workflow.

app/src/models/init-models.ts
cephaschapa: These are persistent city-level records, not per-import. The unique index on (city_id, header_key) means one record per city × file structure — upserted on each approval, never deleted on cancel/finish. They accumulate over time to improve future AI prompts for the same file shape. Only deleted if the city itself is removed (cascade).

PR #2494, opened 2026-04-16
ON-5342: Update Climate Advisor local Docker setup

mircorudolph:
## Summary
Update the Climate Advisor local Docker workflow so it is clearly local-only and works alongside a local CityCatalyst setup.

## Changes
- Add a committed `climate-advisor/docker-compose.yml` for local development that builds the Climate Advisor image from local source and avoids pulling a remote image.
- Change the host-exposed Climate Advisor PostgreSQL port from `5432` to `5433` to avoid clashing with CityCatalyst's local PostgreSQL on `5432`.
- Auto-run Alembic migrations on Climate Advisor startup and document the local Docker workflow in `climate-advisor/README.md`.
- Document the Docker-to-host CityCatalyst connection used for local testing via `host.docker.internal`.

Note: this branch also contains the prompt/schema updates that were already present on the branch.


piotrnowakowski:
I tested the local Docker setup and found two small doc/config cleanup items:

- `climate-advisor/.env.example:10` still points at `postgres:admin@host.docker.internal:5432/climate_advisor`, while the README/compose setup now uses `climateadvisor:climateadvisor`, DB `climateadvisor`, and host port `5433`. `docker compose up -d --build` works because `docker-compose.yml` overrides `CA_DATABASE_URL`, but a fresh local host-service setup copied from `.env.example` would use the wrong connection string.

- `climate-advisor/README.md:444` documents the health response as `{"status":"healthy","service":"climate-advisor","version":"0.1.0"}`, but the real endpoint and tests currently return `{"status":"ok"}`.

piotrnowakowski (approved):
all works lefts some comment to align the .env and Readme as we are on it 
Great work

PR #2495, opened 2026-04-16
fix: increases close delay for module card tooltip

cephaschapa:
## Changes
- Increases close delay for module card tool tips.

## Motivation
- Info buttons are the primary contextual help mechanism. When they close quickly, users lose their most accessible learning path


lemilonkh (approved):
Great, thanks 👍

PR #2496, opened 2026-04-16
Fix/on 5539 chat scroll bleeds to page

cephaschapa:
## Changes
- Adds login to lock body scroll when popover is open
- Removes unused code and vars

### DIFF
--- a/app/src/components/ChatBot/chat-popover.tsx
+++ b/app/src/components/ChatBot/chat-popover.tsx
@@ -76,6 +76,14 @@ export default function ChatPopover({
     setShowDisclaimer(false);
   };
 
+  // Lock body scroll while popover is open
+  useEffect(() => {
+    document.body.style.overflow = open ? "hidden" : "";
+    return () => {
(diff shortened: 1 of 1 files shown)


app/src/components/ChatBot/chat-popover.tsx
```diff
@@ -72,9 +71,13 @@ export default function ChatPopover({
     onOpen();
   };
 
-  const handleDisclaimerCancel = () => {
-    setShowDisclaimer(false);
-  };
+  // Lock body scroll while popover is open
+  useEffect(() => {
+    document.body.style.overflow = open ? "hidden" : "";
```
lemilonkh, line 76: The other thing we could try is applying this CSS style to the chat popup div element:

```css
overscroll-behavior: contain;
```

This should prevent the scrolling from getting out of the div?
Might be worth testing which approach works better in practice though.

lemilonkh (approved):
Great, thanks 👍 ✨

PR #2497, opened 2026-04-16
fix: appends missing population text to region/province and city labels

cephaschapa:
## Changes
- Adds missing label text on population labels

### DIFF
--- a/app/src/components/steps/GHGI/set-population-data-step.tsx
+++ b/app/src/components/steps/GHGI/set-population-data-step.tsx
@@ -374,7 +374,7 @@ export default function SetPopulationDataStep({
               fontWeight="bold"
               lineHeight="24px"
             >
-              {t("region-or-province")}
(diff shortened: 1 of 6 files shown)


app/src/components/steps/GHGI/set-population-data-step.tsx
```diff
@@ -494,7 +494,7 @@ export default function SetPopulationDataStep({
               fontWeight="bold"
               lineHeight="24px"
             >
-              {t("city")}
```
lemilonkh, line 497: It seems only the string for the region/ province population was added, not this one yet?

cephaschapa: The string for this lable already exists, I think the label translations got mixed up after the JN onboarding split

lemilonkh (approved):
Great, thanks 👍

isaaccodekill (approved):

PR #2498, opened 2026-04-16
fix: adds correct links and props that prevent onboarding flow from breaking

cephaschapa:
## Changes
- Adds corrent links on inventory details step that open on a separate tab


lemilonkh (approved):
Great 👍

PR #2500, opened 2026-04-17
ON-5580: Add qualitative ranking explanations in HIAP-MEED

mircorudolph:
## Summary
Adds an optional post-ranking explanation stage in HIAP-MEED that generates grounded qualitative explanations for top-ranked actions.
It also improves observability with run-scoped LLM artifacts and concise stage-level logging while keeping prompts centralized.

## Changes
- Add `createExplanations` request toggle and wire explanation stage into prioritizer orchestration with fail-open behavior
- Add OpenAI-based explanation service with curated evidence payload and structured output parsing
- Add run artifacts under `logs/requests/<run_id>/llm/` including `explanations_io.json`, `explanations_prompt.txt`, and failure artifact
- Update docs and mocks for explanation config/behavior and add unit/integration coverage

## Commits (2)
- dc9a8e637 feat: added explanations feat
- c79b0b2d3 chore: added implementation plan


mircorudolph:
@codex

mircorudolph:
@codex

hiap-meed/implementation-plan_5aa94eba.plan.md
```diff
@@ -0,0 +1,104 @@
+---
```
piotrnowakowski: i think that is leftover output file from planning stage not should if this is needed

hiap-meed/data/mock/city_api_mock.json
```diff
@@ -29,12 +29,12 @@
       "attribute_units": "percent",
       "attribute_category": "high"
     },
-    "transport_logistics_employment": {
+    "employment_in_transport_and_logistics": {
```
piotrnowakowski, line 32: These renamed indicator keys do not currently reach feasibility scoring. The mock city payload is parsed through `CityApiItem` before `CityData` is built, and `CityApiItem` still only models `transport_logistics_employment` and `electricity_access`. Because the new keys (`employment_in_transport_and_logistics`, `electricity_access_rate`) are not modeled there, Pydantic drops them before `city.raw` is created via `city.model_dump()`.

Feasibility then extracts city buckets from `city.raw` using the canonical keys `transport_logistics_employment` and `electricity_access`, while the action catalog looks up socioeconomic rules by `employment_in_transport_and_logistics` and `electricity_access_rate`. The result is that these city values are still treated as missing even though the fixture now contains them.

Can we either add these new names as fields/aliases in `CityApiItem` and the feasibility canonical indicator list, or keep this fixture aligned with the currently modeled keys?

piotrnowakowski (approved):

hiap-meed/app/modules/prioritizer/api.py
piotrnowakowski, line 187: `city_preference_other_text` and `excluded_actions_free_text` come straight from the request and do not currently have length limits, and `ranked_actions_json` serializes the curated payload for every returned ranked action. Since `topN` is also only bounded to be >= 1, one request can make this prompt very large before we call OpenAI.

Can we add an explanation-specific guard here before rendering/sending the prompt? Or a prompt length checker

hiap-meed/implementation-plan_5aa94eba.plan.md
mircorudolph: removed

hiap-meed/data/mock/city_api_mock.json
mircorudolph: Good catch!

hiap-meed/app/modules/prioritizer/api.py
mircorudolph: implemented a truncation strategey

PR #2501, opened 2026-04-17
fix(ui): ON-5532 page freezing when language is changed

lemilonkh:
The useEffect hooks inside of our custom useTranslation hook in `i18n/client.ts` were causing circular state updates. The language changes without them when a different one was selected, so they weren't necessary in the first place.


cephaschapa (approved):
Great thanks 👍

PR #2517, opened 2026-04-20
ON-5498: Add ranking explanations and stabilize other preferences

mircorudolph:
## Summary

Adds the optional post-ranking explanation stage in HIAP-MEED and expands the `cityStrategicPreferenceOther` flow so alignment can score free-text preferences more consistently. It also aligns the merged develop changes with this branch, updates docs, and adds deterministic and smoke-test coverage around the new LLM-backed behavior.

## Changes

- add optional explanation generation with shared OpenAI client configuration, prompt artifacts, and fail-open behavior
- stabilize `cityStrategicPreferenceOther` mapping with `temperature=0.0`, prompt few-shot examples, and clearer alignment evidence
- split ranking coverage into deterministic mocked tests plus a live LLM smoke path, and fix explanation timing artifacts/logging
- update README and architecture/pipeline docs to reflect current behavior, limitations, and testing guidance

## Commits (4)

- chore: updated docs and timing issue
- Bring the explanation-stage updates into the other-priorities branch while keeping its alignment logic and shared OpenAI client flow as the source of truth
- chore: updated the scoring mechanism of preference_other
- feat: free tet scoring on other preferences


hiap-meed/app/modules/prioritizer/services/co_benefit_mapping.py
```diff
@@ -0,0 +1,204 @@
...
+        city_preference_other_text=city_preference_other_text,
+        available_co_benefit_keys=available_co_benefit_keys,
+    )
+    system_prompt = _read_system_prompt_template()
+    client = create_openai_client()
+    completion = client.chat.completions.parse(
+        model=model_name,
+        temperature=0.0,
+        response_format=CoBenefitMappingResponse,
+        messages=[
```
piotrnowakowski, line 194: I cannot see a length check for this prompt before sending. I know it probably wont exceed but it is a good thing to have on all llm calls especially when the function receives data and builds final prompt on the go

hiap-meed/docs/pipeline-description.md
piotrnowakowski, line 612: i think your code uses now 0.75 policy + 0.15 sector + 0.05 other + 0.05 timeframe.
Alkso this description is very large and because of that tbh it does not give a lot in terms of understanding the pipeline. Can we make it more compact or include diagram flows instead

piotrnowakowski (approved):

piotrnowakowski:
One explanation/auditability issue to address across the mapping + explanation flow:

When `cityStrategicPreferenceOther` mapping fails or is not configured, the scoring path correctly falls back to neutral other-preference scoring, but the explanation path still receives the raw city free-text input without being told that the text was not actually applied to the ranking.

Relevant code paths:

- `hiap-meed/app/modules/prioritizer/services/co_benefit_mapping.py`
  - `fallback_missing_model` returns `resolved_preferred_co_benefits: []`
  - `fallback_error` also returns `resolved_preferred_co_benefits: []`
  - `score_action_other_preference_component(...)` then turns an empty resolved list into neutral `0.5`

- `hiap-meed/app/modules/prioritizer/blocks/alignment.py`
  - the neutral value becomes `other_component_value`, while evidence records `other_component_mapping_source`

- `hiap-meed/app/modules/prioritizer/orchestrator.py`
  - `generate_explanations(...)` is still called with the original `city_preference_other_text`

- `hiap-meed/app/modules/prioritizer/services/explanations.py`
  - `_build_prompt(...)` still formats the free-text into the explanation prompt
  - `_build_known_limitations(...)` still checks `other_component_is_stub`, but that flag is no longer emitted by the new alignment code

- `hiap-meed/app/modules/prioritizer/prompts/ranking_explanation.md`
  - the prompt includes `City free-text preference input: {city_preference_other_text}`

Why this matters: this can give users, and us during debugging, a false sense that the city's written "other priorities" were included in the ranking when the mapper actually failed open and treated that signal as neutral. We do not necessarily need to remove the free-text from explanations, but we need to mark what happened clearly.


Just pointing it out not sure if we want to guard or throw it out but it is worth surfacing at least to us maybe to the users.

hiap-meed/app/modules/prioritizer/services/co_benefit_mapping.py
```diff
@@ -0,0 +1,204 @@
...
+    if not unique_preferred:
+        return 0.5, []
+
+    # Only city-selected co-benefits count. Missing keys are treated as neutral `0`.
+    matched_preferred = sorted(
+        key for key in unique_preferred if key in action_co_benefits
+    )
+    total = 0
+    for key in unique_preferred:
+        impact_numeric = action_co_benefits.get(key, {}).get("impact_numeric", 0)
```
piotrnowakowski, line 160: This is unlikely with the current catalog data, but it is worth guarding here because this normalization assumes co-benefit `impact_numeric` values are always in `[-2, 2]`, while the model currently accepts any integer. 
If future upstream data accidentally sends a value from another scale, for example `3` or `5`, `other_component_value` can exceed `[0, 1]` and break the block score contract. 
A small defensive validation before adding to `total` would make this assumption explicit.

hiap-meed/docs/pipeline-description.md
mircorudolph: Actually pipeline description is now a bit stale. I created this for AT during my OOO to check the pipeline. It is not the main document and should probably be removed at some point. It is purely for alignment between AT and me

hiap-meed/app/modules/prioritizer/services/co_benefit_mapping.py
mircorudolph: Added a truncation for that input field

hiap-meed/app/modules/prioritizer/services/co_benefit_mapping.py
mircorudolph: Added validator in pydantic

PR #2521, opened 2026-04-23
ON-5593: Hide Ask AI button on onboarding screens

cephaschapa:
## Summary

Hides the Clima AI assistant button on all onboarding pages. Previously the button was visible during onboarding flows, where it is not relevant and could distract users.

## Changes

- Extend path-based hide logic in `chat-popover.tsx` to return `null` when the current path includes `/onboarding`
- Covers both known onboarding routes (`/cities/:id/GHGI/onboarding/` and `/cities/onboarding/setup/`) and any future onboarding paths

## Commits (1)

- fix: hide Ask AI button on onboarding pages

### DIFF
--- a/app/src/components/ChatBot/chat-popover.tsx
+++ b/app/src/components/ChatBot/chat-popover.tsx
@@ -84,8 +84,11 @@ export default function ChatPopover({
 
   const pathname = usePathname();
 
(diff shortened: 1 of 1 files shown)


lemilonkh (approved):
Thanks 👍

PR #2522, opened 2026-04-24
ON-5497: Update hard filter and exclusion preview flow

mircorudolph:
## Summary
This PR updates the MEED prioritizer pipeline with a stronger hard-filter flow and a new exclusion-preview path so free-text exclusions can be resolved more reliably before final ranking. It also aligns mocks/docs/tests with the updated behavior.

## Changes
- Add and wire exclusion-resolution service and prompts into prioritizer flow
- Update hard-filter, alignment, config, and orchestrator behavior
- Refresh mock inputs and docs for the new exclusion-preview/request shapes
- Add and adjust unit/integration tests for prioritization and exclusion resolution
- Remove outdated mock artifacts and deprecated mock payload files

## Commits (7)
- chore: deleted old files
- fix: fixed some path issued in artifact writer
- chore: made the free text prompt for preflight exclusions a little smaller
- fix: updated file naming and tests
- chore: updated and added mock files
- chore: cleaned up docs
- feat: added second API endpoint for pre flight checks


hiap-meed/future-work.md
```diff
@@ -1,75 +1,68 @@
+# Only add information to this file that was clearly stated. Do not invent or assume future implementation work.
```
piotrnowakowski, line 1: it is like a doc to have an overview of the changes we are going to  make and it will be kept up to date or should not be commited

hiap-meed/.env.example
```diff
@@ -30,6 +30,12 @@ HIAP_MEED_TOP_N=20
 # model used by alignment free-text -> co-benefit mapping
 HIAP_MEED_ALIGNMENT_OTHER_PREFERENCE_MODEL=
 
+# exclusion preview free-text resolver runtime config
+# set to true to call the LLM resolver for excludedActionsFreeText in preview API
+HIAP_MEED_FREE_TEXT_EXCLUSIONS_ENABLED=false
+# model used by preview free-text -> action ID matching
+HIAP_MEED_FREE_TEXT_EXCLUSIONS_MODEL=
```
piotrnowakowski, line 37: Didn't we agree to have model names and configs in llm_config or smth like that

hiap-meed/app/modules/prioritizer/blocks/alignment.py
```diff
@@ -30,14 +30,10 @@
...
-    "III": "waste",
-    "IV": "ippu",
-    "V": "afolu",
-}
+from app.modules.prioritizer.utils.sector_mapping import (
+    resolve_action_sector_tags,
+    normalize_sector_tags,
+)
 
 ACTION_TIMELINE_BUCKET_TO_PREFERENCE: dict[str, str] = {
```
piotrnowakowski, line 38: if those are hardcoded maybe worth creating config file in yaml ?

hiap-meed/app/modules/prioritizer/models.py
piotrnowakowski, line 161: `requestedLanguages` is still accepted in `PrioritizerRequestData`, but the new explanation path never consumes it. In `api.py`, `_run_for_city_input()` only forwards `topN` and `createExplanations`.
`run_prioritization()` and `generate_explanations()` have no language parameter, and the explanation prompt does not specify a target language. 
As a result, `createExplanations=true` will still return English text for non-English requests. 
Since the response only exposes a single `explanation` string, the current `requestedLanguages: list[str]` contract is not satisfiable as written.

hiap-meed/app/modules/prioritizer/utils/sector_mapping.py
```diff
@@ -0,0 +1,53 @@
+"""Canonical sector mapping helpers for prioritizer actions."""
+
+from __future__ import annotations
+
+from app.modules.prioritizer.internal_models import Action
+
+SECTOR_NUMBER_TO_TAG: dict[str, str] = {
```
piotrnowakowski, line 7: yeah i think config is needed i've seen this harcoded values in previous files to align later will be hard so better do it already

hiap-meed/app/modules/prioritizer/prompts/ranking_explanation.md
piotrnowakowski, line 4: shouldn't this prompt build with out tamplate with tags

hiap-meed/app/modules/prioritizer/blocks/alignment.py
```diff
@@ -137,7 +124,7 @@ def run(
     validate_block_component_weights()
 
     # Block 1: Pre-compute shared lookup inputs for all actions.
-    preferred_sectors = _normalize_preference_sectors(city_preference_sectors)
+    preferred_sectors = normalize_sector_tags(city_preference_sectors)
     preferred_timeframes = _normalize_timeframe_preferences(city_preference_timeframes)
     available_co_benefit_keys = list(co_benefit_mapping.ALLOWED_CO_BENEFIT_KEYS)
```
piotrnowakowski, line 129: 1. alignment.run() passes list(ALLOWED_CO_BENEFIT_KEYS) into resolve_city_preferred_co_benefits(), so the resolver sees the full canonical taxonomy, not the co-benefit keys actually present on the remaining actions.
2. In co_benefit_mapping.py, Step 3 says it keeps only keys “present in this action set”, but in practice it filters against that full taxonomy input.
3. score_action_other_preference_component() then treats any mapped key missing from an action as neutral 0, while still counting it in the normalization denominator.
4. Example: if the current action set only has air_quality, but the LLM returns ["air_quality", "water_quality"], an action with air_quality = 1 scores 0.625 instead of 0.75 because water_quality is still included in the denominator even though no action has it.
5. That means the other-preference signal gets artificially damped whenever the mapped preferences include taxonomy keys that are not actually available in the current candidate set.

piotrnowakowski (approved):

hiap-meed/future-work.md
mircorudolph: Not sure what you mean? Its for future implementation and part of the docs so everyone knows what is being implemented in the next iterations and also useful for the LLMs to know.

hiap-meed/app/modules/prioritizer/blocks/alignment.py
mircorudolph: Could make sense. But in a config file should usually be more things that would or could change. This here is just a fixed mapping that will never change

hiap-meed/app/modules/prioritizer/utils/sector_mapping.py
mircorudolph: Same here I feel. Its not a config in that sense but a mapping that will not change

hiap-meed/app/modules/prioritizer/prompts/ranking_explanation.md
mircorudolph: yes good catch

hiap-meed/app/modules/prioritizer/models.py
mircorudolph: Good catch. Will fix it

hiap-meed/app/modules/prioritizer/blocks/alignment.py
mircorudolph: Updated the docs to make the other-preference math explicit: the component is scored only against the city’s resolved preferred co-benefits, missing preferred keys on an action count as 0, and extra action co-benefits outside the city preference set do not affect the score.

PR #2523, opened 2026-04-24
chore: upgrade to NextJS 16 and React 19

lemilonkh:
- Upgrades to NextJS 16 and React 19
- Updates whole codebase using NextJS codemod
- Replaced react-latex-next with react-katex (which it uses internally) because it's more up to date
- Remove react-table as we already use @tanstack/table
- Removes deprecated users view from old settings page which used it
- Replaces Webpack build hook for OpenAPI spec with the npm script `npm run openapi:build`
- Updates ESLint config to new version, enables NextJS TypeScript linter warnings

next-runtime-env is also outdated and gives us trouble since it hasn't been updated for NextJS 15 or 16, but not direct replacement was immediately found. Might create a custom solution that writes the relevant (public) env vars to a JS file that's loaded into the page or create an API endpoint for feature flags (which this is used for).


app/package.json
```diff
@@ -83,7 +85,7 @@
     "katex": "^0.16.25",
     "lodash": "^4.17.23",
     "nanoid": "^5.1.6",
-    "next": "^15.5.15",
+    "next": "16.2.4",
     "next-auth": "^4.24.13",
     "next-runtime-env": "^3.3.0",
```
tatimoreira, line 89: `next-runtime-env` is still in `dependencies` despite being flagged as incompatible with Next.js 16 / React 19 in the PR description. The lockfile reveals it's bundling its own React 18 and React DOM 18 (`node_modules/next-runtime-env/node_modules/react`) as a workaround, meaning it's silently running a separate React version rather than the project's React 19. This could cause subtle runtime issues or unexpected behavior depending on how the package is used.

The PR description mentions two possible replacements (a public env-var JS file served at build time, or a feature-flag API endpoint), but neither is implemented here. Should we create a tech debt ticket for this?

app/eslint.config.mjs
```diff
@@ -1,21 +1,23 @@
...
+  globalIgnores([
+    "node_modules/**",
+    ".next/**",
+    "out/**",
+    "build/**",
+    "coverage/**",
+    "next-env.d.ts",
+  ]),
+  {
+    rules: {},
```
tatimoreira: I think this can be removed

app/package.json
```diff
@@ -41,24 +41,26 @@
...
-    "@next/env": "^15.5.2",
+    "@next/env": "16.2.4",
     "@nivo/bar": "^0.99.0",
     "@nivo/line": "^0.99.0",
     "@nivo/tooltip": "^0.99.0",
     "@react-email/components": "^1.0.1",
     "@react-pdf/renderer": "^4.3.2",
     "@reduxjs/toolkit": "^2.2.7",
     "@tanstack/react-table": "^8.21.2",
+    "@types/react-katex": "^3.0.4",
```
tatimoreira: should we add `@types/react-katex` to `devDependencies` instead of `dependencies`. Type declaration packages only exist at compile time they have no runtime value and get compiled away entirely. 

This means `@types/react-katex` will be installed in production environments unnecessarily. Small issue, but easy to fix before merge.

app/src/components/ChatBot/DialogItem.tsx
```diff
@@ -3,9 +3,9 @@ import { ReactElement } from "react";
 import { BodyLarge } from "@/components/package/Texts/Body";
 
 interface DialogItemProps {
-  icon: ReactElement;
+  icon: ReactElement<any>;
```
tatimoreira: `ReactElement<any>` is unnecessary noise — in React 19, `ReactElement` without a type argument already defaults to `any` props. The plain `ReactElement` form is cleaner and is what the React 19 types recommend.

app/src/components/GHGI/inventory-result/EmissionBySectorChart.tsx
```diff
@@ -140,15 +140,15 @@ function CustomCombinedBarLayer<D extends BarDatum>({
         const handleMouseEnter = (e: React.MouseEvent) => {
           setFocusedBar(combinedPoint.data.year);
           showTooltipFromEvent(
-            customTooltip(combinedPoint) as React.ReactElement,
+            customTooltip(combinedPoint) as React.ReactElement<any>,
```
tatimoreira: Same as above — `as React.ReactElement<any>` can just be `as React.ReactElement`. The `<any>` adds no type safety and wasn't needed before React 19 either; this cast was valid without it.

app/src/components/ui/avatar.tsx
```diff
@@ -11,7 +11,7 @@ export interface AvatarProps extends ChakraAvatar.RootProps {
   src?: string
   srcSet?: string
   loading?: ImageProps["loading"]
-  icon?: React.ReactElement
+  icon?: React.ReactElement<any>
```
tatimoreira: `React.ReactElement<any>` → `React.ReactElement`. `ReactElement` already accepts any props by default in React 19.

app/package.json
lemilonkh: Good point, I'll create a ticket

app/src/components/ChatBot/DialogItem.tsx
lemilonkh: Done. Interesting, because this was added by the NextJS 16/ React 19 codemod. Maybe it has been changed since then.

app/src/components/PlanDetailsBox.tsx
```diff
@@ -58,9 +58,9 @@ const PlanDetailsBox: React.FC<PlanDetailsBoxProps> = ({ organization }) => {
       <BodyLarge color="content.tertiary">
         {t("contact-us-to-upgrade")}{" "}
         <Link href={`mailto:${process.env.NEXT_PUBLIC_SUPPORT_EMAILS}`}>
-          <BodyLarge color="content.link">
+          <chakra.span color="content.link">
```
cephaschapa: Could also be `<Text as="span">...</Text>`?

app/package.json
lemilonkh: https://linear.app/openearth/issue/CC-514/replace-package-next-runtime-env-with-other-package-or-custom-solution

cephaschapa (approved):
Great work 👍👍👍

PR #2524, opened 2026-04-24
fix(ui): ON-5530 keep inventory details step data alive

lemilonkh:
Step state is hoisted to parent component (onboarding setup page) in order to keep it around between step navigations (going back from the population or confirmation steps).


cephaschapa (approved):
Great work, 👍

PR #2531, opened 2026-04-28
feat(storage): migrate inventory file storage from PostgreSQL BYTEA to S3

cephaschapa:
The import pipeline was storing raw file bytes as BYTEA directly in Postgres and re-reading the full buffer from the DB on every background job (upload validation, AI interpretation, PDF extraction). On a pod capped at 1 GB RAM, a single 20 MB XLSX could consume up to ~250 MB across its pipeline stages, limiting safe concurrency to ~3 uploads.

Changes

Migration — adds s3_key VARCHAR(512) column to ImportedInventoryFile; data BYTEA kept nullable for backfill compatibility

InventoryFileStorageService (new) — thin S3 wrapper with uploadFile, getFileBuffer, deleteFile; key convention imports/{cityId}/{inventoryId}/{fileName}; exposes isS3Configured() flag

import/route.ts — POST handler uploads buffer to S3 and stores s3Key; background job resolves buffer via S3; includes try/catch around req.formData() for large-file errors

interpret/route.ts — fetches buffer from S3 by s3Key instead of reading importedFile.data
extract/route.ts — same for PDF extraction path

Dev fallback — when AWS_IMPORT_FILES_BUCKET is not set, all three routes fall back to the BYTEA data column so local dev works without any AWS setup


lemilonkh (approved):
Great! This might be the better solution after all 👍

PR #2551, opened 2026-05-06
Feat/navigation related improvements

cephaschapa:
Changes
- Adds breadcrumbs to hero section
- Alights data page breadcrumbs hierarchy styles
- Adds conditional start here badge on ghgi module card
- Changes module link behaviour and replaces link icon
- Change Lean nav link to Leaning Hub


app/src/app/[lng]/[inventory]/data/[step]/[subsector]/page.tsx
```diff
@@ -258,20 +258,30 @@ function SubSectorPage(props: {
                     <BreadcrumbLink
                       href={pathname.replace(`/${step}/${subsector}`, "")}
                       color="content.tertiary"
+                      textDecoration="none"
+                      _hover={{
+                        textDecoration: "underline",
+                        textUnderlineOffset: "4px",
+                      }}
```
Comment (lemilonkh) [inline app/src/app/[lng]/[inventory]/data/[step]/[subsector]/page.tsx:265]: Should we add this as the default styling for the BreadcrumbLink component? This way we don't need to repeat it each time. (either in the theme or in the component/ a wrapper).

app/src/i18n/locales/en/dashboard.json
```diff
@@ -304,5 +304,7 @@
   "subcategory": "Sub-category",
   "inventory-versions-restore-failed": "Failed to restore version",
   "legend": "Legend",
-  "citycatalyst": "CityCatalyst"
+  "citycatalyst": "CityCatalyst",
+  "breadcrumb-hiap": "High Impact Action Plan",
```
lemilonkh: I think the HIAP module is called "Actions & Plans" in the frontend (at least in the list of modules)?

lemilonkh (approved):
Great, thanks 🙌

app/src/i18n/locales/en/dashboard.json
cephaschapa: Yes, thank you for spotting it.

app/src/app/[lng]/[inventory]/data/[step]/[subsector]/page.tsx
Comment (cephaschapa) [reply app/src/app/[lng]/[inventory]/data/[step]/[subsector]/page.tsx:265]: good point

PR #2555, opened 2026-05-08
chore: get rid of langchain rework CA to use UV instead of pip

piotrnowakowski:
- get rid off langchain rewriting the text splitter 
- write test and benchmark to make sure the performance is aligned after switch ( not sure if we still need it but i included it for visibility)
- rewrite repo to use uv instead of pip


.github/workflows/climate-advisor-develop.yml
```diff
@@ -10,6 +10,8 @@ on:
     branches: ["develop"]
```
mircorudolph, line 10: We are not watching climate-advisor/vector_db/** even though we watch this in /test workflow.

I would suggest watching the entire climate-advisor path. This might trigger some unneccesary workflows but we ensure to not forget any path by accident.

.github/workflows/climate-advisor-tag.yml
```diff
@@ -10,6 +10,8 @@ on:
   push:
     paths:
```
mircorudolph, line 11: We are not watching climate-advisor/vector_db/** even though we watch this in /test workflow.

I would suggest watching the entire climate-advisor path. This might trigger some unneccesary workflows but we ensure to not forget any path by accident.

climate-advisor/service/app/config/settings.py
```diff
@@ -391,7 +391,7 @@ def model_post_init(self, __context: Any) -> None:
                 self.langsmith_tracing_enabled = False
                 return
```
mircorudolph, line 393: Isn't the new env var just called "LANGSMITH_TRACING"?

I am usually not a big fan of keeping backwards compatible naming especially since we have full controll over the version we are using. So lets consider removing all the alias naming and so and keeping a lean code.

.github/workflows/climate-advisor-develop.yml
mircorudolph, line 1: The is no update to .github/dependabot.yml. This is essential to move also to uv now that pip has been removed

climate-advisor/vector_db/utils/text_processing.py
mircorudolph, line 1: We could just add this package: [langchain-text-s
langchain-text-s](https://pypi.org/project/langchain-text-splitters/)

Then we would not need to build this ourselfes and we would still remove most of langchain packages from the repo.

mircorudolph (approved):

.github/workflows/climate-advisor-develop.yml
piotrnowakowski: adjusted

.github/workflows/climate-advisor-tag.yml
piotrnowakowski: adjusted

.github/workflows/climate-advisor-develop.yml
piotrnowakowski: updated

PR #2569, opened 2026-05-12
ON-5590: Sub-sector impact mapping and clearer prioritization explanations

mircorudolph:
## Summary

This work aligns MEED prioritization impact scoring with **sub-sector level** inventory mapping, tightens **emissions validation**, and refreshes **ranking explanations** so they read more clearly for city users. It also removes committed `logs_temp` request artifacts, updates mocks and docs, and adds or extends tests (including E2E mock coverage).

## Changes

- **hiap-meed:** Sub-sector impact mapping (`impact.py`, `subsector_mapping.py`), model/API/orchestrator updates, explanation prompts and `explanations` service, config and README
- **hiap-meed:** Mock data and pipeline docs; delete large `logs_temp` JSON payloads from the tree

## Commits (2)

- chore: added better validation for emissions, adapted explanations to be more readable for intended users
- chore: made impact mapping true sub sector level mapping


hiap-meed/app/modules/prioritizer/blocks/impact.py
```diff
@@ -129,58 +260,105 @@ def run(
...
+        matched_city_subsector_keys = activity_mapping_metadata[
+            "matched_subsector_keys_by_action_id"
+        ].get(action.action_id, [])
+
         total_reduction_amount = 0.0
         if reduction_multiplier is not None:
             total_reduction_amount = sum(
-                city_emissions_by_gpc_ref[gpc_ref] * reduction_multiplier
-                for gpc_ref in matched_city_gpc_refs
+                city_emissions_by_subsector_key[subsector_key] * reduction_multiplier
```
piotrnowakowski, line 322: This makes Impact scores leave the documented `[0,1]` range when valid AFOLU negative emissions are present. Because AFOLU negatives are allowed and counted as matchable, a matched `V.*` subsector contributes a negative `total_reduction_amount`, while the same negative value is also included in `total_city_emissions`. For example, with `{"I.1": 100, "V.2": -50}`, an AFOLU-targeting action can score `-0.64`, and an action targeting `I.1` can score `1.48`. We should normalize the denominator and reduction contribution so removals do not produce negative or >1 Impact scores before this ranking change ships.

hiap-meed/app/modules/prioritizer/prompts/ranking_explanation.md
```diff
@@ -7,13 +7,17 @@ Write one short explanation for each ranked action using only the provided evide
...
-- Explain why the action is placed at its rank in plain language.
+- Explain why the action is placed at its rank in plain language for a non-technical city user.
+- Focus on the biggest ranking drivers, especially the provided `main_strengths` and `main_constraints`.
 - Highlight impact, alignment, and feasibility drivers when the evidence supports them.
 - Briefly acknowledge any listed `known_limitations` without adding speculation.
 - Write every explanation in English.
 - Do not mention raw numeric scores, equations, decimals, or hidden scoring logic.
 - Do not invent city preferences, legal constraints, policy support, or implementation facts that are not present.
+- Do not infer extra benefits or implementation facts from the action ID or action theme alone.
+- Avoid meta phrases like `the evidence shows`, `the payload indicates`, `mixed profile`, or `grounded in`.
```
piotrnowakowski, line 18: The example output later in this file still conflicts with this new instruction because it uses `the evidence shows` and `the payload indicates`. Since examples often steer model style strongly, we should rewrite that sample explanation using the plain-language style requested by the prompt so it does not teach the model to use the exact phrases we are trying to suppress.

piotrnowakowski (commented):
Two review comments from the branch review.

hiap-meed/app/modules/prioritizer/services/explanations.py
```diff
@@ -370,13 +322,100 @@ def _build_feasibility_signals(
...
+        )
+
+    if bool(alignment_evidence.get("sector_match", False)):
+        strengths.append("Matches the city's preferred sector.")
+
+    if alignment_evidence.get("timeframe_match_label") == "preferred_match":
+        strengths.append("Fits the city's preferred implementation timeframe.")
+
+    policy_signals_count = int(alignment_evidence.get("policy_signals_count", 0))
+    if policy_signals_count > 0:
```
piotrnowakowski: This turns any non-zero `policy_signals_count` into "Has supportive policy context...", but the alignment block derives actual support from `policy_support_score`, not from raw signal count. That means we can generate a positive explanation even when policy support is `0.0`, which contradicts the scored evidence. I'd gate this on `policy_component_score > 0` (or another explicit support threshold) instead of the presence of policy rows alone.

piotrnowakowski (commented):
One inline note from Codex.

hiap-meed/app/modules/prioritizer/prompts/ranking_explanation_system.md
piotrnowakowski, line 19: can't we specify the fields here ? they change on each request ?

hiap-meed/docs/detailed-block-architecture.md
```diff
@@ -108,7 +108,7 @@ Canonical score policy:
 Current implementation detail:
 
 - `impact_block_score = (0.80 × reduction_share_of_city_emissions) + (0.20 × timeline_score)`
-- `reduction_share_of_city_emissions` is computed from matched action `gpc_reference_number` keys only.
+- `reduction_share_of_city_emissions` is computed from matched action `sector.subsector` keys.
```
piotrnowakowski, line 115: This hunk updates the active Impact join to `sector.subsector`, but the output description a few lines above still says "top contributing activities and multipliers". The branch now emits subsector-level evidence (`subsector_contributors`), not activity-level contributors, so that bullet looks stale relative to the current payload.

piotrnowakowski (commented):
One doc note from Codex.

hiap-meed/docs/service-architecture.md
```diff
@@ -99,7 +99,7 @@ sequenceDiagram
 
     CC->>API: POST /v1/prioritize PrioritizerApiRequest (meta + requestData.cityDataList)
     Note over API: FastAPI validates request body (Pydantic)
-    API->>Orch: run_prioritization(locode, city_emissions_by_gpc_ref, clients, per_city_options...)
+    API->>Orch: run_prioritization(locode, city_emissions_context, clients, per_city_options...)
```
piotrnowakowski, line 102: This parameter rename is correct, but the same sequence block still documents CityCatalyst calling `POST /v1/prioritize` with `PrioritizerApiRequest`. The app code in this repo still posts the legacy `/prioritizer/v1/start_prioritization` and `/prioritizer/v1/start_prioritization_bulk` payloads from `app/src/backend/hiap/HiapApiService.ts`, so if this diagram is meant to describe current end-to-end behavior it is ahead of the actual integration. If this is target-state documentation, it would help to label it that way explicitly.

piotrnowakowski (commented):
One doc note from Codex.

piotrnowakowski:
Doc note: `hiap-meed/README.md` still says each block computes named component values in `0..1`, but the current Impact implementation can still go outside that range on mixed-sign AFOLU inputs. I would only keep this wording if we treat it as the intended contract and fix the AFOLU scoring path accordingly; otherwise the docs currently overstate the guarantee.

piotrnowakowski (approved):

hiap-meed/app/modules/prioritizer/prompts/ranking_explanation_system.md
mircorudolph: We could but it is better to do this in the user prompt. The system here for high level rules and the user prompt defines them in detail. Otherwise both system and user would just be the same. Technically for such a setup maybe even only user pompt might be enough since we control both.

hiap-meed/app/modules/prioritizer/blocks/impact.py
mircorudolph: Nice catch and a real bug!

PR #2571, opened 2026-05-13
ON-5686: Add city context API and prioritization tests

mircorudolph:
## Summary

Extends HIAP-MEED prioritization with city context via the city attributes API, stricter upstream contract validation for clearer failures, and broader integration/unit coverage (including optional live city-attributes API tests when configured).

## Changes

- Add reusable HTTP client wrapper, city attributes client, and orchestration changes to fetch and pass city context through prioritization.
- Tighten prioritizer/API models and validation; refresh mock city payloads and docs to match contracts and field names.
- Add `hiap-meed/tests/integration/test_city_attributes_live_api.py` and expand smoke, E2E mock, and unit tests around data clients and prioritizer blocks.

## Commits (3)

- chore: added city context API and http wrapper
- chore: made API contracts strict for fail fast and better validation
- chore: added API tests, fixed field names

### DIFF
--- a/hiap-meed/app/modules/prioritizer/api.py
+++ b/hiap-meed/app/modules/prioritizer/api.py
@@ -44,6 +44,7 @@ from app.services.data_clients import (
     get_legal_data_api_client,
     get_policy_signals_data_api_client,
 )
+from app.services.http_client import UpstreamApiError
 from app.utils.artifacts import ArtifactWriter
 
 
@@ -55,6 +56,22 @@ router = APIRouter(tags=["prioritization"])
 def _error_payload(request_id: str, message: str) -> dict[str, str]:
     """Build a consistent JSON error body including the request ID."""
     return {"request_id": request_id, "error": message}
+
+
+def _upstream_error_payload(
+    request_id: str,
+    error: UpstreamApiError,
+) -> dict[str, str | int]:
+    """Build a consistent JSON error body for upstream dependency failures."""
+    payload: dict[str, str | int] = {
+        "request_id": request_id,
+        "error": error.message,
+    }
+    if error.upstream_status_code is not None:
+        payload["upstream_status_code"] = error.upstream_status_code
+    if error.url is not None:
+        payload["upstream_url"] = error.url
+    return payload
 
 
 def _extract_city_emissions_context(city_input: FrontendCityInput) -> CityEmissionsContext:
@@ -256,6 +273,17 @@ def preview_exclusions(
             status_code=422,
             detail=_error_payload(request_trace_id, str(error)),
         ) from error
+    except UpstreamApiError as error:
+        logger.warning(
+            "Exclusion preview upstream dependency failed request_id=%s status_code=%s error=%s",
+            request_trace_id,
+            error.status_code,
+            error,
+        )
+        raise HTTPException(
+            status_code=error.status_code,
+            detail=_upstream_error_payload(request_trace_id, error),
+        ) from error
     except Exception as error:
         logger.exception("Exclusion preview failed request_id=%s", request_trace_id)
         raise HTTPException(
@@ -365,6 +393,17 @@ def prioritize(
         raise HTTPException(
             status_code=422,
             detail=_error_payload(request_trace_id, str(error)),
+        ) from error
+    except UpstreamApiError as error:
+        logger.warning(
+            "Prioritization upstream dependency failed request_id=%s status_code=%s error=%s",
+            request_trace_id,
+            error.status_code,
+            error,
+        )
+        raise HTTPException(
+            status_code=error.status_code,
+            detail=_upstream_error_payload(request_trace_id, error),
         ) from error
     except Exception as error:
         logger.exception("Prioritization failed request_id=%s", request_trace_id)
--- a/hiap-meed/app/modules/prioritizer/models.py
+++ b/hiap-meed/app/modules/prioritizer/models.py
@@ -547,7 +547,7 @@ class UpstreamApiContext(BaseModel):
 class UpstreamMeta(BaseModel):
     """Common metadata envelope returned by upstream APIs."""
 
-    model_config = ConfigDict(extra="forbid")
+    model_config = ConfigDict(extra="ignore")
 
     generated_at_utc: str
     api_context: UpstreamApiContext
@@ -559,7 +559,7 @@ class UpstreamMeta(BaseModel):
 class UpstreamDatasource(BaseModel):
     """One datasource entry attached to upstream city metadata."""
 
-    model_config = ConfigDict(extra="forbid")
+    model_config = ConfigDict(extra="ignore")
 
     datasource_name: str
     publisher_name: str | None = None
@@ -624,7 +624,7 @@ class CityApiItem(BaseModel):
 class ActionImpactEntry(BaseModel):
     """Single impact entry (emissions or co-benefit category) for one action."""
 
-    model_config = ConfigDict(extra="forbid")
+    model_config = ConfigDict(extra="ignore")
 
     sector_number: str
     subsector_number: list[int] = Field(min_length=1)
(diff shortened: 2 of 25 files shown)


hiap-meed/app/modules/prioritizer/api.py
```diff
@@ -366,6 +394,17 @@ def prioritize(
             status_code=422,
             detail=_error_payload(request_trace_id, str(error)),
         ) from error
+    except UpstreamApiError as error:
```
piotrnowakowski, line 397: This helps for transport/HTTP failures, but upstream schema/config validation can still be reported as `422` by the preceding `except ValueError` block. `CityApiResponse.model_validate(payload)` in the new city client can raise `ValidationError`, and `ValidationError` is a `ValueError`, so an upstream contract drift would currently look like a caller request error. Could we narrow the `422` path to true caller-input failures and wrap upstream payload validation closer to the city client as `UpstreamApiError(502)` instead?

hiap-meed/app/services/city_attributes_api.py
```diff
@@ -0,0 +1,54 @@
...
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+from app.modules.prioritizer.internal_models import CityData
+from app.modules.prioritizer.models import CityApiResponse
+from app.services.http_client import get_json_with_retries
+
+CITY_ATTRIBUTES_BASE_URL = "https://ccglobal.openearth.dev"
```
piotrnowakowski: The new city-context path is pinned to one host here. If this service is intentionally always meant to call the production city-attributes backend, can we make that explicit in config/docs? If not, this looks like a good candidate for parameterization so dev/test can point at different upstreams without code changes. Even if we stay on one host today, leaving it hard-coded makes environment-specific verification harder later.

hiap-meed/app/services/city_attributes_api.py
```diff
@@ -0,0 +1,54 @@
...
+@dataclass
+class CityAttributesApiService:
+    """Fetch and map city context from the upstream city attributes API."""
+
+    base_url: str = CITY_ATTRIBUTES_BASE_URL
+
+    def _build_city_url(self, locode: str) -> str:
+        """Return the full upstream city attributes URL for one locode."""
+        normalized_locode = locode.strip().upper()
+        return f"{self.base_url.rstrip('/')}/api/v0/city_attributes/{normalized_locode}"
```
piotrnowakowski: We are interpolating the normalized locode directly into the path here, but these values include spaces like `CL IQQ`. `httpx` will usually normalize the request, but we also persist and surface this URL in `source_metadata` and error payloads. Could we quote the path segment explicitly so request construction is unambiguous and we do not expose raw-space URLs downstream?

hiap-meed/app/modules/prioritizer/models.py
```diff
@@ -509,40 +537,79 @@ class ExclusionPreviewApiResponse(BaseModel):
...
     attribute_category: str | None = None
+    datasource: str | None = None
+    version_label: str | None = None
 
 
 class CityApiItem(BaseModel):
-    """City item shape returned by upstream `GET /v1/cities/{locode}`."""
+    """City item shape returned by upstream `GET /api/v0/city_attributes/{locode}`."""
+
+    model_config = ConfigDict(extra="ignore")
```
piotrnowakowski, line 600: The docs/README now describe the city upstream contract as strict, but this DTO still uses `extra="ignore"` here, and the same pattern also appears on `CityApiMeta` and `CityIndicator`. If additive upstream fields are intentionally tolerated, I think the docs should say that explicitly and reserve `strict` for the caller-facing request DTOs. Otherwise, if strictness is the goal, these city-response models need to move to `extra="forbid"`.

hiap-meed/app/modules/prioritizer/models.py
```diff
@@ -509,40 +537,79 @@ class ExclusionPreviewApiResponse(BaseModel):
...
+
+    generated_at_utc: str
+    api_context: UpstreamApiContext
+    datasources: list[UpstreamDatasource] = Field(default_factory=list)
 
 
 class CityIndicator(BaseModel):
     """Single city indicator object used in city API payload."""
 
+    model_config = ConfigDict(extra="ignore")
```
piotrnowakowski, line 588: The README says the city upstream contract is strict and that unknown or misnamed fields are rejected, but this keeps the indicator object lenient. With `extra="ignore"`, payloads like `attributeCategory` are accepted and silently become `attribute_category=None`, which changes feasibility inputs without failing fast. If the intended behavior is truly strict contract validation, this probably needs `extra="forbid"` here, and likely the same treatment for the surrounding city response models, with tests updated to assert rejection rather than silent dropping.

piotrnowakowski (commented):
Adding one inline comment on the city contract strictness behavior.

hiap-meed/docs/pipeline-description.md
```diff
@@ -163,6 +163,13 @@ File:
 
 - `data/mock/city_api_mock.json`
 
+Shape note:
+
+- This mock now follows the upstream `GET /api/v0/city_attributes/{locode}` schema.
+- The city payload uses snake_case names such as `city_name`, `country_code`, `population_size`, `population_density`, and `area_km2`.
+- The city payload also includes a `population` indicator object alongside the top-level `population_size` and `population_density` fields.
+- The city API contract is strict, so unexpected keys or wrong casing now fail validation instead of being ignored.
```
piotrnowakowski: This overstates the current city contract. The checked-in mock still uses `populationSize` / `populationDensity`, and the current DTO accepts those via aliases while also ignoring unexpected keys. As written, this reads as `snake_case only` and `fail on extra keys`, which is stricter than the implementation in this branch.

piotrnowakowski (commented):
Docs follow-up comments on the current hiap-meed pipeline description.

piotrnowakowski (approved):

piotrnowakowski:
`hiap-meed/docs/pipeline-description.md` around lines 1154 and 1228:

This section still describes the old singular `explanation` contract and frames multilingual output as future work. The current branch already returns `explanations` keyed by language and also exposes `/v1/explanations/translate`, so these paragraphs should be updated to describe the current response shape rather than the pre-multilingual behavior.

piotrnowakowski:
`hiap-meed/docs/pipeline-description.md` around line 125:

Listing `activities[].activityType` as a field that affects the result is too strong for the current implementation. The branch preserves `activityType` for future activity-data-level matching, but ranking still runs on subsector keys only by default, and the enabled flag currently uses a stub that does not change the matches. I'd reword this as retained for future matching/diagnostics rather than active scoring input.

piotrnowakowski:
`hiap-meed/docs/pipeline-description.md` around line 569:

This subsector-based Impact description looks right, but section 3.3 earlier in the doc still says the preprocessing step sums emissions "for each emissions category". The code now normalizes each GPC ref to a `sector.subsector` key before aggregation, so that earlier setup section should be updated to use the same subsector terminology.

piotrnowakowski:
`hiap-meed/docs/pipeline-description.md` around line 330:

The validation bullets here imply that callers must provide all three `weightsOverride` keys. The current implementation merges any provided override keys onto the defaults, so partial overrides are valid as long as the resolved set sums to `1.0`. The docs should reflect that behavior to avoid overstating the request contract.

hiap-meed/docs/pipeline-description.md
piotrnowakowski: This overstates the current city contract. The checked-in mock still uses `populationSize` / `populationDensity`, and the current DTO accepts those via aliases while also ignoring unexpected keys. As written, this reads as "snake_case only" and "fail on extra keys", which is stricter than the implementation in this branch.

piotrnowakowski (commented):
Adding one inline documentation comment on the city contract wording.

hiap-meed/app/modules/prioritizer/api.py
mircorudolph: solved by moving this into the city client

hiap-meed/app/services/city_attributes_api.py
mircorudolph: added that clenly in docs and in env.example and also updated the github workflows

hiap-meed/app/modules/prioritizer/models.py
mircorudolph: Updated the docs that upstream additional fields are allowed wiht extra="ignore" being the default

PR #2574, opened 2026-05-14
feat(test): use expectStatusCode helper consistenly in all tests

lemilonkh:
Makes debugging tests a lot easier by actually adding the request body to the error message, so you don't just get the status code but the actual error from the request.


cephaschapa (approved):
Good catch 👍👍👍

PR #2579, opened 2026-05-16
ON-5687/ON-5688: Align hiap-meed with legal and policy APIs

mircorudolph:
## Summary

This PR aligns `hiap-meed` with the live legal assessments API and the live action policy scores API. It replaces the old legal requirement mock flow and the old policy-signals mock flow, wires both contracts through ranking, artifacts, mocks, docs, and tests, and keeps the public frontend response lean while preserving richer upstream detail in artifacts.

## Logic changes

The old legal mock modeled multiple per-action legal requirements with fields like `strength` and `alignment_status`. Ranking used those requirement rows in two ways: hard filtering when mandatory or required requirements were not aligned, and a soft legal feasibility component based on softer requirement alignments.

The new legal API is flatter and uses one assessment row per action with fields like `srcActionId`, `countryCode`, `verdictCategory`, and `verdictScore`. The backend now:

- fetches legal rows by request `countryCode` and maps them by `srcActionId`
- hard-filters only when `verdictCategory = blocked`
- uses `verdictScore` directly as the legal part of feasibility
- falls back to neutral `0.5` when no legal row exists or when legal scoring is missing

For policy, the backend now consumes the live city-scoped action policy scores endpoint directly. The Alignment block uses upstream `policy_support_score` as the policy component score. The upstream policy endpoint does not return a score for every action, so when an eligible action has no policy score row, the Alignment block currently uses `0.0` for the policy component and records `policy_score_present=false` in block evidence.

## Changes

- add dedicated API services for legal assessments and action policy scores, plus shared top-level JSON list support in the HTTP client
- switch the prioritizer legal path to country-based legal assessment fetching, `verdictCategory` hard filtering, and `verdictScore` feasibility scoring with neutral `0.5` fallback
- switch the prioritizer policy path to the live action policy scores contract and replace `actions_policy_signals_api_mock.json` with `action_policy_scores_api_mock.json`
- keep richer policy evidence and upstream metadata in artifacts while keeping the frontend alignment summary at score level only
- update docs, env examples, and unit, smoke, e2e, and live-contract tests for the legal and policy API schemas

## Notes

- The files under `hiap-meed/logs/requests/prioritization/...` are temporary artifacts added for inspecting the last run and reviewing the new legal and policy artifacts.
- In the inspected CL/IQQ run, legal coverage is still sparse: 48 legal rows were fetched, 19 actions were blocked, and many valid actions still used the neutral legal fallback. That is not a code bug by itself, but it is important reviewer context because legal feasibility is still often neutral with the current upstream coverage.
- In the same run, policy scores were fetched for 102 actions, while 53 valid actions had no policy score row and therefore received the current `0.0` policy fallback.

## Commits (5)

- feat: implemented new API with data models
- chore: added doc updated and tests
- chore: fixed artifacts
- chore: implement policy API
- chore: avoid repeated policy source metadata


amanda-eames (approved):
Looks great, no notes

PR #2581, opened 2026-05-19
fix(module-card): updates text style and tag border

cephaschapa:
Changes
- updates text style and tag border


lemilonkh (approved):
Great, thanks 👍

PR #2583, opened 2026-05-19
fix(module-card): removes start here button and local storage logic

cephaschapa:
Changes
- Removes start here button and local storage logic


lemilonkh (approved):
Ok, let's see... In case we need this again, we  can revert the PR I suppose.

PR #2591, opened 2026-05-21
Feat/on 5706 ask clima improvements

cephaschapa:
### Summary

- **Context-aware suggestions** — each accordion item in the GPC Wording help drawer now has its own set of 3 suggested questions. Clicking "Ask Clima AI" closes the drawer and opens the chat popover pre-loaded with topic-specific suggestions (GPC, GPC BASIC, GPC BASIC+, AR5, AR6). Falls back to the default generic suggestions when the popover is opened via the main button.

- **Drawer → popover handoff** — help drawer is now controlled; it closes before the chat popover opens to avoid stacking UI layers.

- **Z-index fix** — floating Ask Clima AI button z-index lowered to `1300` so it renders behind the drawer overlay (`1400`) instead of on top of it.

- **Input focus styling** — chat textarea now has a neutral border at rest and switches to a 2px `interactive.primary` border on focus with no browser outline.

- **i18n** — 30 new translation keys added to `onboarding.json` for per-topic AI suggestions (EN; other locales auto-translated via CI). `ask-ai` key added to all 5 locale files.

### Test plan

- [ ] Open the GPC Wording help drawer on the inventory details step
- [ ] Expand each accordion item and click "Ask Clima AI" — verify the drawer closes and the chat popover opens with the correct 3 topic-specific suggestions
- [ ] Open the chat popover via the main floating button — verify the default generic suggestions still appear
- [ ] Click inside the chat textarea — verify the border turns `interactive.primary` at 2px
- [ ] While the drawer is open, verify the floating Ask Clima AI button is behind the drawer overlay


app/src/components/ChatBot/chat-popover.tsx
```diff
@@ -150,22 +175,17 @@ export default function ChatPopover({
             borderTopRadius={4}
             p={6}
           >
-            {t("ask-ai-expert")}
+            {t("ask-clima")}
```
lemilonkh, line 182: Have you removed the previous translation string that is now unused?

app/src/components/steps/GHGI/InventoryDetailsHelpDrawer.tsx
```diff
@@ -188,10 +229,30 @@ export default function InventoryDetailsHelpDrawer({ t }: { t: TFunction }) {
...
+                            px="24px"
+                            py="16px"
+                            onClick={() => {
+                              onClose();
+                              window.dispatchEvent(
+                                new CustomEvent("open-clima-ai", {
+                                  detail: { suggestions: item.suggestions },
+                                }),
+                              );
+                            }}
```
lemilonkh, line 246: Is this the best way to communicate with the AI assistant component? I suppose the other option would be to have either a shared data store in Redux or to have a context at the page level that allows this interaction?

I suppose it's fine if it works well, just seems like we're going outside of the React world here and it's a very global approach...

app/src/components/icons.tsx
```diff
@@ -1895,3 +1895,34 @@ export const AboutPrivacyIcon = () => {
     </svg>
   );
 };
+
+export const AskAiIconOutline2 = () => {
```
lemilonkh, line 1899: I assume this icon isn't available from the react-icons package we use?

lemilonkh (approved):
Amazing work, thanks 👍 ✨

app/src/components/icons.tsx
cephaschapa: Yes

PR #2630, opened 2026-05-25
Feat/on 5753 combine screens of city selection and city context info

cephaschapa:
## Summary

Combine the "Select City" and "Confirm City Information" screens into a single step and extend the cities onboarding into a full inventory-creation wizard.

### Onboarding flow
- Replace the 2-step picker with a 4-step wizard: select-and-create city → inventory details → population → third-party data, then redirect to the new inventory's GHGI page.
- Merge the old "Confirm city information" view into `SelectCityStep`: new "GHG INVENTORY CREATION" tag, refreshed copy, and the selected city's flag, name, area, and map shown directly under the search input.
- Show a small spinner in place of the city area while the boundary query is loading.
- Disable the Continue button on every step until that step's required fields are filled.
- Switch the "some third-party sources failed to connect" toast from error (red) to warning (yellow); the inventory still completes successfully.

### Invites and routing
- Point the organization invite email URL at `/cities/onboarding` instead of `/organization/invites` so new admins land directly in the wizard.
- Update middleware to treat `/cities/onboarding?token&organizationId&...` as an invite route — unauthenticated users go to `/auth/signup` instead of `/auth/login`.
- Onboarding welcome page silently accepts the org invite on mount when token/email/organizationId are present.

### Other
- Add `UseWarningToast` hook for yellow toasts.
- Unhide the Clima AI button on the new cities onboarding pages (still hidden on `/auth/*`).

## Test plan

- [ ] New user receives org invite email; link lands on `/cities/onboarding`, prompts signup, and after signup auto-accepts the invite.
- [ ] Existing user clicking the email link is auto-authenticated and the invite is accepted silently.
- [ ] Walk through the 4-step wizard end to end; verify Continue is disabled until each step's fields are filled and verify the area spinner appears briefly on city select.
- [ ] On step 3, choose "yes" and force a third-party connection failure; toast renders yellow, inventory is still created, redirect to the inventory page works.
- [ ] Clima AI button is visible on the cities onboarding welcome and setup pages, and hidden on `/auth/login` and `/auth/signup`.


app/e2e/helpers.ts
```diff
@@ -56,64 +56,133 @@ export async function createInventory(
   return await result.json();
 }
 
-export async function createCityThroughOnboarding(page: Page): Promise<string> {
-  // Step 1: Start the city onboarding process
+/** Inventory year selected during cities onboarding E2E flows. */
+export const E2E_ONBOARDING_INVENTORY_YEAR = String(new Date().getFullYear());
```
lemilonkh: Are we sure this is going to keep working consistently when the year changes? Why does it need to be the current year?

lemilonkh (approved):
Amazing work, thanks 👍 ✨

app/e2e/helpers.ts
cephaschapa: Got catch, so this constant is only being use for e2e tests.

PR #2637, opened 2026-05-27
On 5709 agentic ccc

mircorudolph:


docs/AgenticModuleScope.md
```diff
@@ -0,0 +1,646 @@
...
+routes, internal CA token plumbing, feature-flag administration, and
+operational observability/hardening.
+- Experimental work already built as a separate proof of concept should not set
+the scope for the broader module plan.
+
+## MCP Layer
+
+> This section is transport-only. It is not the product module map.
+
+Current MCP tools:
```
mircorudolph, line 22: Let us explore if MCP is really the best way to go here.
It adds overhead to gain portability.
A direct function/ API call might be easier to implement/ debug/ stable

docs/AgenticModuleScope.md
```diff
@@ -0,0 +1,646 @@
...
+summaries
+
+The important rule is that CA should target these wrappers, not the full CC
+route catalog. A wrapper may internally call one existing service, one existing
+route, or several existing reads before returning one coherent result. If CA
+remains a separate service, the cross-service call can still go through a thin
+internal transport, but the wrapper contract is the thing the agent sees and
+the registry scopes.
+
+From CA's perspective, the LLM still calls tools. Those tools call a smaller
```
mircorudolph, line 220: Are we calling tools directly or are we calling the MCP tools?

I really like the idea of the capability wrapper. Would that wrapper expose the possible tools than can be called as a sort of registry? But the actual tool call then is via MCP? Or how is that flow?

docs/AgenticModuleScope.md
```diff
@@ -0,0 +1,646 @@
...
+set of module-owned capability wrappers, which then delegate to existing CC
+routes and services while reusing the secure token exchange already in place
+today.
+
+In the desired flow, CA should not decide tool registration manually. The
+scoped context loader should determine the active workflow step, load the
+minimal context needed for that step, resolve the matching capabilities from
+the registry, and build the agent's scoped tool set from that.
+
+```mermaid
```
mircorudolph, line 230: In this diagram it seems that `Scoped Context Loader` is calling `Capability registry`.
Is that on purpose? 
It feels our current CA service should act as an orchestrator here and make all those calls.
E.g. orchestrator calls DB service, calls scoped context, calls capabilitie registry, .....?
Like "What context and capabilities do I need for HIAP service?"

docs/AgenticModuleScope.md
```diff
@@ -0,0 +1,646 @@
...
+sequenceDiagram
```
mircorudolph, line 231: It might make sense from the flow to create first the token (after context and capabilities have resolved) but before creating the agent?
In case the user has no sufficient rights and token creation fails, we would not need to create that agent in the first place

docs/AgenticModuleScope.md
```diff
@@ -0,0 +1,646 @@
...
+
+    Agent->>CCWrapper: Call wrapper with Bearer access_token
+    CCWrapper->>CCLogic: Reuse existing CC product logic
+    CCLogic-->>CCWrapper: Raw product data/result
+    CCWrapper-->>Agent: Summarized capability payload
+    Agent-->>User: Final answer or decision-ready response
+```
+
+
+
```
mircorudolph, line 267: Good and clear flow. In this flow it seems we would not use an MCP, but rather expose the tools via the existing CC logic?

docs/AgenticModuleScope.md
```diff
@@ -0,0 +1,646 @@
...
+- input/output schema
+- required resource scope
+- confirmation behavior
+- result shape and file/draft reference shape if any
+- transport exposure: internal only, chat, MCP, public read-only
+
+This registry should become the source of truth for tool generation and scope
+selection.
+
+In the desired CA runtime, the registry should be the thing that decides which
```
mircorudolph, line 290: I really like that concept

docs/AgenticModuleScope.md
```diff
@@ -0,0 +1,646 @@
...
+  user_facing_surface: "GHGI Import"
+  active_resource: inventory
+  steps:
+    - id: open_import_review
+      functions:
+        - ghgi.import.getJob
+        - ghgi.import.getStatus
+      enabled_capabilities:
+        - ghgi.import.read
+      context:
```
mircorudolph, line 333: what is inside `context`?
E.g. in detail what would: `city: summarize name, country, project` be?
Is that the provided context to the agent? So we provide here just summarized names and so on instead of fully dumped json outputs?

docs/AgenticModuleScope.md
mircorudolph, line 1: Generally very good flow. What seems to be missing is treating production observability as a first-class citizen rather than an afterthought.

This is something we curently have different tools (e.g. logs via Grafana, LLM traces via Langsmith, ...).
I would love to have each interaction write and export a clear log artifact with inputs to the interaction, used context, ..... outputs and so on. I am thinking about using something like mlflow or similar.
We do not need to define the actual final tool but we should already incorporate the artifacts (similar to what we do in hiap-meed where we create very rich artifacts and they make life a lot easier)

docs/AgenticModuleScope.md
```diff
@@ -0,0 +1,646 @@
...
+or module data should declare:
+
+- whether version history is supported, required, or unavailable for that step
+- what snapshot or change record is created before the mutation
+- what user-facing label and summary will appear in version history
+- which fields, sectors, actions, or files are affected
+- whether restore/rollback is available after the action
+- which capability owns the restore path
+
+This matters most for GHGI imports, manual GHGI value edits, source connects,
```
mircorudolph, line 396: libs like langgraph have `checkpointers` that allow exactly that, going back to a previous LLM interaction step.
Have you thought about using frameworks vs building these in-house?

docs/AgenticModuleScope.md
```diff
@@ -0,0 +1,646 @@
...
+- user-facing progress summary
+
+This can be added to the current CA as additional workflow pieces instead of
+starting as a separate public service. For workflows that need writes, CA should
+reuse the existing secure DB-write path rather than introducing a new write
+surface.
+
+New workflow endpoints should introduce a consistent polling mechanism:
+
+- start endpoint returns a workflow/job id and initial status
```
mircorudolph, line 424: This is the current HIAP workflow. It is easy to implement however the downside is that the client (CC in this case) needs to preiodically poll the status and then the result endpoint. 
Better would be a hook that CA calls whenever the workflow is finished and then CC just calls that result endpoint once. Milan had some thoughts to this.

docs/AgenticModuleScope.md
```diff
@@ -0,0 +1,646 @@
...
+We also need an explicit database ownership split between CA and CC. The current
+documents describe runtime behavior, but the durable state model still needs to
+be made concrete.
+
+The rule should be:
+
+- CA owns conversation and agent-session state
+- CC owns durable product workflow state
+
+CA should keep:
```
mircorudolph, line 470: don't we currently already have 2 different DBs for those?

docs/AgenticModuleScope.md
```diff
@@ -0,0 +1,646 @@
...
+- The capability registry is the source of truth for those capabilities.
+- MCP is generated from that registry for the capabilities we actually want to
+expose through that transport.
+
+So the answer is not "direct functions or MCP". The durable answer is:
+
+- define real user-facing capabilities first
+- implement them as module-owned function wrappers
+- add scope chaining on top of that registry
+- generate scoped MCP exposure from the same registry
```
mircorudolph, line 577: I am not sure I fully understand the hybrid path.

Assuming we have the capability wrappers with clear tool definitions as an interface and inside the wrapper we use CC native tooling (what already exists) and for the outside we defined the tool towrads the LLM, then where does the MCP come into play and why? 
The biggest benefit would usually be the standadizayion but the capability wrapper would provide this no?

docs/AgenticModuleScope.md
piotrnowakowski: yeah you are right this is a mistake CA is orchestrating. in terms of what context i need 
But we will be calling the CC layer to do db service as this is current implementation and we dont want to rewrite it

docs/AgenticModuleScope.md
piotrnowakowski: good point i will readjust that

docs/AgenticModuleScope.md
mircorudolph: Yes the orchestrator calls the CC layer. It does not call the DB directly, but all the calls and decissions should go via current CA layer (which might need a rename to orchestrator at some point as it becomes more than just the chat interface)

docs/AgenticModuleScope.md
piotrnowakowski: we call tools directly there is now MCP in the current implementation

piotrnowakowski: Wrapper exposes possible tools per capability and we manage tools and context in the same way

docs/AgenticModuleScope.md
piotrnowakowski: Yes we expose tools to be called no mcp

docs/AgenticModuleScope.md
piotrnowakowski: good point i will ask milan

docs/AgenticModuleScope.md
piotrnowakowski: yes we have and it will stay like that

docs/AgenticModuleScope.md
piotrnowakowski: You have a clear good description but exclude the MCP 
__" we have the capability wrappers with clear tool definitions as an interface and inside the wrapper we use CC native tooling (what already exists) and for the outside we defined the tool towrads the LLM_" - that is a good definition we won't be using MCP

docs/AgenticModuleScope.md
piotrnowakowski: I will add that into our discussion points for tommorow but i am generally pretty happy with langsmith i wanted to include this as tracing for agent decision but never heard about checkpoints there yet i will read about them

docs/AgenticModuleScope.md
piotrnowakowski: I will include also this into our 1:1

docs/AgenticModuleScope.md
piotrnowakowski: Yes. Here `context` means the bounded, model-visible payload for the current Stationary Energy draft step, not a full dump of raw CityCatalyst responses.

For Stationary Energy, the agent would get a structured payload roughly like this:

```json
{
  "task": "generate_stationary_energy_draft_proposals",
  "rules": [
    {... }
    {... }
  ],
  "allowed_capabilities": [
    "ghgi.stationary_energy.load_context"
  ],
  "city": {
    "name": "Warsaw",
    "locode": "PLWAW",
    "country": "Poland",
    "region": "Mazowieckie",
    "population": 1863056
  },
  "inventory": {
    "year": 2022,
    "inventory_type": "BASIC",
    "gwp": "AR6"
  },
  "taxonomy": [
    {
      "sector_name": "Stationary Energy",
      "subsector_name": "Residential buildings",
      "scope_name": "Activity data"
    },
    {
      "sector_name": "Stationary Energy",
      "subsector_name": "Commercial and institutional buildings",
      "scope_name": "Activity data"
    }
  ],
  "current_values": [
    {
      "value": "1280000",
      "unit": "MWh",
      "emissions_value": "245000",
      "emissions_unit": "tCO2e",
    }
  ],
  "source_candidates": [
   { ... }
  ],
  "guidance_context": {
    "method_notes": [
      "Prefer more specific geography matches when year and scope also align",
      "Use guidance for methodology only, not as observed activity data"
    ],
    "unit_conventions": {
      "activity": "MWh",
      "emissions": "tCO2e"
    }
  }
}

docs/AgenticModuleScope.md
piotrnowakowski: yes it is done like that

PR #2641, opened 2026-05-29
ON-5652: Add MLflow tracking and artifact logging for hiap-meed

mircorudolph:
﻿## Summary

Adds best-effort MLflow run, metric, and artifact logging to hiap-meed, with Docker Compose for local development against a separate MLflow server. Deployed dev, test, and prod workloads use the hosted tracking URI via Kubernetes env vars. Request artifacts keep the existing run-folder layout as the default structure for both MLflow and optional local output.

## Changes

- Add `mlflow_logging` helpers for runs, params, metrics, and JSON/text artifacts (fail-open)
- Wire prioritization, exclusion preview, and explanation translation endpoints to MLflow runs
- Update `ArtifactWriter` to log default artifact paths to MLflow; rename `ARTIFACT_LOG_JSONL` to `LOCAL_ARTIFACTS_ENABLED`
- Add `compose.yaml` for local hiap-meed + MLflow
- Set `MLFLOW_*` env in k8s deployments (dev/test/prod) for hosted tracking at `https://mlflow-dev.openearth.dev`
- Document Docker, Compose, env vars, and artifact layout in README and `.env.example`
- Add unit tests for MLflow helpers and artifact path behavior

## Commits (2)

- feat: added mlflow with artifact logging
- chore: added support for remote mlflow obs


hiap-meed/k8s/deployment-test.yml
```diff
@@ -32,6 +32,16 @@ spec:
               value: "0.0.0.0"
             - name: API_PORT
               value: "8000"
+            - name: MLFLOW_ENABLED
+              value: "true"
+            - name: MLFLOW_TRACKING_URI
+              value: "https://mlflow-dev.openearth.dev"
```
piotrnowakowski, line 38: this means we send prod test and dev to dev cluster. Does the MLflow allow separation of the projects so we know where the traces are coming from ?

hiap-meed/app/utils/mlflow_logging.py
```diff
@@ -0,0 +1,217 @@
...
+    """Initialize MLflow tracking and OpenAI autologging once per process."""
+    global _INITIALIZED, _INITIALIZATION_ATTEMPTED
+
+    if not is_mlflow_enabled():
+        _INITIALIZATION_ATTEMPTED = True
+        _INITIALIZED = False
+        return False
+    if _INITIALIZED:
+        return True
+    if _INITIALIZATION_ATTEMPTED:
```
piotrnowakowski: This one-shot latch makes any transient MLflow startup/network failure permanent for the life of the process. In the documented Compose flow, or any cold start where MLflow becomes reachable a few seconds later, all later start_run() calls will silently no-op until restart; retry initialization on later requests instead of short-circuiting forever.

hiap-meed/app/utils/mlflow_logging.py
```diff
@@ -0,0 +1,217 @@
...
+@contextmanager
+def start_run(
+    *,
+    run_name: str,
+    tags: Mapping[str, object] | None = None,
+    params: Mapping[str, object] | None = None,
+    nested: bool = False,
+) -> Iterator[Any | None]:
+    """Start one MLflow run if available, otherwise yield a no-op context."""
+    if not initialize_mlflow():
```
piotrnowakowski, line 111: When mlflow.start_run() fails, the context falls back to yield None but the later log_* helpers still run because _INITIALIZED stays true. MLflow’s fluent APIs auto-create an implicit active run when no run is open, so this path can create orphaned runs and then make later explicit start_run() calls fail with “run already active”; only log when mlflow.active_run() exists or fully disable logging after run-start failure.

piotrnowakowski (approved):

hiap-meed/k8s/deployment-test.yml
mircorudolph: yes exactly. You can set filters so we can separate those

hiap-meed/app/utils/mlflow_logging.py
mircorudolph: Fixed, this now retries MLflow initialization on later requests after a 60s cooldown instead of latching failure for the worker lifetime.

hiap-meed/app/utils/mlflow_logging.py
mircorudolph: Fixed, the log helpers now no-op unless an explicit MLflow run is active, so a failed start_run() can’t create orphaned implicit runs.

PR #2643, opened 2026-06-01
Fix/on 5720 update navigation according to design specs

cephaschapa:
Changes
- Adds hover style for nav links to recipe
- Fixes navbar alignment and removes duplicate layout component structures
- Adds correct translations for module names used in the navbar


app/src/i18n/locales/de/navigation.json
```diff
@@ -5,7 +5,7 @@
   "settings": "Einstellungen",
   "account-settings": "Kontoeinstellungen",
   "account-frozen-warning-text": "Ihr Konto ist eingefroren. Bitte wenden Sie sich an einen Administrator oder an <bold>{{ email }}</bold>, um es zu reaktivieren",
-  "page-title-ghg-inventories": "GHG Inventories",
-  "page-title-hiap": "High Impact Action Plan",
-  "page-title-dashboard": "Dashboard"
+  "page-title-ghg-inventories": "Inventários de GEI",
```
lemilonkh, line 8: You have a portuguese translation in the german i18n file here

app/src/i18n/locales/en/navigation.json
```diff
@@ -6,6 +6,6 @@
   "account-settings": "Account Settings",
   "account-frozen-warning-text": "Your account is frozen. Please reach out to an admin or to <bold>{{ email }}</bold> to reactivate it",
   "page-title-ghg-inventories": "GHG Inventories",
-  "page-title-hiap": "High Impact Action Plan",
+  "page-title-hiap": "Actions & Plans",
   "page-title-dashboard": "Dashboard"
```
lemilonkh, line 10: In the other languages Dashboard changed to City Dashboard, should that happen here as well?

app/src/i18n/locales/de/navigation.json
lemilonkh: The german translation is: `Treibhausgas-Inventare`

lemilonkh (approved):
Great work, thanks 👍 ✨

PR #2644, opened 2026-06-01
On 5781 ideal agentic architecture

piotrnowakowski:
Adjustments made to the plan and ideal architecture 
- added CA draft workflow routes, services, models, migration, and tests for Stationary Energy
- added CC-to-CA bridge support for the draft flow
- kept the workflow gated behind `STATIONARY_ENERGY_AGENTIC` feature flag
- clarified that CA owns pre-commit draft state, while CC owns saving committed writes and version history (This reuses existing CC version-history writes when accepted rows are actually saved.)


climate-advisor/service/app/config/__init__.py
```diff
@@ -1,3 +1,10 @@
+from .feature_flags import FeatureFlags, get_feature_flags, has_feature_flag
```
mircorudolph: We try to avoid those relative imports.

PR #2645, opened 2026-06-02
feat(ui): ON-5758 always show 3 actions on public HIAP widget

lemilonkh:
Sometimes less than 3 actions (ranked and unranked) are selected. This makes sure there are always 3 by falling back to the highest ranked actions.


cephaschapa (approved):
👍👍👍

PR #2650, opened 2026-06-03
Feat/on 5760 third party data improvements

cephaschapa:
Changes
- Adds text component to show total emissions 
- Adds logic to block connection of data sources that already exist/filled
- Adds text component to show data source name and link
- Updates card layout anatomy and style


lemilonkh (approved):
Amazing work, thanks 👍 ✨

PR #2659, opened 2026-06-04
feat/ON-5768-add-manually-added-data-note

cephaschapa:
Changes
- Adds a note to indicate manually added data in the activity accordion
- Appends the same text to the inventory progress labe for manully added data


lemilonkh (approved):
Great, thanks 👍

PR #2665, opened 2026-06-05
Fix S3-backed inventory import validation and approve flow

cephaschapa:
## Summary

Fixes inventory import validation and approval when uploaded files are stored in S3 instead of PostgreSQL BYTEA. Column mapping now loads correctly after upload, and near-ecrf imports approve from pre-extracted rows instead of failing with "File data not found".

## Changes

- Add `resolveImportedFileBuffer` to load import file content from S3 or BYTEA
- Use resolved buffers in validation and approve API routes
- Fall back to stored column detection when file parse fails on validation
- Include near-ecrf adapter in the extracted-rows approval path
- Document S3-related env vars in `env.example`

## Test plan

- [ ] Upload an ECRF/near-ecrf file with S3 storage enabled and confirm column mappings appear on the validation step
- [ ] Complete import approval and verify inventory data is created without "File data not found"
- [ ] Confirm local dev still works when S3 is not configured (BYTEA fallback)

## Commits (1)

- fix: load S3-backed import files for validation and approve steps

### DIFF
--- a/app/e2e/dashboard.spec.ts
+++ b/app/e2e/dashboard.spec.ts
@@ -34,10 +34,11 @@ test.describe("Dashboard", () => {
       );
       await expect(addNewInventoryButton).toBeVisible({ timeout: 10000 });
 
-      // Filter by the expected text to handle multiple elements with the same test ID
+      // Multiple inventories can share the same year; assert the year is shown.
       const inventoryYearValue = page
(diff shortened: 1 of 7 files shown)


lemilonkh (approved):
Great work, thanks 👍✨

PR #2678, opened 2026-06-11
feat(ui): ON-5837 connect button state on SourceDrawer

lemilonkh:
- Shows Go to source button instead of Connect source on SourceDrawer in case the source is already connected
- Make separators on SourceDrawer grey instead of green
- Remove unused state and variables on data step page (add data page)


cephaschapa (approved):
Great work 👍

PR #2683, opened 2026-06-15
fix(db): remove duplicate modules for GHGI and HIAP

lemilonkh:
Also moves correct translation data from migration into seeder (where the other modules are defined as well)


cephaschapa (approved):
Nice 👍👍👍

PR #2684, opened 2026-06-15
refactor(ui): remove old (pre-JN) onboarding page

lemilonkh:
Also remove associated components that were only used on the old onboarding pages, and changes all links to point to the new onboarding page (/cities/onboarding/setup) instead of the old one (/onboarding/setup).


cephaschapa (approved):
🙌🙌🙌

PR #2686, opened 2026-06-15
fix(ui): ON-5806 help sidebar cut off on JN onboarding flow

lemilonkh:
Removes zIndex property on bottom bar of JN onboarding flow to match GHGI onboaring flow


cephaschapa (approved):
Thanks 👍

PR #2687, opened 2026-06-16
Adds a consistent upload existing inventory path for GHGI and city onboarding.

cephaschapa:
Summary
Adds a consistent upload existing inventory path for GHGI and city onboarding. Upload mode (?mode=upload) skips the third-party data step and sends users to the import wizard after inventory creation.

Changes
- GHGI setup: hide third-party step when mode=upload
- City onboarding landing: add Upload existing inventory button; preserve project param to setup
- City onboarding setup: upload mode skips third-party and redirects to import
- JN Drawer: “Add new city” goes to /cities/onboarding?project={id} instead of setup


lemilonkh (approved):
Great work, thanks 👍🏻✨

PR #2688, opened 2026-06-16
fix(modules): adds translations to module taglines and descriptions

cephaschapa:
Summary
Adds missing locale strings for POC module cards so titles, taglines, and descriptions display in the user's language instead of falling back to English.

Changes
- Added de, es, fr, and pt translations in app/seed-data/modules/modules.json for 8 English-only POC modules (GHG Inventory Comparison, Prisma CCRA, Flourish suite, NBS/Solar builders, Geospatial AI Risk Map)

### DIFF
--- a/app/seed-data/modules/modules.json
+++ b/app/seed-data/modules/modules.json
@@ -121,13 +121,25 @@
     "type": "POC",
     "stage": "implement",
     "name": {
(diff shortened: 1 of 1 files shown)


lemilonkh (approved):
Great, thanks 👍🏻✨

PR #2690, opened 2026-06-16
docs(hiap-meed): methodology variants & migration analysis

MartinWainstein:
## What

Adds a single shareable reference for the team behind the HIAP methodology: [`hiap-meed/docs/methodology-variants-and-migration.md`](../blob/docs/hiap-methodology-variants/hiap-meed/docs/methodology-variants-and-migration.md).

It compares the three live variants of the prioritization methodology — **as-built**, the proposed **Chile (MEED+ / Option B / TEF)** direction, and the proposed **Brazil (City Climate Compass, Phase 3)** direction — at both the conceptual and technical-architecture levels.

## Contents

- **Current-state diagnostic** of the distributed-documentation problem — knowledge scattered across repo / Notion / Drive / partner-owned docs, naming drift (HIAP v3 = MEED+ = Mitigation Accelerator; the DOC-A…H register that only the CORFO Compendium understands), doc–code divergence (e.g. alignment `0.80` in docs vs `0.75` in `scoring_config.py`), and distributed methodology authority — plus the **kernel + adapter** corrective direction.
- **Conceptual methodology**: three variant diagrams + a divergence table (the two rows that matter: *unit ranked* and *where the I/F/A pillars attach*).
- **Technical architecture**: the as-built call graph (read from `orchestrator.py` + `api.py`) and the two proposed migration architectures, with a per-layer **migration-delta table**.

## Notes for reviewers

- All diagrams are **inline Mermaid** and render directly in the GitHub diff/preview.
- The **as-built diagram is exact** (from code); the **migration diagrams are proposed target architectures** — new module names (`run_prioritization_te`, `impact_shift`, `shifts_api`, …) are suggested, not existing.
- Intended as the seed of an OEF-owned **methodology kernel spec**; please comment inline on the open *"where the pillars attach"* decision.

### DIFF
--- a/hiap-meed/docs/methodology-variants-and-migration.md
+++ b/hiap-meed/docs/methodology-variants-and-migration.md
@@ -1,0 +1,307 @@
+# HIAP / MEED+ — Methodology Variants & Technical Migration
+
+> **Status:** Draft for team review
+> **Prepared by:** Martin Wainstein · **Owner:** TBD (proposed: OEF Methodology Owner)
+> **Last updated:** 2026-06-16
+> **Source of truth:** `hiap-meed` service (as-built diagrams generated from `orchestrator.py` + `api.py`)
+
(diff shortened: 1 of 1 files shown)


lemilonkh (approved):
Great, thanks for the documentation! ✨

PR #2696, opened 2026-06-17
ON-5833: Align inventory fallback redirects to city GHGI routes

cephaschapa:
## Summary

Align inventory fallback navigation with city-scoped GHGI routes to reduce reliance on legacy `/${lng}/${inventoryId}` URLs.

## Changes

- Updated inventory validation redirects to prefer `/${lng}/cities/${cityId}/GHGI/${inventoryId}`
- Updated GHGI project drawer city navigation to use city-scoped inventory URLs
- Updated missing-inventory flows and CTA redirects to use city-scoped defaults when available
- Kept legacy inventory URL fallback only when `defaultCityId` is missing

## Test plan

- Open GHGI project drawer and navigate to a city inventory
- Trigger missing-inventory route and verify redirect behavior
- Verify fallback still works when only `defaultInventoryId` exists


lemilonkh (approved):
Great, thanks 🙌 ✨

PR #2698, opened 2026-06-17
feat: ON-5816 don't add notation keys during onboarding

lemilonkh:
- Don't add notation key entries for missing GPC reference numbers during onboarding connect-all route
- Remove error and warning messages during onboarding when data sources fail to connect
- Remove translation keys for these messages


cephaschapa (approved):
Thanks 👍✨

PR #2700, opened 2026-06-18
Fix Ask Clima thread init for users without an inventory

joaquinOEF:
## Summary

A user with no inventory yet (e.g. on the onboarding screen) couldn't open **Ask Clima**: the frontend sent `inventory_id` as `""`, and the `POST /api/v1/chat/threads` Zod schema rejected it (`z.string().uuid().optional()` allows `undefined`, not `""`), returning a 400 and showing the "Failed to initialize thread" / "Failed to send message" toasts. The Climate Advisor service itself already tolerates a thread with no inventory — only the app's front-door validation rejected it.

## Changes

- **chat-popover**: coalesce a missing inventory to `undefined` instead of `""`.
- Make `inventoryId` optional through `ChatBot → useChat → ChatService`, coercing only the legacy assistant path that still requires a string (behavior preserved).
- Move thread-create validation to `util/validation.ts` (`createChatThreadRequest`) and preprocess `"" → undefined` so an inventory-less thread is created; any caller sending `""` is now safe.
- Add the missing `an-error-occurred` key to the **en** `chat` namespace so the toast title renders text instead of the raw key.
- Extend `tests/api/chat-routes.jest.ts`: empty/omitted `inventory_id` succeed; a malformed non-empty value still 400s.

## Verification

- `npm run jest` (chat-routes): 10/10 pass · `tsc --noEmit` clean · ESLint/Prettier clean.
- End-to-end against a local app + live Climate Advisor: empty `inventory_id` → 200 (CA `POST /v1/threads 201`), and a message streams a real response on the inventory-less thread; malformed → 400.


lemilonkh (approved):
Great, thanks 👍🏻✨

PR #2701, opened 2026-06-18
Fix Ask Clima bubble alignment and hide suggestions scrollbar

joaquinOEF:
## Summary

Two small UI fixes for the Ask Clima chat panel.

## Changes

- **Bubble alignment**: assistant replies were pushed to the right because a `Spacer` was rendered before every bubble regardless of role. Now the `Spacer` sits before the bubble only for user messages and after it for assistant messages, so assistant bubbles align left next to the avatar (user bubbles stay right).
- **Suggestions scrollbar**: hide the horizontal scrollbar on the suggested-prompt row (`ChatSuggestions`) while keeping it scrollable.

## Verification

- ESLint + Prettier clean · `tsc --noEmit` clean.
- Visual: the assistant greeting and streamed replies render left-aligned; the suggestion chip row scrolls with no visible scrollbar.

### DIFF
--- a/app/src/components/ChatBot/ChatSuggestions.tsx
+++ b/app/src/components/ChatBot/ChatSuggestions.tsx
@@ -18,7 +18,21 @@ export function ChatSuggestions({
   disabled = false,
 }: ChatSuggestionsProps) {
   return (
-    <Box display="flex" flexDir="row" gap={2} whiteSpace="nowrap" pb="3" overflowX="auto" flexShrink={0}>
+    <Box
+      display="flex"
+      flexDir="row"
+      gap={2}
+      whiteSpace="nowrap"
+      pb="3"
+      overflowX="auto"
+      flexShrink={0}
+      // Keep horizontal scrolling but hide the scrollbar across browsers.
+      css={{
+        scrollbarWidth: "none",
+        msOverflowStyle: "none",
+        "&::-webkit-scrollbar": { display: "none" },
+      }}
+    >
       {suggestions.map((suggestion, i) => (
         <Button
           key={i}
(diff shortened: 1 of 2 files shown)


app/src/components/ChatBot/ChatSuggestions.tsx
```diff
@@ -18,7 +18,21 @@ export function ChatSuggestions({
...
+      whiteSpace="nowrap"
+      pb="3"
+      overflowX="auto"
+      flexShrink={0}
+      // Keep horizontal scrolling but hide the scrollbar across browsers.
+      css={{
+        scrollbarWidth: "none",
+        msOverflowStyle: "none",
+        "&::-webkit-scrollbar": { display: "none" },
+      }}
```
lemilonkh, line 34: Are we sure that we don't want to have a vertical scroll bar here at least to give the user feedback that they can go back to previous messages?
We can disable horizontal scrolling if that's the issue.

lemilonkh (approved):
Thanks 👍🏻✨

app/src/components/ChatBot/ChatSuggestions.tsx
joaquinOEF: Good catch raising it
just to clarify what this change actually touches: this ChatSuggestions box is only the horizontal row of suggested-prompt chips, so it only ever scrolls sideways (single row, whiteSpace: nowrap). hiding the bar here keeps the chips swipeable but tidies up the look ‚ we're not disabling the scroll, just the visible bar.

the vertical scrollbar for going back to previous messages lives in a separate component (ChatMessageList, overflowY: auto) which this PR doesn't touch, so that feedback is still there. you can see it in the screenshot below: vertical bar on the right with the messages stacked, and the chip row underneath scrolling sideways with no bar.

Screenshot 2026-06-30 at 9 37 14 AM

let me know if there is something i'm missing

PR #2702, opened 2026-06-18
On 5827 full chat experience

piotrnowakowski:
1. Reworked Stationary Energy into a full chat experience instead of previous view 

2. Added source review panels, staged decisions, bulk accept, change, rollback, and confirmation flows as a ,,tools" for agent to make decisions

3. Separated “save draft in Clima” from “save rows to CityCatalyst inventory,” with explicit inventory confirmation.

4. Added Climate Advisor review tools and backend services for listing, staging, validating, saving, and rolling back review choices.

5. Added database persistence for staged Stationary Energy review selections.

6. Expanded tests and cleaned up prompts, docs, naming, formatting, and review-flow helper logic.


climate-advisor/prompts/data_analysis.md
```diff
@@ -1,46 +0,0 @@
-<!--
-NOTE: This prompt is currently configured in llm_config.yaml but not loaded in runtime code.
```
piotrnowakowski, line 2: it was not used at any moment residue dead code so i throw it out

climate-advisor/prompts/stationary_energy_draft_generation.md
```diff
@@ -1,86 +0,0 @@
-<role>
```
piotrnowakowski, line 1: we do it with CC now so now LLM included

climate-advisor/prompts/tools/default_tool_arguments.md
mircorudolph, line 1: Good shape. 
Maybe even clearer for an LLM giving an example like:

- `get_user_inventories`: no arguments.
- `city_inventory_search`: `{"city_name": "New York"}` or `{"city_name": "Paris", "year": 2023}`
  - `city_name` (string, required)
  - `year` (integer, optional; omit when not specified)
- `get_inventory`: `{"inventory_id": "<id-from-prior-tool-output>"}`
  - `inventory_id` (string, required)
- `get_all_datasources`: `{"inventory_id": "<id-from-prior-tool-output>"}`
  - `inventory_id` (string, required)
- `climate_vector_search`: `{"question": "What is scope 2 emissions under GPC?"}`
  - `question` (string, required)

Not strictly needed but might be even clearer for the LLM.

climate-advisor/service/app/services/stationary_energy/stationary_energy_draft_service.py
mircorudolph, line 597: review_draft() persists durable review_decisions but never clears or marks the corresponding staged_review_selections as saved. After the next refresh, buildInitialDecisionState() in app/src/components/StationaryEnergyDraft/flow-review.ts still prefers those lingering staged rows first, so a successfully saved draft or inventory save can reopen with stale agent choices and later re-submit them; mark staged selections inactive/saved when this route accepts a full review payload.

climate-advisor/docs/architecture.md
```diff
@@ -48,509 +55,248 @@ sequenceDiagram
...
+  - `stationary_energy_accept_multiple`
+  - `stationary_energy_accept_all_recommended`
+  - `stationary_energy_request_bulk_review_confirmation`
+  - `stationary_energy_request_all_recommended_confirmation`
+  - `stationary_energy_request_staged_source_change_confirmation`
+  - `stationary_energy_request_staged_sources_rollback_confirmation`
+  - `stationary_energy_rollback_staged_sources`
+  - `stationary_energy_save_review_draft`
+  - `stationary_energy_request_inventory_save_confirmation`
+  - `stationary_energy_review` prompt block
```
mircorudolph: This syntax looks weird and I cannot find neither stationary_energy_review` tool or stationary_energy_review_prompt_block

is this an error?

climate-advisor/prompts/stationary_energy_review.md
```diff
@@ -0,0 +1,132 @@
+<role>
+You are Climate Advisor assisting with an active GPC Stationary Energy draft review.
```
mircorudolph: I think we call it actually Clima? 

Maybe lets change this in all the prompts so we are not cunfusing poor Clima with different personalities hahaha

climate-advisor/prompts/stationary_energy_review.md
```diff
@@ -0,0 +1,132 @@
...
+</role>
+
+<task>
+Help the user review, choose, stage, and save Stationary Energy draft source selections using the active draft context and the scoped Stationary Energy review tools.
+
+Follow these rules:
+- Prefer tool calls over free-text inference for active Stationary Energy draft review requests.
+- Use only proposal ids, source ids, candidate ids, rows, values, and units already present in the draft context or returned by `stationary_energy_list_review_options`.
```
mircorudolph: Consider replacing generic terms (“proposal ids”, “source ids”, “values”, “units”) with the exact JSON field names in backticks — e.g. `proposal_id`, `candidate_id`, `datasource_id` / `selected_source_id`, and the evidence fields from <output>. “Source ids” is ambiguous (datasource_id vs selected_source_id), and matching the real keys should make tool calls more reliable.

If you agree, please also check other prompts for similar patterns. These small changes should make LLM tool calls more resilient

climate-advisor/prompts/stationary_energy_review.md
mircorudolph, line 1: This is a very long prompt. Can we somehow make the task here more atomic?
It seems to work but I am worried that it might occasionally behave weirdly because of so many instructions.

mircorudolph (approved):
Awesome PR. Very clear and good structure!

app/src/components/Cards/SectorCard.tsx
```diff
@@ -1,33 +1,34 @@
 "use client";
-import SubSectorCard from "@/components/Cards/SubSectorCard";
-import { InventoryResponse, SectorProgress } from "@/util/types";
-import { Box, Heading, Icon, Text, SimpleGrid } from "@chakra-ui/react";
-import { Button } from "@/components/ui/button";
-import NextLink from "next/link";
```
lemilonkh, line 6: Not sure why this file was reworked and renamed, it mostly seems like re-ordering and reformatting? Which is fine, but the previous PascalCase name was fine for component files.

app/src/components/StationaryEnergyDraft/stationary-energy-chat-artifact-page.tsx
```diff
@@ -0,0 +1,352 @@
...
+        xl: desktopViewportHeight
+          ? `${desktopViewportHeight}px`
+          : "calc(100dvh - 88px)",
+      }}
+      h={{
+        base: "auto",
+        xl: desktopViewportHeight
+          ? `${desktopViewportHeight}px`
+          : "calc(100dvh - 88px)",
+      }}
```
lemilonkh, line 86: Is this necessary? Is it doing anything major over `h="100%" minH="100%"`. If we don't need it, the use effect with the event listeners above would also not be required.

app/src/components/StationaryEnergyDraft/stationary-energy-chat-primitives.tsx
```diff
@@ -37,14 +42,183 @@ function useStationaryEnergyAgenticTranslation() {
   return useTranslation(lng, "stationary-energy-agentic");
 }
 
+const agentMarkdownComponents: Components = {
```
lemilonkh: We also have markdown components defined in `app/src/components/ChatBot/ChatMessageList.tsx`. Would it make sense to unify them or are they best kept separate?

lemilonkh (approved):
Great, thanks 👍 ✨ 
Would be good to reset the components/Cards/SectorCard.tsx file changes, as it's being used in other files and has been renamed (but not the usages it seems).

climate-advisor/service/app/services/stationary_energy/stationary_energy_draft_service.py
piotrnowakowski: Fixed in `e654a8d87`. `review_draft()` now marks active staged selections as saved immediately after durable review decisions are persisted, and I removed the duplicate cleanup from the agent-save wrapper. Added `test_review_marks_staged_agent_choices_saved` to cover the refresh/status behavior.

climate-advisor/docs/architecture.md
piotrnowakowski: Yes, good catch. It was documentation wording, not a runtime tool. I updated the architecture doc to separate registered tools from prompt blocks, and now document `stationary_energy_review` as an appended prompt block loaded by `AgentService.create_agent()`.

climate-advisor/prompts/stationary_energy_review.md
piotrnowakowski: Updated. The active prompt persona now says Clima in `default.md`, `inventory_context.md`, and `stationary_energy_review.md`. I left service/docs references as “Climate Advisor” where they refer to the backend service name rather than the assistant persona.

climate-advisor/prompts/stationary_energy_review.md
piotrnowakowski: Agreed and updated. The Stationary Energy prompt now uses exact keys like `proposal_id`, `candidate_id`, `selected_source_id`, `selected_candidate_id`, `activity_value`, `activity_unit`, `emissions_value`, and `emissions_unit`. I also added prompt-rendering assertions for the important field names.

climate-advisor/prompts/stationary_energy_review.md
piotrnowakowski: Reworked the `<task>` section into an ordered intent route table: inspect/explain, stage one focused choice, request bulk confirmation, apply confirmed bulk choices, change/rollback staged choices, apply confirmed rollback, and save. The tool contracts remain in `<output>`, so routing and JSON argument shapes are now separated.

piotrnowakowski:
Thanks, agreed. I reset the `SectorCard` change back to the existing PascalCase path and removed the lowercase `sector-card.tsx` usage. The shared `SectorCard.tsx` file is no longer part of this feature change.

app/src/components/Cards/SectorCard.tsx
piotrnowakowski: Agreed. This was unnecessary churn, so I restored the previous `SectorCard.tsx` filename and removed the rework from this PR.

app/src/components/StationaryEnergyDraft/stationary-energy-chat-artifact-page.tsx
piotrnowakowski: Good point. I removed the runtime height measurement, `surfaceRef`, and resize listener. The page now uses static responsive height props, so we avoid the extra effect and DOM measurement.

app/src/components/StationaryEnergyDraft/stationary-energy-chat-primitives.tsx
piotrnowakowski: Yes, I unified the shared markdown element mapping into a reusable helper while keeping separate style options for the regular chatbot and Stationary Energy chat. That removes the duplication without forcing both surfaces to look identical.

climate-advisor/prompts/tools/default_tool_arguments.md
piotrnowakowski: Good suggestion. I added concrete JSON examples for the default tool argument contracts: no-argument inventory listing, city search with and without `year`, inventory/data-source lookups using the prior tool’s `inventory_id`, and `climate_vector_search` with a `question`.

PR #2714, opened 2026-06-23
ON-5835: Stabilize flaky Playwright CI tests

cephaschapa:
## Summary

Fixes flaky Playwright CI failures in global teardown and the report-results dashboard assertion.

## Changes

- Add \scripts/delete-e2e-test-admin.ts\ and use it from global setup/teardown instead of fragile inline \	sx -e\ DB scripts
- Run \
eport-results.spec.ts\ as a serial flow with shared city/inventory (\beforeAll\)
- Require methodology cards before scope 1/2 data entry; skip only when activity already exists


app/e2e/global-setup.ts
```diff
@@ -6,57 +6,55 @@ const execAsync = promisify(exec);
...
-        }
-        console.log('Database initialized, cleaning up user...');
-        await db.models.User.destroy({ where: { email: '${TEST_ADMIN_EMAIL}' } });
-        console.log('Cleaned up existing test admin user');
-        await db.sequelize?.close();
-        console.log('Database connection closed');
-      })()"
-    `, { timeout: 30000 });
+    const { stdout: deleteOutput } = await execAsync(
+      "npx tsx scripts/delete-e2e-test-admin.ts",
```
lemilonkh, line 20: Good to refactor this from a string to a text file, however I wonder why this can't just be executed inside of the current file? Instead of spawning a new process.
But we can think about this in a future refactoring, good enough for now imo.

lemilonkh (approved):
Amazing! Glad to have these reliability issues figured out hopefully 🙌✨

PR #2715, opened 2026-06-23
Fix navbar links to city-scoped routes

cephaschapa:
## Summary

Fix navbar links to use city-scoped routes instead of legacy inventory URLs.


lemilonkh (approved):
Good spot, thanks 👍

PR #2733, opened 2026-06-24
ON-5857: Align navbar language switch with main

cephaschapa:
Keep locale in sync across routes by using URL activeLng, awaiting i18n updates, and refreshing after navigation.

### DIFF
--- a/app/src/components/navigation-bar.tsx
+++ b/app/src/components/navigation-bar.tsx
@@ -29,7 +29,7 @@ import {
   MdOutlineMenu,
 } from "react-icons/md";
 import Cookies from "js-cookie";
-import { useRouter } from "next/navigation";
+import { useParams, useRouter } from "next/navigation";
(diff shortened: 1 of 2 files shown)


app/src/components/navigation-bar.tsx
```diff
@@ -74,7 +75,9 @@ export function NavigationBar({
   restrictAccess?: boolean;
   isOrgOwner?: boolean;
 }) {
-  const { t } = useTranslation(lng, "navigation");
```
lemilonkh, line 77: Can we remove the lng property of this component then?

lemilonkh (approved):
Great, thanks 👍🏻✨

app/src/components/navigation-bar.tsx
cephaschapa: Yes!

PR #2734, opened 2026-06-25
On 5883 whole inventory tools

piotrnowakowski:
Added tools to show the agentiic flow agent some full inventory data


climate-advisor/docs/architecture.md
```diff
@@ -172,7 +193,9 @@ workflow state in PostgreSQL.
   - Uses `stationary_energy_review` as the full prompt for active Stationary
     Energy review chat.
   - Keeps general inventory and vector-search tools out of active review chat.
```
mircorudolph, line 205: This is a question we need to figure out.
E.g. what is the "Basic Clima functionality" that should ALWAYS be present?

If a user is doing its GHGI emissions and then asks, "by the way, what was GPC", and then Clima says "sorry I cannot look up the vector store right now" that feels like a bad UX.

Therefore basic clima functionality should always be present in all the agents.

climate-advisor/docs/architecture.md
```diff
@@ -197,7 +220,10 @@ workflow state in PostgreSQL.
 - `tools/climate_vector_sync.py`
   - General climate knowledge retrieval.
 - `tools/cc_inventory_wrappers.py`
-  - The CityCatalyst inventory tool pack used by the general prompt.
+  - The temporary legacy datasource wrapper used by the general prompt.
```
mircorudolph, line 233: why temporary? Why don't we fix this?

climate-advisor/prompts/tools/default_tool_policy.md
```diff
@@ -1,18 +1,21 @@
...
 
-- `get_inventory`
-  - Use after an inventory is selected to fetch detailed inventory metadata.
+- `inventory_emissions_context`
+  - Use after an inventory is selected to summarize total emissions, sector shares, top emitters, and source mix.
+  - Pass the selected internal `city_id` and `inventory_id` from `inventory_list_accessible`.
 
 - `get_all_datasources`
-  - Use only after an inventory is identified to summarize available successful data sources.
+  - Temporary legacy datasource tool. Use only after an inventory is identified to summarize available successful data sources.
```
mircorudolph, line 21: Again, why keep a temporaray legacy tool and not switch over to the new tool?

climate-advisor/prompts/default.md
mircorudolph, line 1: In my mind, this is the standard prompt that ALL agents should have. All should be a child of the main Clima.
otherwise e.g. none of the other agents would know that they are Clima, what their general task is and so on.

climate-advisor/prompts/stationary_energy_review.md
mircorudolph, line 1: This is the real question here. 
Should the specialized agent not have all the core tools? If the user askes for GPC data, and couldn't call the vector store, that would be an odd user experience

climate-advisor/service/app/tools/payload_trimmers.py
mircorudolph, line 1: How are we doing this trimming now? E.g. how do we prevent bloating of context?

climate-advisor/README.md
```diff
@@ -11,7 +11,8 @@ conversational experience for CityCatalyst (CC). The service lives under
 - **Vector Search**: Semantic search over climate knowledge base using pgvector
 - **Tool Integration**:
   - General chat tools: climate knowledge search plus CityCatalyst inventory
-    tools (`get_user_inventories`, `city_inventory_search`, `get_inventory`,
+    tools (`inventory_list_accessible`, `inventory_status_overview`,
+    `inventory_emissions_context`, and temporary legacy
```
mircorudolph, line 15: temp tools?

climate-advisor/prompts/default.md
```diff
@@ -37,11 +37,12 @@ Tool invocation argument contracts:
 
 Output behavior rules:
 
-- Follow this inventory flow: identify inventory (`get_user_inventories` or `city_inventory_search`) -> confirm selected inventory -> `get_inventory` -> optionally `get_all_datasources`.
+- Follow this inventory flow: identify inventory (`inventory_list_accessible`) -> confirm selected city/year -> `inventory_status_overview` and/or `inventory_emissions_context` -> optionally `get_all_datasources`.
```
mircorudolph: The new flow tells the assistant to confirm only city/year before calling inventory_status_overview or inventory_emissions_context, but inventory_list_accessible can return multiple inventories for the same city/year that differ by inventory_name, type, or gwp. That makes it easy for the model to pick the wrong inventory_id and summarize the wrong inventory data; the prompt should require disambiguation on those fields whenever more than one inventory matches.

climate-advisor/service/app/services/citycatalyst_client.py
mircorudolph, line 403: post_internal_capability() posts the new inventory capability routes directly and never goes through the existing 401 refresh path in post_with_auto_refresh(). That means inventory_list_accessible, inventory_status_overview, and inventory_emissions_context now fail as soon as the CC bearer token expires mid-thread, which is a real regression from the old inventory tooling; route these calls through the same refresh-aware path or pass enough user/thread context to refresh and persist the token on retry

climate-advisor/prompts/chat.md
```diff
@@ -0,0 +1,47 @@
...
+</tools>
+
+<output>
+Return either:
+1) a normal assistant response in plain text, or
+2) a tool invocation using one available general-chat tool and valid arguments.
+
+General chat output rules:
+- Exact tool argument contracts come from the registered runtime tool definitions and are not duplicated here.
+- For inventory exploration, follow this flow: `inventory_list_accessible` -> confirm selected city/year -> `inventory_status_overview` and/or `inventory_emissions_context` -> optionally `get_all_datasources`.
```
mircorudolph: The general-chat contract still tells the model to confirm only city/year before calling inventory_status_overview or inventory_emissions_context, even though inventory_list_accessible can return multiple same-city same-year inventories and already exposes inventory_name, type, and gwp as the disambiguators. That can still select the wrong inventory_id and summarize the wrong inventory; require disambiguation on those extra fields whenever city/year is not unique.

app/src/components/StationaryEnergyDraft/stationary-energy-chat-artifact-panels.tsx
```diff
@@ -329,6 +332,34 @@ export function ClimaChatPanel({ actions, state }: ClimaChatPanelProps) {
     state.loadingAction,
   ]);
 
+  useEffect(() => {
```
mircorudolph, line 382: Not a frontend guy so just posting this as-is:

This effect still forces focus back into the chat composer whenever a non-text widget appears, regardless of what the user is currently interacting with. When a confirmation/save card renders, keyboard focus gets stolen from the new actionable controls, so Enter can send chat text instead of confirming the review action; only restore composer focus if it previously owned focus, or move focus into the newly rendered card instead.

mircorudolph (approved):
Great, thanks!

app/src/components/StationaryEnergyDraft/stationary-energy-chat-artifact-panels.tsx
piotrnowakowski: That is an intended change that Carlos wanted, so we focus mostly on chat even when widgets arrive. The user has to switch out of chat to interact with them.

climate-advisor/service/app/services/citycatalyst_client.py
piotrnowakowski: Fixed. Internal inventory capability calls now refresh and retry on expired/401 CC tokens, update the shared tool token, and persist the refreshed token for later calls. Added regression coverage for the retry path.

climate-advisor/prompts/chat.md
piotrnowakowski: Fixed. The general chat prompt now requires disambiguation by `inventory_name`, `type`, and `gwp` whenever city/year is not unique before calling status or emissions tools. Added prompt coverage for this contract.

app/src/app/api/v1/internal/ca/capabilities/ghgi/inventory/list-accessible/route.ts
```diff
@@ -0,0 +1,86 @@
...
+} from "@/backend/agentic/ghgi/stationary-energy/auth";
+import { apiHandler } from "@/util/api";
+
+export const POST = apiHandler(async (req, { session }) => {
+  requireStationaryEnergyAgenticEnabled();
+  requireClimateAdvisorServiceRequest(req);
+
+  const body = inventoryListAccessibleInputSchema.parse(await req.json());
+  const userId = session?.user?.id;
+  if (!userId) {
```
cephaschapa, line 78: We may need to check if user has access to this inventory using the permissions service.

app/src/components/StationaryEnergyDraft/use-stationary-energy-chat-artifact-controller.ts
```diff
@@ -995,15 +1038,23 @@ export function useStationaryEnergyChatArtifactController(
...
-          );
-        });
+        pendingDraftStatusRefreshCountRef.current += 1;
+        void refreshDraftStatusSilently(toolDraftRunId)
+          .catch((error) => {
+            setErrorMessage(
+              resolveErrorMessage(
+                t,
+                error,
+                "error-failed-to-load-stationary-energy-draft-status",
```
cephaschapa, line 1117: let make sure we add this translation string to the translation file here: app\src\i18n\locales

cephaschapa (approved):

app/src/app/api/v1/internal/ca/capabilities/ghgi/inventory/list-accessible/route.ts
piotrnowakowski: Fixed. The list-accessible route now filters every returned inventory through PermissionService.canAccessInventory and drops inventories/cities the user cannot access. Added regression coverage for the filter path.

app/src/components/StationaryEnergyDraft/use-stationary-energy-chat-artifact-controller.ts
piotrnowakowski: Fixed. Added the missing error-failed-to-load-stationary-energy-draft-status key to the stationary-energy-agentic locale files under app/src/i18n/locales.

PR #2738, opened 2026-06-26
Stationary Energy: agentic chat UX overhaul + start-draft from chat

carlosgraffi:
## Summary

Reworks the Stationary Energy agentic draft into a modern agentic chat: fixes five reported UX issues, adds a proper Clima welcome, and — the substantive change — lets the chat agent **start a draft from natural language** (previously a button-only action). The full technical breakdown lives in `docs/stationary-energy-agentic-chat-ux.md` on this branch.

## Changes

- **Composer keeps focus** after each send.
- **Fresh / new drafts open to a Clima welcome empty state** and generate on demand — no auto "Coverage check" panel or auto-staged review cards.
- **New left "Drafts" sessions panel**; the right panel is now a **static inventory overview** that no longer mirrors the hovered row.
- **"Recommended" source shown as a hint** (no pre-filled radio); selecting a source is an explicit click with a distinct staged state.
- **New Climate Advisor `stationary_energy_start_draft` tool** + prompt/wiring so "draft the empty rows" (and compare / stage / save) work by chat.
- **Streaming-chat polish** — suggestion chips no longer overlap the in-chat review widgets, the composer auto-grows (Enter sends / Shift+Enter newline) then caps, and auto-scroll follows reader intent (live-edge only; pauses on scroll-up or text selection).

## Next steps (backend limitations left out of this UI-focused rework)

- **Read-only whole-inventory context panel (quick win).** The full-inventory overview already exists as read-only data — this PR's `inventory_status_overview` / `inventory_emissions_context` tools (completion, per-sector data state, total/by-sector emissions, top emitters) — but it is **chat-only with no UI**. Surface it as a **view-only panel** so the user can see the current sector's share of total inventory emissions and overall completion as a reference. Read-only: needs a browser-facing route (or reuse `inventory/{id}/progress`) plus the panel — no write path involved.
- **Draft/session CRUD** — drafts can be created, listed, switched, and started, but **not deleted or renamed**. Add `DELETE` (and a title `PATCH`) on `/api/v1/stationary-energy-drafts/{id}` so the sidebar can manage sessions like real chat history.
- **User-named sessions** — drafts are auto-labeled by status + timestamp; there's no way to title a session. This needs a name field on the draft run before the sidebar can show custom titles.
- **Writing beyond Stationary Energy** — drafting/writing is **Stationary-Energy-only** today; every other sector is read-only. Extending it means a per-sector draft pipeline (context loader, proposal/value builder, save-to-inventory path, agent tools + prompt, FE wiring) — ideally by generalizing the existing SE pipeline. This is a major effort and a roadmap item, not a quick follow-up.
- **Proactive agent tuning** — bulk-apply and save still route through confirmation cards by design; tuning the prompt to act more directly on clear natural-language requests is a follow-up.
- **Draft-run lifecycle / cleanup** — starting a new draft does not cancel or clean up a prior in-flight or abandoned generation, so stale runs accumulate (tied to the missing delete).
- **Capability-call token refresh** — the whole-inventory read tools can fail with an expired CityCatalyst session token on long-lived sessions; the refresh path for capability calls should be hardened.
- **(FE polish)** — on a brand-new draft the overview briefly shows the sub-sector skeleton as "queued / 0 of N" before generation; it should read as a true empty overview.

## Notes

- **Stacked on `ON-5883-whole-inventory-tools`** (the base of this PR) — it depends on that PR's Stationary Energy capability routes and inventory tools. Retarget to `develop` once 5883 merges.
- The Climate Advisor service runs from a pre-built image; **rebuild it** to pick up the agent changes (command in the doc).
- Verification: `tsc` / ESLint / Jest clean, and verified end-to-end in-browser. i18n is EN-only (CI translates the other locales on merge).


piotrnowakowski (approved):

app/src/components/StationaryEnergyDraft/stationary-energy-artifact-panel.tsx
```diff
@@ -481,17 +426,6 @@ export function ArtifactPanel({
...
-  // Keep the focused row visible as the user steps through reviews.
-  useEffect(() => {
-    if (!focusedProposalId) {
-      return;
-    }
-    const target = rowsScrollRef.current?.querySelector<HTMLElement>(
-      `[data-row-id="${CSS.escape(focusedProposalId)}"]`,
-    );
-    target?.scrollIntoView({ block: "nearest", behavior: "smooth" });
-  }, [focusedProposalId]);
```
lemilonkh, line 494: A lot of logic was removed in this file. Can we double check if that was intentional?

lemilonkh (approved):
Great work, thanks 🙌 ✨

app/src/components/StationaryEnergyDraft/stationary-energy-artifact-panel.tsx
piotrnowakowski: Yes, intentional. The draft selector/new-draft controls moved out of `ArtifactPanel` into `StationaryEnergyDraftsPanel`, and the start/review flow now lives in the chat panel. The one behavior not moved 1:1 is progress-row auto-scroll/focus; that was dropped as part of making the progress panel read-only rather than the review driver.

PR #2739, opened 2026-06-26
fix: globalapi version

amanda-eames:
### Summary
Sets up the structural groundwork for a v2 of the Global API without changing any existing
behaviour. Existing v0/v1 endpoints keep the exact same URLs and responses — this PR is about
*organisation*, a new response standard, and a reference v2 endpoint to build on.

### Why
The API had v0 and v1 prefixes scattered across route files with no clear boundary, hand-built
response dicts with no shared contract, and docs that intermixed versions. This PR introduces a
clean seam so new work lands in a consistent, self-describing v2 surface while v0/v1 stay frozen
and stable for current consumers (e.g. HIAP).

### Changes

**Versioned route structure**
- Moved the existing endpoint files into `routes/legacy/` (frozen; prefixes and behaviour
  unchanged) and repointed imports in `main.py`.
- Added `routes/v2/` for the new surface. `v2_router` is prefix-less; the mount point supplies
  `/api/v2`.

**v2 response standard**
- New `engineering-standards/api-response-semantics.md`: defines the "self-describing
  observation" contract and a conformance checklist an endpoint must pass to be v2 (typed values,
  explicit units, inline provenance, notation keys, request echo).
- New `schemas/v2.py` Pydantic models (`Observation`, `NotationKey`, `Provenance`,
  `MonetaryAmount`, `Meta`, `Envelope`) that encode the standard as types.

**Reference v2 endpoint**
- `routes/v2/finance.py` migrates the finance-opportunities endpoint to the v2 contract: typed
  `Envelope[...]` response_model, camelCase keys, `amount` bound to its currency, typed
  `provenance`. Same data/query as the legacy version — only the contract changed.
- `db/dependencies.py` adds an injectable `get_session`, so v2 handlers take the session via
  `Depends` (overridable in tests).

**Version-aware docs**
- `main.py` adds an app description, tag descriptions, and a custom `/docs` with a version
  selector (top-right dropdown) switching between v0/v1 and v2.
- v2 is mounted as a sub-app with its own clean page at `/api/v2/docs`.

**Tests**
- Split into `tests/legacy/` (existing tests, moved) and `tests/v2/`.
- `tests/v2/test_finance.py` is the new template: swaps the DB via `app.dependency_overrides`
  instead of string-path monkeypatching, so it doesn't break when files move.

**Small cleanups**
- Removed dead files: `database_init.py` (unused duplicate of the new session dependency) and an
  empty `tests/test.py`.
- `db/database.py` reuses the shared `settings` singleton instead of creating a second instance.
- `.gitignore`: ignore `.fuse_hidden*`.

### Behaviour / compatibility
- **No change to existing endpoints.** All v0/v1 URLs and responses are identical.
- Mounting the v2 sub-app makes `/api/v2` reachable when deployed — additive and non-breaking
  (one new endpoint). No `deprecated` flags set on legacy yet, pending downstream comms.

### Verification
- App imports and serves all 42 existing endpoints; OpenAPI generates cleanly.
- v2 finance tests pass via dependency overrides.
- All touched modules compile; no stale import or monkeypatch references.

### How to review
- The bulk of the diff is the `routes/` → `routes/legacy/` file move (renames, no content
  change) — review that first and quickly, then the genuinely new code in `routes/v2/`,
  `schemas/v2.py`, `db/dependencies.py`, and the docs wiring in `main.py`.


lemilonkh (approved):
Amazing work restructuring the API responses to be self describing. Thanks 🙌✨

PR #2768, opened 2026-07-01
fix(ui): ON-5971 replace automatic admin panel redirect

lemilonkh:
Replace with access denied UI so when the cached session is updated the full page can be rendered (instead of immediately triggering the redirect because the outdated session data from a previously logged in account is still in the cache).

Also fixed hydration issue on the starting admin panel page.


cephaschapa (approved):
Great work 🐐

PR #2769, opened 2026-07-01
Title: fix: ON-5968 city dashboard shows N/A for total emissions on prefilled inventory

tatimoreira:
[Ticket: ON-5968](https://openearth.atlassian.net/browse/ON-5968)

## Problem
The city dashboard and GHGI homepage showed "N/A" for Total Emissions even when a prefilled inventory had data sources connected. Two root causes:

1. `SUM(co2eq)` on all-null `InventoryValue` rows (e.g. notation-key-only prefill) returns SQL NULL — the frontend truthy check treated this as "no data"
2. The `/dashboard/` route returned raw DB inventories via `inv.toJSON()`, reading the `total_emissions` column which is never written to — so `totalEmissions` was always null regardless of connected data

## Solution
- **`InventoryService.ts`** — changed the SQL to return `0` when `InventoryValue` rows exist but all `co2eq` are null, and `NULL` only for truly empty inventories (no rows at all)
- **`dashboard/route.ts`** — added a bulk `SUM(co2eq)` query per inventory before building the response, replacing the never-populated DB column value
- **`HomePageJN/Hero.tsx` + `GHGIHomePage/HomePage.tsx`** — changed truthy check (`totalEmissions ?`) to `!= null` so `0` renders as "0 kg CO2e" instead of falling through to N/A

## Testing
- [ ] City with a prefilled inventory (notation keys only) — dashboard shows "0 kg CO2e" instead of N/A
- [ ] City with a prefilled inventory with actual data — dashboard shows correct emissions value
- [ ] Brand new city with no inventory data — dashboard still shows N/A
- [ ] `/cities/[cityId]/dashboard/` route
- [ ] `/cities/[cityId]/GHGI/[inventoryId]` route

## Notes
The underlying architectural issue [(ON-2429)](https://openearth.atlassian.net/browse/ON-2429?search_id=9f928bc2-2c79-49d0-9d05-140e41eeeba3&referrer=quick-find) — `Inventory.total_emissions` is never kept in sync — is out of scope for this fix but should be addressed to prevent the same class of bug recurring in future routes.

### DIFF
--- a/app/src/app/api/v1/city/[city]/dashboard/route.ts
+++ b/app/src/app/api/v1/city/[city]/dashboard/route.ts
@@ -15,6 +15,7 @@ import type {
 import { Inventory } from "@/models/Inventory";
 import type { AppSession } from "@/lib/auth";
 import { logger } from "@/services/logger";
+import { QueryTypes } from "sequelize";
 
 /**
  * @swagger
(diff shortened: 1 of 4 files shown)


lemilonkh (approved):
Wow, this is great! 🎉 ✨ 
And agreed, we should do a refactoring soon and remove unpopulated columns like the `totalEmissions` in `Inventory` to prevent issues like these.

PR #2771, opened 2026-07-01
fix: gracefully handle missing subcategories in GHGI version history

tatimoreira:
## Summary

Fix undefined subcategory display in the GHGI history tab. When a version record references a subCategoryId that no longer exists in the current GPC hierarchy, the API was throwing a 404 and crashing the entire history response. Even on the path where subCategoryId was null, the frontend concatenated undefined + " " + undefined, showing "undefined undefined" in the sub-category column.

## Changes

- app/src/app/api/v1/inventory/[inventory]/version-history/route.ts — findSubCategory() now returns undefined instead of throwing a 404 when a subCategoryId is not found in Inventory_Sector_Hierarchy; logs a warning instead of an error so history still loads for all other entries.

- app/src/components/GHGI/inventory-versions/VersionEntry.tsx — guard the subcategory display string so rows with an unresolvable subcategory show "-" instead of "undefined undefined".

## How to test

1. Navigate to a city → inventory → GHGI → History tab.
2. Confirm the tab loads without error even when version records contain stale subCategoryId values.
3. Rows whose subcategory can't be resolved should display "-" in the SUB-CATEGORY column instead of "undefined".

## Ticket
[ON-5974](https://openearth.atlassian.net/browse/ON-5974)

## Notes

The createHttpError import in route.ts is still used for the BadRequest on invalid module names — it was intentionally left in place.

## UI before and after 

Screenshot 2026-07-01 at 2 11 13 PM

Screenshot 2026-07-01 at 2 14 47 PM

[ON-5974]: https://openearth.atlassian.net/browse/ON-5974?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/app/api/v1/inventory/[inventory]/version-history/route.ts
+++ b/app/src/app/api/v1/inventory/[inventory]/version-history/route.ts
@@ -14,7 +14,7 @@ import { VersionHistoryEntry } from "@/util/types";
 
 const validModules = ["ghgi", "hiap"];
 
-function findSubCategory(subCategoryId: string): SubCategory {
+function findSubCategory(subCategoryId: string): SubCategory | undefined {
   const subCategories = Inventory_Sector_Hierarchy.flatMap((sector) =>
     sector.subSectors.flatMap((subSector) => subSector.subCategories),
   );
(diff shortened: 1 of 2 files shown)


lemilonkh (approved):
Nice, thanks 👍

PR #2772, opened 2026-07-02
fix: improve inventory mapping screen display values

tatimoreira:
## Summary

Improves the inventory mapping screen (upload flow) to display empty/undefined field values as "Not specified" in tertiary color instead of a bare dash, making it clear when a field is intentionally unspecified rather than an error. Also groups detected fields visually at the top of the mapping table, converts emission factor example values to metric tonnes (t CO₂e) for readability, and fixes an S3 PermanentRedirect error that prevented file uploads when the bucket's AWS region did not match the client's default.

## Changes

- Replace "-" placeholder with "Not specified" styled in content.tertiary color across the validation results, mapping columns, and review/confirm steps
- Sort mapping columns so auto-detected fields (interpretedAs !== null) appear first, grouping required fields together visually
- Convert emission factor example values (CO₂, CH₄, N₂O, Total CO₂e) from raw kg units to metric tonnes (t CO₂e) in both the validation results and mapping columns steps
- Add followRegionRedirects: true to the S3 client in InventoryFileStorageService to handle buckets hosted in a region that differs from AWS_FILE_UPLOAD_REGION — previously the SDK threw a PermanentRedirect error instead of following the redirect automatically
- Add "not-specified" i18n key across all 5 supported locales (en, es, fr, de, pt)

## How to test

1. Navigate to /{lng}/cities/{cityId}/GHGI/onboarding/import?inventory={inventoryId}
2. Upload a CSV or Excel inventory file (e.g. a CRF-format file)
3. Validation Results step: columns where no example value or interpretation was detected should show "Not specified" in light grey instead of "-"; emission factor columns (CO₂, CH₄, N₂O, Total CO₂e) should display values in t CO₂e (e.g. 63.1 gCO₂e/MJ → 0.0631 t CO₂e)
4. Mapping Columns step: same "Not specified" and t CO₂e behaviour; detected fields should appear above undetected ones
5. Review & Confirm step: if source file or format cannot be determined, those fields show "Not specified" in tertiary color rather than "-"
6. If AWS_FILE_UPLOAD_S3_BUCKET_ID is set and the bucket lives in a different region than AWS_FILE_UPLOAD_REGION, uploads should succeed rather than failing with a PermanentRedirect error

## Preview

Screenshot 2026-07-01 at 10 08 51 PM

## Ticket

[ON-5985](https://openearth.atlassian.net/browse/ON-5985?atlOrigin=eyJpIjoiNjJjNWY3OWM4Y2U3NDI3NTliNDlhY2YyNDFjNDliMWUiLCJwIjoiaiJ9)

## Notes

S3 redirect fix detail: the AWS SDK v3 S3 client by default treats a 301 PermanentRedirect response (returned when the bucket's region doesn't match the request endpoint) as a fatal error. Adding followRegionRedirects: true to the client config instructs the SDK to re-issue the request against the correct regional endpoint automatically. This means uploads work correctly even if AWS_FILE_UPLOAD_REGION is unset (defaults to us-east-1) or misconfigured — no change to environment variable setup is required for existing deployments.

[ON-5985]: https://openearth.atlassian.net/browse/ON-5985?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/components/steps/GHGI/import/validation-results-step.tsx
+++ b/app/src/components/steps/GHGI/import/validation-results-step.tsx
@@ -46,6 +46,32 @@ export default function ValidationResultsStep({
     (col) => col.status === "detected",
   ).length;
 
+  const EMISSION_FACTOR_LABELS = new Set([
+    "Emission Factor - CO2",
+    "Emission Factor - CH4",
+    "Emission Factor - N2O",
+    "Emission Factor - Total CO2e",
+  ]);
+
+  const formatExampleValue = (column: ColumnInfo): string | null => {
+    if (!column.exampleValue) return null;
+    if (
+      column.interpretedAs &&
+      EMISSION_FACTOR_LABELS.has(column.interpretedAs)
+    ) {
+      const num = parseFloat(column.exampleValue);
+      if (!isNaN(num)) {
+        const tonnes = num / 1000;
+        const formatted =
+          tonnes < 0.0001
+            ? tonnes.toExponential(2)
+            : parseFloat(tonnes.toFixed(4)).toString();
+        return `${formatted} t CO2e`;
+      }
+    }
+    return column.exampleValue;
+  };
+
   return (
     <Box w="full">
       <Box display="flex" flexDir="column" gap="24px" mb={6}>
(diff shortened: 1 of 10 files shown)


app/src/components/steps/GHGI/import/validation-results-step.tsx
```diff
@@ -46,6 +46,32 @@ export default function ValidationResultsStep({
     (col) => col.status === "detected",
   ).length;
 
+  const EMISSION_FACTOR_LABELS = new Set([
+    "Emission Factor - CO2",
+    "Emission Factor - CH4",
+    "Emission Factor - N2O",
+    "Emission Factor - Total CO2e",
+  ]);
```
lemilonkh, line 54: Have you tested if this works when another language is selected? So are these always in english even when CityCatalyst is set to e.g. spanish?

lemilonkh (approved):
Amazing work, thanks 🙌 ✨

app/src/components/steps/GHGI/import/validation-results-step.tsx
tatimoreira: image
Yes all of them are in english no matter the selected language, should I add translations to that in this ticket ?

lemilonkh: No, that's fine for now as long as the functionality works as intended. Let's keep this in mind for a future improvement.

tatimoreira: as discussed in [Slack](https://openearthfoundation.slack.com/archives/D0BCC2P4HLZ/p1783355438735789) with @lemilonkh translations will be resolved as an improvement in a upcoming ticket

PR #2779, opened 2026-07-03
fix(ON-5952): email error when changing the user role

cephaschapa:
### Summary
Fixes a 500 error when changing a user's role to organization admin. The PATCH /organizations/{id}/role endpoint crashed with "Transaction cannot be rolled back because it has been finished with state: commit" when email sending failed after the database transaction had already been committed.

### Changes
Track committed state on the transaction and only call rollback() if the transaction is still pending
Wrap post-commit email sends in try/catch so email failures don't crash the request after a successful DB change
Return { success: true, emailSent: false } instead of throwing 500 when the invite email fails (the invite/role was already persisted)
Add missing "updating-role" i18n key to all 5 locale files (en, fr, es, de, pt)

### How to test
Go to Admin → Organization → Team
Promote an existing user to organization admin (or invite a new email)
Verify the role change succeeds and a loading toast shows "Updating role..." (localized)
To reproduce the original bug: temporarily misconfigure SMTP credentials — the role change should still succeed with a logged warning, instead of returning a 500 error
Ticket
ON-5952

### DIFF
--- a/app/src/app/api/v1/organizations/[organization]/role/route.ts
+++ b/app/src/app/api/v1/organizations/[organization]/role/route.ts
@@ -58,6 +58,7 @@ export const PATCH = apiHandler(async (req, { params, session }) => {
   const user = await db.models.User.findOne({ where: { email } });
 
   const transaction = await db.sequelize!.transaction();
+  let committed = false;
   try {
     if (user) {
       // Step 1: Promote to OrganizationAdmin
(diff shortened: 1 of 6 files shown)


app/src/app/api/v1/organizations/[organization]/role/route.ts
```diff
@@ -125,17 +126,22 @@ export const PATCH = apiHandler(async (req, { params, session }) => {
...
+      try {
+        await EmailService.sendRoleUpdateNotification({
+          email,
+          brandInformation: {
+            color: org.theme.primaryColor,
+            logoUrl: org.logoUrl as string,
+          },
+          organizationName: org.name as string,
+          user,
+        });
```
Comment (lemilonkh) [inline app/src/app/api/v1/organizations/[organization]/role/route.ts:141]: It would be good to also look into why the email failed to send?
That doesn't seem to be addressed by this PR.
It might have been an email template rendering error?

lemilonkh (approved):
Thanks 👍 ✨

PR #2780, opened 2026-07-05
ON-5884 add Stationary Energy notation key tools

piotrnowakowski:
## What
- Add Stationary Energy notation-key capability routes for Climate Advisor to list and commit notation-key targets.
- Extend the CA review tooling, prompt policy, draft review state, serializers, and UI decision flow for staging, confirming, saving, committing, and rolling back notation keys.
- Add the database migration, docs, and regression coverage for the notation-key flow.

## Why
- Lets the CA agent handle newly introduced notation-key decisions through the same staged review/commit workflow used for accepted activity data, while preserving user confirmation boundaries.

## Validation
- PASS: browser record-break flow recorded at `output/browser-demo-recording/on5884-notation-key-edge-cases.mp4`; exercised invalid key rejection, invalid target rejection, bulk confirmation, save-confirmation boundary, single valid stage, and rollback.
- PASS: `npm run build` in `app`.
- PASS: app Jest on Windows with explicit ESM flags, `NODE_OPTIONS=--experimental-vm-modules NODE_ENV=test npx jest --runInBand`; 42 suites passed, 3 skipped; 348 tests passed, 31 skipped, 1 todo.
- PASS: CA pytest split around a pytest import-order artifact: `uv run pytest tests/test_text_splitter.py -q -s --import-mode=importlib` passed 5 tests, and `uv run pytest tests/ -q --ignore=tests/test_text_splitter.py` passed 166 tests with 5 skipped.
- KNOWN FAILING: one-shot CA `uv run pytest tests/ -q` still collection-fails on `tests/test_text_splitter.py` importing `scripts.splitter_baseline`, while that file passes alone with `--import-mode=importlib`.
- KNOWN FAILING: `npx playwright test --reporter=line` after installing Firefox runs the configured browser matrix but fails 3 validation-attribute expectations unrelated to notation keys: `e2e/login.spec.ts:39` in Chromium and Firefox, and `e2e/signup.spec.ts:67` in Firefox. Result: 12 passed, 60 skipped, 3 failed.


climate-advisor/docs/architecture.md
```diff
@@ -103,14 +103,19 @@ sequenceDiagram
   - `inventory_status_overview`
   - `inventory_emissions_context`
```
mircorudolph, line 104: We add them here "- Added when the request is scoped to a Stationary Energy draft run:"

But arent those already added further above under "- Added when the request has CityCatalyst credentials and thread scope:"?

Since we cannot enter stationary energy drafting/ emissions with out the " Added when the request has CityCatalyst credentials and thread scope:" step?

mircorudolph (commented):
climate-advisor/docs/architecture.md says that we have different tools registered e.g. per:

"- Added outside active Stationary Energy review chat:"
"- Added when the request has CityCatalyst credentials and thread scope:"
"- Added when the request is scoped to the Stationary Energy draft surface and
  no draft run is active:"
... and so on.

However we seem to only have 2 .md files for prompts in prompts/tools.
Therefore it is likely that we describe a tool, that might not be registered (since our registering is more granular then these 2 prompt files.)

This means we are likely telling the agent it has a tool available in prompt and tell it how to use it, while it is not registered.

Suggestion: Align this more granular so that we have 1-2 more prompt files that we load fitting to the in architecture docs describe sections
Goes hand-in-hand with my other comment that according to architecture docs we are not fully clear on in which stage something gets registered (e.g. inventory_list_accesible when the credentials are scoped and later again for stationary energy)

mircorudolph (approved):
Thanks!

climate-advisor/docs/architecture.md
piotrnowakowski: docs drift i aligned them

PR #2781, opened 2026-07-06
feat(hiap): restyle action plan PDF exports with CityCatalyst branding

carlosgraffi:
## Summary

Restyles both PDF exports in the **HIAP / Climate Action** module to a shared, on-brand CityCatalyst visual language. Previously the two exports looked unrelated — different fonts, no branding, plain black-on-white text. They now share a brand-blue logo, consistent Helvetica typography, styled section headers, and branded footers.

There are two separate exports, built with two different libraries:

| Export | Trigger | Library |
|---|---|---|
| **Ranked actions** | "Export as PDF" on the HIAP tab | `@react-pdf/renderer` |
| **Implementation plan** | "Export PDF" in the generate-plan drawer | `jsPDF` |

## Motivation

The exports are shared with stakeholders, so they should look like a finished CityCatalyst artifact. This is a **styling-focused** pass (targeted for the current release), deliberately scoped to be low-risk.

### Ranked actions PDF — `PrintableActionPlanPDF.tsx`
- Brand palette, logo header (city name + action type), `#rank` chip, blue section labels, metadata pills, and a per-metric cost/timeline block.
- **Adds two fields that the action drawer already shows but the PDF was omitting:** primary purposes (tags) and the ranking **Explanation** (reusing the drawer's exact `lng → en → "no explanation available"` fallback).
- Now shows sectors for adaptation actions too, matching the drawer.
- Passes `lng` through from `HiapTab` for the explanation language.

### Implementation plan PDF — `PDFExportService.ts`
- Brand palette, logo header, accent-bar section headings, numbered circle badges, colored intro labels, and a branded footer with a hairline rule.
- **Bug fix:** the service called `doc.setFont("", …)` with an empty font family, which made jsPDF silently fall back to **Times** (serif). It now explicitly uses **Helvetica**, matching the ranking PDF. This is why the two exports previously had different fonts.
- Made `async` to embed the logo; `GeneratePlanDrawer` now `await`s it. Logo loading is guarded — if it fails, the PDF still generates without it.

### New asset — `citycatalyst-logo-blue.png`
- The ring (`logo.svg`) + wordmark (`city_catalyst_logo.svg`) recomposited from the brand SVGs, recolored to `#2351DC` on a **transparent** background, rasterized at **6×** (996×216) for crisp print.
- Neither PDF library embeds SVG cleanly (`@react-pdf/renderer` `<Image>` and jsPDF `addImage` both want raster; jsPDF would need an extra `svg2pdf` dependency), so a high-DPI PNG derived from the vector is the sharp, dependency-free choice.

## Design decision: only export what the UI shows

Both data models carry more fields than the UI renders (co-benefits, KPIs, equity considerations, MER indicators, institution URLs). By deliberate decision, **the PDFs only include data that is already displayed in the corresponding UI** (the action drawer and the plan drawer). So those extra fields stay out — nothing new is surfaced to users through the PDF that they can't already see on screen.

## Scope & safety

- **Frontend-only.** No backend / Python changes.
- **No i18n changes.** All labels reuse existing translation keys, so no missing-translation risk in other languages.
- Typechecks clean (`tsc --noEmit`); the only reported errors are pre-existing stale `.next/types` files, unrelated to this change.

## Testing

- Both PDFs were rendered from the actual code with real production content (a Barra do Quaraí action + plan) and visually verified: logo sharpness, font consistency, section styling, the new purpose/explanation fields, and the empty-explanation fallback.
- To review locally: open a city's HIAP tab → **Export as PDF** for the ranked actions; open an action's generate-plan drawer → **Export PDF** for the plan.

## Files
- `app/src/components/HIAP/PrintableActionPlanPDF.tsx`
- `app/src/app/[lng]/cities/[cityId]/HIAP/HiapTab/index.tsx`
- `app/src/services/PDFExportService.ts`
- `app/src/components/HIAP/GeneratePlanDrawer.tsx`
- `app/public/assets/citycatalyst-logo-blue.png` (new)


lemilonkh (approved):
Great, thanks 👍 ✨

PR #2790, opened 2026-07-06
Fix post-deploy CA smoke checks

piotrnowakowski:
﻿## Summary
- Remove endpoint inspection from the post-deploy CA/CC smoke path so endpoint RBAC cannot block deployments.
- Remove the extra CA `/health` probe and its `CA_SERVICE_NAME` workflow input; Kubernetes rollout/readiness already covers CA health.
- Remove the redundant `CC_EXPECTED_AUDIENCE` workflow input; the CA-to-CC contract smoke now validates against the deployed CA pod's own `CC_BASE_URL`.
- Keep the blocking smoke focused on fixture seeding, Climate Advisor rollout/pod lookup, and `python -m scripts.smoke_cc_contract` from the deployed CA pod.

## Why
The previous post-deploy smoke mixed real contract validation with Kubernetes operations that can fail for RBAC reasons or duplicate checks Kubernetes already performs. The blocking post-deploy test should fail when the CA-to-CC contract is broken, not because the Actions user cannot inspect endpoints or create extra probe pods.

## Validation
- `bash -n .github/scripts/cc-ca-post-deploy-smoke.sh`
- `uv run --directory service pytest tests/test_deployment_contract.py -q` -> 19 passed
- `git diff --check`

### DIFF
--- a/.github/scripts/cc-ca-post-deploy-smoke.sh
+++ b/.github/scripts/cc-ca-post-deploy-smoke.sh
@@ -5,9 +5,7 @@ NAMESPACE="${K8S_NAMESPACE:-default}"
 JOB_TIMEOUT="${JOB_TIMEOUT:-180s}"
 
 : "${SMOKE_FIXTURE_MANIFEST:?SMOKE_FIXTURE_MANIFEST is required}"
-: "${CA_SERVICE_NAME:?CA_SERVICE_NAME is required}"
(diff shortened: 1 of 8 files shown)


lemilonkh (approved):
Thanks 👍

lemilonkh (approved):
🙌

PR #2791, opened 2026-07-06
fix: Remove duplicate 'edit' translation keys in onboarding

pablo-ibco:
## Summary
- Fixed untranslated EDIT button in onboarding flow by removing duplicate translation keys
- Affected all non-English languages: Spanish, Portuguese, French, and German

## Changes
Removed duplicate "edit" keys from the end of each non-English `onboarding.json` translation file:
- `app/src/i18n/locales/es/onboarding.json`
- `app/src/i18n/locales/pt/onboarding.json`
- `app/src/i18n/locales/fr/onboarding.json`
- `app/src/i18n/locales/de/onboarding.json`

The duplicate keys were causing JSON parsers to only use the last occurrence, which was preventing proper translation of the EDIT button in the onboarding confirmation step.

## Root Cause
Each translation file had two identical "edit" keys:
1. One correctly placed around line 199-200 (kept)
2. One duplicate at the end of the file (removed)

Since JSON does not allow duplicate keys, only the last occurrence was being used, which may have been causing translation inconsistencies.

## Test plan
- [x] Validated JSON syntax for all modified files
- [ ] Manual testing: Navigate through onboarding flow in Spanish/Portuguese/French/German
- [ ] Verify EDIT button is translated in the confirm step

## Tests added
N/A - This is a translation fix without code changes. All JSON files are syntactically valid.


lemilonkh (approved):
Good spot, thanks 👍 ✨

PR #2794, opened 2026-07-07
fix: block inventory import when inventory already contains data

cephaschapa:
### Summary
Prevents duplicate inventory imports by blocking the upload wizard when the current inventory already contains data (InventoryValue records), with both a frontend warning banner and a backend safety net.

### Changes
- Frontend (import page): Query getInventoryProgress to detect existing InventoryValue records. When data exists (thirdParty + uploaded + reasonNE + reasonNO > 0), show an orange warning banner on the upload step, disable file upload, and disable the Continue button

- Backend (approve route): Count InventoryValue records for the inventory before approving import. Return 409 Conflict if any records exist, preventing duplicate imports even if the frontend is bypassed
- i18n: Added inventory-already-has-data-title and inventory-already-has-data-description keys to all 5 locale onboarding.json files (en/fr/es/de/pt)
- 
### How to test
Navigate to the import wizard for an inventory that already has data (e.g. connected third-party sources or uploaded values)

### DIFF
--- a/app/src/app/[lng]/cities/[cityId]/GHGI/onboarding/import/page.tsx
+++ b/app/src/app/[lng]/cities/[cityId]/GHGI/onboarding/import/page.tsx
@@ -1,8 +1,8 @@
 "use client";
 
 import { useTranslation } from "@/i18n/client";
-import { MdArrowBack, MdArrowForward } from "react-icons/md";
-import { Box, Icon, Text, useSteps } from "@chakra-ui/react";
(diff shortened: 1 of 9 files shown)


lemilonkh (commented):
Can we generically disable creating inventories for cities that already have an inventory for this year instead? The same problem exists if you create an inventory for an existing year for the city with the normal process.

lemilonkh (approved):
Amazing, thanks 🙌 ✨

PR #2796, opened 2026-07-07
fix: show 'Manually added data' in version history source column

cephaschapa:
### Summary
Fixes the version history table's SOURCE column to show meaningful labels instead of a dash (-) when no third-party datasource is linked.

### Changes
VersionEntry.tsx: Updated source label logic to distinguish between:
Datasource name — shown when a third-party source is linked (e.g., "Global Energy Monitor")
"Manually added data" — shown when data was entered manually (has co2eq but no datasourceId)
"Not specified" — fallback when neither applies
i18n: Added manually-added-data key to all 5 locale dashboard.json files (en/fr/es/de/pt)


lemilonkh (approved):
Awesome, thanks 👍 ✨

PR #2797, opened 2026-07-07
fix: resolve invalid element error in GrowthRatesExplanationModal

cephaschapa:
### Summary
Fixes the runtime error Element type is invalid: expected a string...but got: undefined in GrowthRatesExplanationModal by updating Chakra UI v2 patterns to v3.


lemilonkh (approved):
Thanks! ✨

PR #2798, opened 2026-07-07
fix(ui): show total emissions on datasource cards when non-zero

cephaschapa:
### Summary
Improves the third-party datasource cards layout and emissions data visibility across the GHGI data steps page and the activity tab.

### Changes

Grid layout: Replaced SimpleGrid with CSS Grid using templateColumns="repeat(3, 1fr)" for a consistent 3-column card layout
Card container: Changed VStack to HStack for scope group layout
Emissions visibility: Always display total emissions (co2eq_100yr) on datasource cards, including connected ones — hidden only when the value is null, undefined, or 0
activity-tab.tsx
Emissions visibility: Same logic — show total emissions regardless of connection status, hide only when value is null, undefined, or 0

### DIFF
--- a/app/src/components/GHGI/inventory-pages/add-data-steps-page.tsx
+++ b/app/src/components/GHGI/inventory-pages/add-data-steps-page.tsx
@@ -34,6 +34,7 @@ import {
   Card,
   Center,
   Flex,
+  Grid,
   Heading,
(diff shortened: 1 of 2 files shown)


lemilonkh (approved):
Great, thanks 👍

PR #2801, opened 2026-07-07
ON-6043: fix subsector total emissions + org invite email success check

pablo-ibco:
## Summary

Two small fixes surfaced by ON-6043 and ON-6041.

- **ON-6043 — GHGI subsector "Total emissions" showed `0 t CO2e`.** The footer read `inventoryValue.co2eq` (a backend aggregate that can be stale/`0`) instead of the emissions actually shown in the scope. It now sums the scope's activity values, matching the group header, so the footer is always consistent with the table.
- **ON-6041 (related hardening) — email delivery failures were silently swallowed.** `EmailService.sendOrganizationInvitationEmail` returns `{ success, inviteUrl }`; the object is always truthy, so `if (!emailSent)` in the org role route never detected SES failures. It now checks `.success` (keeping the existing non-fatal behavior: the invite is still created and the endpoint returns `{ success: true, emailSent: false }`).

## Changes

- `app/src/components/Tabs/Activity/emission-data-section.tsx`: compute the scope total from `activityValues` instead of `inventoryValue?.co2eq`.
- `app/src/app/api/v1/organizations/[organization]/role/route.ts`: check `emailResult.success` instead of the always-truthy object.

## How to test

1. Open a GHGI inventory → enter a subsector with activities in a scope (e.g. `I.2`, Scope 1, Direct Measure).
2. Confirm the footer **Total Emissions** now matches the sum of the listed activities (previously `0 t CO2e`).
3. (Role route) With SES misconfigured, promoting a non-existing user to org admin no longer crashes; it returns `{ success: true, emailSent: false }` and logs the failure.

## Ticket

ON-6043, ON-6041

## Notes

The primary ON-6041 outage was an **infra** issue, fixed separately: production sends from `info@citycatalyst.io`, but that domain was not a verified SES identity (SES rejected every send with `554 Message rejected: Email address is not verified`). The `citycatalyst.io` domain has now been added as an SES identity with DKIM records in Route53. This PR only contains the related code hardening.


lemilonkh (approved):
Amazing, thanks 👍 ✨

PR #2777, opened 2026-07-03
ON-5981: Consolidate import flow into 3-step inventory mapping screen

tatimoreira:
## Summary

Consolidates the previous 4-step import flow (Upload → Validation Results → Mapping Columns → Review & Confirm) into a 3-step flow (Upload → Inventory Mapping → Review & Confirm). The new Inventory Mapping step combines validation feedback and column mapping into a single inline-editable table, gated by a Continue button that requires required fields to be mapped.

## Changes

- Add `InventoryMappingStep` component: single table with FIELD NAME / VALUE / MAP TO columns, mandatory fields marked with red asterisks, inline `NativeSelect` dropdowns for all columns, red error banner when required fields are unmapped, orange warning banner when optional fields are unmapped, "Needs remapping" indicators per row, mandatory columns sorted to the top
- Refactor import `page.tsx` from 4 steps to 3 steps; derive `canContinue` via `useMemo` in the parent (instead of `useEffect` in child) to ensure the Continue button updates synchronously when mappings change
- Add i18n keys for the new step across all 5 locales (en, pt, es, fr, de)
- Fix Continue button gate: required fields (GPC ref. no., Sector, Sub-sector, Activity Amount) must all have a non-empty mapping selected; clearing and re-selecting any value correctly enables/disables the button
- Display city name above the heading on all three import steps for context

## Preview 

image
image
image

## How to test

1. Navigate to GHGI → Import inventory for any city with an inventory year
2. **Step 1 – Upload**: city name appears above heading; upload a CSV/XLSX file
3. **Step 2 – Inventory Mapping**: all columns appear in the table; required fields have red asterisks; red error banner is shown until GPC ref. no. (or Sector + Sub-sector) and Activity Amount are all mapped; orange warning appears when optional fields are unmapped; Continue button only enables when all required fields have a selection
4. Clear a required field → Continue disables; re-select any value → Continue re-enables
5. **Step 3 – Review & Confirm**: city name appears above heading; summary and field mappings are correct; Import button triggers the import

## Ticket

[ON-5981](https://openearth.atlassian.net/browse/ON-5981)

## Notes

The `canContinue` check mirrors the backend validation rule: identity is satisfied by `gpcRefNo` alone **or** both `sector + subsector` together; `activityAmount` is always required. The gate enforces that every auto-detected mandatory column has a non-empty value selected (not necessarily the originally detected key), which matches the backend's flexibility.

[...]

PR #2804, opened 2026-07-08
fix(ui): port emission factor display fixes to InventoryMappingStep

tatimoreira:
## Summary

PR #2772 improved example value display in the import mapping screen, but those fixes targeted `mapping-columns-step.tsx` which was deleted by PR #2777 when the 4-step import flow was consolidated into `InventoryMappingStep`. The new component was left with the original `col.exampleValue || "-"` raw display.

- Port `EMISSION_FACTOR_KEYS` + `formatExampleValue` to convert CO2/CH4/N2O/TotalCO2e emission factor example values from raw units to metric tonnes (e.g. `63.1545` → `0.0632 t CO2e`)
- Replace `"-"` placeholder with `"Not specified"` in `content.tertiary` color so empty fields are visually distinct from actual data
- Update sort order: required fields first → auto-detected fields → undetected, so Activity data amount and other mandatory fields always group at the top

## Test plan

- [ ] Go to `/GHGI/onboarding/import/` and upload a GHGI file
- [ ] On the mapping step, confirm emission factor columns show values in `t CO2e` format
- [ ] Confirm columns with no example data show "Not specified" in grey instead of "-"
- [ ] Confirm required fields (GPC Ref No, Sector, Subsector, Activity Amount) appear at the top of the list
- [ ] Confirm auto-detected non-required fields appear above undetected fields

### DIFF
--- a/app/src/components/steps/GHGI/import/inventory-mapping-step.tsx
+++ b/app/src/components/steps/GHGI/import/inventory-mapping-step.tsx
@@ -18,6 +18,13 @@ import type { ColumnInfo, RequiredMappingOption } from "@/util/types";
 
 const MANDATORY_KEYS = new Set(["gpcRefNo", "sector", "subsector", "activityAmount"]);
 
+const EMISSION_FACTOR_KEYS = new Set([
+  "emissionFactorCO2",
+  "emissionFactorCH4",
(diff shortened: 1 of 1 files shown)


lemilonkh (approved):
Great, thanks 👍

PR #2821, opened 2026-07-08
Add Concept Note Builder architecture

piotrnowakowski:
## Summary
- Adds the Concept Note Builder architecture document.
- Includes the branch's existing deployment workflow, smoke-test, and deployment-contract updates.

## Validation
- Not run locally; PR creation only.

Base: `develop`
Head: `ON-5885-Concept_Note_Architecture`


pablo-ibco (commented):
Reviewed the architecture doc (the net change in this PR is `docs/ConceptNoteBuilderArchitecture.md`). Overall it's a solid, well-structured design — clear CC vs Climate Advisor separation, step-scoped tools, soft-delete + revision audit trail, and good "no fabrication" guardrails.

These are just small, document-level comments (design consistency / contract clarity), not blockers:

1. **Access control isn't specified on run creation/access.** `concept_note_start_run` takes `user_id`/`city_id`/`project_id` and `GET /v1/concept-notes/{run_id}` exposes run state, but nothing says who authorizes access to a city/project or to an existing `run_id`. Since CNB pulls GHGI/CCRA/HIAP city data and snapshots it, worth a line stating the CC bridge enforces permissions and CA validates run ownership on every route.

2. **Start contract vs. step machine mismatch.** `concept_note_start_run` already requires `funder_id` + `opportunity_id` and returns `workflow_step="assembling_context"`, but the Workflow Steps diagram starts at `selecting_scope → ingesting_user_files → profiling_funder → matching_examples → assembling_context`. Worth clarifying whether scope/funder selection happens before the run exists (so the machine starts later) or the initial step should be `selecting_scope`.

3. **Chapter schema/contract mismatch.** The Document Workspace field list (and `document_list_chapters` output) includes `latest_revision_id`, but the `concept_note_chapters` ERD omits it. Since the doc is an explicit integration contract for the datateam DB, worth aligning the ERD and field list (and reconciling the `status` enum vs the separate `deleted` bool).

4. **Gap reopen isn't modeled.** `document_restore_chapter` should "reopen any gaps that were closed only because the chapter was deleted", but `concept_note_gaps` has only `status` — no closure reason. A `closed_reason`/`closed_by_chapter_delete` field would make that rule implementable.

5. **PDF converter integration contract undefined.** The converter is on the critical ingest path; internals being out of scope is fine, but the doc could still define *how* CNB calls `PDF_converter` (library vs HTTP service vs subprocess), sync/async, size/timeout limits, and the failure payload beyond "converter failed".

Minor nits: Architecture Decision 2 ("CA owns pre-commit workflow state") reads as contradicting the State Ownership table (all workflow state lives in the datateam CNB DB) — a small reword would remove the ambiguity; and a line on permission/retention for the CC context snapshot stored in `concept_note_context_bundles` would be nice alongside comment 1.

pablo-ibco (approved):

docs/ConceptNoteBuilderArchitecture.md
```diff
@@ -0,0 +1,1614 @@
...
+  revisions, gaps, and export.
+- Funding reference tables for funders, funder criteria, templates, and similar
+  funded projects in the datateam managed CNB database.
+- A curated research ingest pipeline for funder profiles and funded-project
+  examples.
+- Runtime matching between the user's project and similar funded projects.
+- PDF upload ingestion using the supporting PDF converter repository.
+- Reuse of current CityCatalyst-to-Climate-Advisor connection for CC data.
+
+Out of scope:
```
mircorudolph, line 44: I would add: setting up databases (CA only consumes those sources)

docs/ConceptNoteBuilderArchitecture.md
```diff
@@ -0,0 +1,1614 @@
...
+
+Concept Note Builder should be implemented as a new Climate Advisor agentic
+workflow, following the same direction as the Stationary Energy workflow:
+
+1. CityCatalyst owns product data, user permissions, and committed module state.
+2. Climate Advisor owns conversation state and pre-commit agentic workflow state.
+3. The datateam managed CNB database stores reusable funder and funded-project
+   tables alongside CNB workflow tables.
+4. PDF ingestion should use the supporting PDF converter repository to convert
+   PDFs to markdown.
```
mircorudolph, line 62: as dicussed, this might be a useful integration into CA as an endpoint instead of that stand-alone repo currently being used)

docs/ConceptNoteBuilderArchitecture.md
```diff
@@ -0,0 +1,1614 @@
...
+        text body_markdown
+        jsonb patch_summary
+        timestamp created_at
+    }
+
+    concept_note_evidence_links {
+        uuid evidence_link_id
+        uuid chapter_id
+        uuid revision_id
+        string selected_source_label
```
mircorudolph, line 414: I’d make evidence links point to stable source/excerpt records rather than only `selected_source_label` + `source_location`. Since uploads, matched projects, and context bundle sections can be refreshed, labels/locations can become stale after a rebuild. Could we version the context bundle by section here, e.g. `context_bundle_version`, per-section versions or hashes, source/excerpt IDs, and generated-at timestamps, then have evidence links reference those stable IDs? That would make resumed runs and chapter citations auditable against the exact context slice the agent used.

docs/ConceptNoteBuilderArchitecture.md
```diff
@@ -0,0 +1,1614 @@
...
+| Project sector / action type | Projects in unrelated sectors or action categories. |
+| Applicant / recipient type | Awards to recipient types that do not match the user's applicant profile, such as nonprofit-only awards for a city-led project. |
+| Funded award or valid pipeline entry | Records that are not actual awards or, for priority-list routes, valid pipeline entries. |
+| Usable source evidence | Records without enough source evidence to show the user why the example is relevant. |
+
+If the user project or funder profile is missing a field needed for the future
+scoring concept, the workflow should not invent it. It should record a match
+caveat and create a gap if the missing field matters for drafting.
+
+## PDF Conversion
```
mircorudolph, line 722: For implementation, this needs an artifact contract: file ownership, object-store location, max size, timeout, retry, conversion status, content hash, virus/mime validation, markdown provenance, export template version, and cleanup/retention. This should be a first-class job model, not just a tool call.

Also like discussed this can be not that trivial with potential queueing, storage, ...

docs/ConceptNoteBuilderArchitecture.md
```diff
@@ -0,0 +1,1614 @@
...
+- Funder record and funding opportunity record, including route, instrument,
+  geography, live status, and award-size ranges.
+- Template chapter schema.
+- Stated eligibility criteria from program documents.
+- Derived matching signals, marked as derived.
+- Funded-project records, project-action records, and funding links.
+- Pipeline entries for priority-list routes where the funding order matters.
+- Evidence links for each important claim.
+
+## Similar Project Matching
```
mircorudolph, line 625: This can be done but I would advise that we also/ primarily work with deterministic prefiltering, tags, etc.

Projects can be tagged, funders can be tagged, etc. We should explore this and then if this is not enough, use an LLM filtering as a second stage

mircorudolph: This comment applied to all the ext. sources like projects, funders, ...
Best way would always be deterministic prefiltering

docs/ConceptNoteBuilderArchitecture.md
```diff
@@ -0,0 +1,1614 @@
...
+    }
+  ],
+  "document_context": {
+    "chapters": [],
+    "gaps": []
+  }
+}
+```
+
+## Database Model
```
mircorudolph, line 290: Its good as a proposal, but should be owned by data team to define this clearly and we follow what they build.

docs/ConceptNoteBuilderArchitecture.md
```diff
@@ -0,0 +1,1614 @@
...
+        string user_id
+        jsonb context
+    }
+
+    concept_note_runs {
+        uuid run_id
+        uuid thread_id
+        string user_id
+        string city_id
+        string project_id
```
mircorudolph, line 355: What is project_id here? The project ID of a city catalyst project? probably not I assume? Are those tied to a project?

mircorudolph (approved):

docs/ConceptNoteBuilderArchitecture.md
```diff
@@ -0,0 +1,1614 @@
...
+- Loading what CityCatalyst already knows.
+- Ingesting what the user uploads at intake or mid-flow.
+- Retrieving the selected funder's profile, rubric, and template.
+- Matching the project against comparable funded projects.
+- Identifying decisions or missing facts that cannot be grounded.
+
+The agent and document workspace then use that context bundle to:
+
+- Draft document chapters and show evidence links for user review.
+- Ask only for the identified decisions or missing facts.
```
carlosgraffi, line 134: The agent should also proactively provide insights, potential next steps and identify blind spots for the user. If it only asks for confirmation or missing information without providing more clarity, then we’ll be building just another form chat-like wizard

carlosgraffi (approved):
Overall seems aligned to this first attempt. There are still some moving parts like funder archetypes, its influence in the document format and chapter structure, the context bundle orchestration specially when the user either has a lot of data or very little. But we’ll figure these out on the way I guess. As long as this can behave openly (meaning it can react properly to several use cases like the ones described in the early mockup) and not tied to a single assumption/use case, I feel we’re ok to move forward.

docs/ConceptNoteBuilderArchitecture.md
piotrnowakowski: Added this to the interview behavior: alongside questions, the agent must proactively surface grounded insights, blind spots, and actionable next steps.

piotrnowakowski:
@pablo-ibco Thanks - we addressed the contract clarifications:

1. Added a split authorization boundary: CC checks CC-owned data, while CA checks run and conversation access on every route using trusted actor context.
2. Moved scope and funder/opportunity selection before run creation; runs now start at `ingesting_user_files`.
3. Added `latest_revision_id` and made `deleted` the sole soft-delete flag, separate from chapter status.
4. Added gap-closure provenance and explicit chapter delete/restore transitions.
5. Defined PDF conversion as a CA-owned API/module adapted from `PDF_converter`, not a standalone runtime dependency. Detailed job handling remains a follow-up.
6. Clarified that CA orchestrates pre-commit workflow state while the datateam database stores it.
7. Defined snapshot permission inheritance, authorization rechecks on resume/export, and retention/deletion behavior.

docs/ConceptNoteBuilderArchitecture.md
piotrnowakowski: Added this explicitly to out of scope: CA consumes the datateam database and upstream sources through contracts; it does not provision or migrate them.

docs/ConceptNoteBuilderArchitecture.md
piotrnowakowski: Aligned this to a CA-owned endpoint: the relevant converter code will be moved and adapted into Climate Advisor and exposed through the concept-note upload API, with no standalone runtime service.

docs/ConceptNoteBuilderArchitecture.md
piotrnowakowski: Added immutable bundle and section versions and hashes, plus stable `source_id` and `excerpt_id` references. Evidence links now target the exact context version used; labels and locations are display metadata only.

docs/ConceptNoteBuilderArchitecture.md
piotrnowakowski: Agreed this needs a separate design pass. We intentionally left the full artifact/job, queueing, retry, validation, and cleanup contract out of this revision.

docs/ConceptNoteBuilderArchitecture.md
piotrnowakowski: Updated the matching design to tag all external records during curated ingest, then apply deterministic gates and tag-based filtering first. LLM filtering is disabled in v1 and shown only as an optional later refinement.

docs/ConceptNoteBuilderArchitecture.md
piotrnowakowski: No additional change here: the diagrams remain logical integration requirements, with the datateam owning the physical schema and migrations.

docs/ConceptNoteBuilderArchitecture.md
piotrnowakowski: Good question. We confirmed this should not be the CityCatalyst tenancy `Project` ID. Our proposed contract is a CNB-owned `concept_project_id` tied to the city, with an optional typed source reference for HIAP, manual, or upload origins; no document change has been made yet pending agreement.

piotrnowakowski:
@carlosgraffi Thanks. We've kept the architecture configuration-driven and composable so it can support different funder formats and both sparse and rich city contexts without being tied to one scenario.

PR #2828, opened 2026-07-10
Feat/on 6007 onboarding password validation update

cephaschapa:
## Summary
Improves the password validation UX on the signup page. The 4-item password checklist is replaced with a single contextual hint sentence that is gray on page load, turns red as the user types an invalid password, and disappears once the pattern is satisfied. Real-time mismatch detection now shows a "Passwords do not match" error on both fields as the user types, and the submit button is disabled until all inputs are valid and passwords match.
## Changes
- **`password-input.tsx`**: Replaced the 4-item checklist with a single hint sentence using the existing `password-hint` i18n key. Hint is gray by default, turns red on invalid input, and is hidden when the password satisfies the pattern. Added a `mismatch` prop that shows a red border and "Passwords do not match" message (suppressed if a form error is already shown to avoid duplication).
- **`signup/page.tsx`**: Added `mode: "onChange"` to `useForm` for real-time validity tracking. Watches both `password` and `confirmPassword` to compute `passwordsMismatch` and `passwordPatternValid`. The submit button is disabled when `!isValid || passwordsMismatch || !passwordPatternValid`.
- **`en/auth.json`** (and `de`, `es`, `fr`, `pt`): Updated `password-hint` to a full sentence and added the `passwords-dont-match` translation key.
## How to test
1. Navigate to `/auth/signup`.
2. Leave the password field empty — verify the hint appears in gray.
3. Start typing an invalid password (e.g. missing uppercase or number) — verify the hint turns red and the input border turns red.
4. Enter a valid password (8+ chars, uppercase, lowercase, number) — verify the hint disappears.
5. Enter a different value in the **Confirm Password** field — verify both fields show a red border and "Passwords do not match".
6. Verify the **Create Account** button remains disabled until all fields are valid and passwords match.
## Ticket
ON-6007
## Notes
The `mismatch` text is suppressed on the field when react-hook-form already sets an `error` on that field (e.g. after submit), preventing a duplicate message.


app/src/components/password-input.tsx
```diff
@@ -58,108 +68,43 @@ export default function PasswordInput({
...
-                hasMinLength: (value: string) =>
-                  value.length >= 8 || t("password-min-length"),
-                hasUpperCase: (value: string) =>
-                  /[A-Z]/.test(value) || t("password-upper-case"),
-                hasLowerCase: (value: string) =>
-                  /[a-z]/.test(value) || t("password-lower-case"),
-                hasNumber: (value: string) =>
-                  /[0-9]/.test(value) || t("password-number"),
-              }
-            : undefined,
```
lemilonkh, line 74: Are we still showing this feedback to the user so they know what's wrong with their chosen password? We should only show it once they have started entering text into the password field though (you can potentially use isDirty from the react hook form hook).

lemilonkh (approved):
Amazing work, thanks 👍🏻✨

app/src/components/password-input.tsx
cephaschapa: Yes feedback is shown to the user for both password validation and matching

PR #2830, opened 2026-07-13
fix(ON-6008): remove language select from signup, detect browser language

cephaschapa:
## Summary
Removes the preferred language dropdown from the signup form. Instead of asking users to explicitly select a language, the app now detects the browser's language via `navigator.language` and uses it as the `preferredLanguage` value submitted to the API. This simplifies the signup flow without changing the underlying data model or API contract.
## Changes
- Removed the visible `<LanguageSelector>` field from the signup form
- Added `resolveBrowserLanguage()` helper that maps `navigator.language` to a supported locale (`en`, `es`, `pt`, `de`, `fr`), falling back to `en` for unsupported languages
- Added a `useEffect` that sets `preferredLanguage` on mount via `react-hook-form`'s `setValue`
- Replaced the visible field with a `<input type="hidden">` so the value is still part of the form submission payload
- `LanguageSelector.tsx` is kept — it is still used in the account settings pages
## How to test
1. Navigate to `/auth/signup`
2. Confirm the preferred language dropdown is no longer visible on the form
3. Open DevTools → Network, submit the form, inspect the `/api/v1/auth/register` POST body — confirm `preferredLanguage` is present and matches your browser's language (or `en` as fallback)
4. To test fallback: temporarily override `navigator.language` in the console to an unsupported locale (e.g. `ja`) and re-submit — `preferredLanguage` should be `en`
## Ticket
ON-6008
## Notes
The API payload and Zod validation schema (`preferredLanguage: z.nativeEnum(LANGUAGES)`) are unchanged — the field is still required and validated server-side; it's just sourced from the browser instead of user input.

### DIFF
--- a/app/src/app/[lng]/auth/signup/page.tsx
+++ b/app/src/app/[lng]/auth/signup/page.tsx
@@ -7,7 +7,7 @@ import { useTranslation } from "@/i18n/client";
 import { Box, Heading, Icon, Input, Link, Text } from "@chakra-ui/react";
 import LabelLarge from "@/components/package/Texts/Label";
 import { useRouter, useSearchParams } from "next/navigation";
-import { useState, use, useEffect } from "react";
+import { useState, use } from "react";
 import { useForm, SubmitHandler } from "react-hook-form";
 import { Trans } from "react-i18next/TransWithoutContext";
 import { logger } from "@/services/logger";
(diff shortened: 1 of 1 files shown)


app/src/app/[lng]/auth/signup/page.tsx
```diff
@@ -32,6 +31,14 @@ type Inputs = {
   preferredLanguage: LANGUAGES;
 };
 
+function resolveBrowserLanguage(): LANGUAGES {
+  if (typeof navigator === "undefined") return LANGUAGES.en;
+  const base = navigator.language.split("-")[0];
+  return Object.values(LANGUAGES).includes(base as LANGUAGES)
+    ? (base as LANGUAGES)
+    : LANGUAGES.en;
+}
```
Comment (lemilonkh) [inline app/src/app/[lng]/auth/signup/page.tsx]: Can we use the currently selected language in the navigation bar's language picker instead of the browser's default language? Some people might prefer interacting with the platform e.g. in English even though their browser/ OS is set to Spanish.

lemilonkh (approved):
Great, thanks 👍🏻

app/src/app/[lng]/auth/signup/page.tsx
Comment (cephaschapa) [reply app/src/app/[lng]/auth/signup/page.tsx]: Yeah sure

PR #2835, opened 2026-07-14
ON-6003: GHGI upload AI optimization MVP — Path B/C improvements, parallel chunks, progress UX

DavidGBbr:
## Summary

Optimizes the GHGI "Upload Existing Inventory" AI import paths with faster processing (Path B: −38% time F3, −59% F4), proper error handling for Path C, and real-time chunk progress UI. Part of the AI-DLC methodology construction cycle.

## Changes

- **U1**: Skip `interpretTabular` for long-tidy/wide-year/multi-sheet adapters — 1 LLM call instead of 2
- **U2**: Adaptive table shape chunks (15–40) with `shapeTruncated` warning signal
- **U3**: Path C fail-closed — extraction errors set status `failed` + `errorLog`
- **U4**: Path B progress bar reads `extractionProgress` from backend
- **U5**: Replaced `pdf-parse` with `unpdf` (Path C PDF fix), re-measured all fixtures
- **U6**: Parallel LLM chunk execution via `asyncPool.ts` + DB mutex for progress writes
- New test fixtures (F0–F6) and 6 test suites (22 passing, 1 environment-dependent)

## How to test

1. Import a tidy/wide-year CSV via Path B — verify single LLM call, progress bar
2. Import a malformed PDF via Path C — verify `failed` status with error message
3. Import a large file (F6 fixture, 250 rows) — verify multi-chunk parallel processing
4. Run `npx jest --testPathPatterns="tests/backend/"` (22 pass, pdf-to-text needs `--experimental-vm-modules`)

## Ticket

ON-6003

## Notes

- F6 parallel gain not demonstrated due to OpenRouter timeouts at concurrency=3 with 90s LLM timeout — not a code bug
- `unpdf` test fails under Jest without `--experimental-vm-modules` flag; works correctly at runtime (verified with F5 fixture)


aidlc-docs/construction/u1-path-b-call-reduction/functional-design.md
lemilonkh, line 1: I think we might not need to keep the `aidlc-docs` folder this file is contained in as it seems to be mostly temporary information?

app/tmp-import-fixtures/01-near-ecrf-BR-RIO-2022.csv
lemilonkh, line 1: Do we need to keep the `tmp-import-fixtures` folder in the repository? The name suggests it's temporary and I don't think it's used in our jest test suite?

aws-aidlc-rule-details/common/ascii-diagram-standards.md
lemilonkh, line 1: Can we potentially remove the `aws-aidlc-rule-details` folder from the repository as well?

aws-aidlc-rules/core-workflow.md
lemilonkh, line 1: Is this file necessary for other developers? It seems otherwise we've standardized to use either Claude Code or Cursor, so it might not be necessary?

app/src/backend/PdfToTextService.ts
```diff
@@ -32,13 +52,16 @@ export async function pdfBufferToText(
...
-  const data = await pdfParse(buffer, { max: maxPages });
-  const text = typeof data.text === "string" ? data.text : "";
-  const pageCount = typeof data.numpages === "number" ? data.numpages : 0;
+  const result = await extractText(new Uint8Array(buffer), {
+    mergePages: true,
+  });
+  const pageCount =
+    typeof result.totalPages === "number" ? result.totalPages : 0;
   const truncated = pageCount > maxPages;
+  const text = normalizeExtractedText(result.text).trim();
```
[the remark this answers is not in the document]
DavidGBbr: Good catch, during the `pdf-parse` → `unpdf` swap we lost real page capping and only set the `truncated` flag. Fixed: we now extract per-page (`mergePages: false`) and slice to `maxPages` before returning text, so large PDFs can’t inflate LLM input. Added a Jest case for the truncate behavior. Thanks AI hahahah!

app/tests/backend/pdf-to-text.jest.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * Unit tests for Path C PDF→text extraction (pdf-parse via server-external load).
```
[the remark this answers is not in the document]
DavidGBbr: Agreed, leftover docstring drift from `pdf-parse`. Updated the header to `unpdf` and clarified the `--experimental-vm-modules` note. Also pointed the fixture at `tests/fixtures/` now that `tmp-import-fixtures` is out of the PR.

app/src/i18n/locales/en/onboarding.json
```diff
@@ -29,6 +29,7 @@
   "breaking-into-chunks": "Breaking file into smaller chunks for easy extraction",
   "interpreting-file": "Interpreting your file",
   "interpreting-file-description": "AI is mapping columns and extracting inventory rows. This may take a moment.",
+  "interpreting-chunk-progress": "Interpreting... (chunk {{current}} of {{total}})",
```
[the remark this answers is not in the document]
DavidGBbr: Correct. Added `interpreting-chunk-progress` to de/es/fr/pt (mirroring the existing `extracting-chunk-progress` phrasing per locale). Done in the latest push.

aidlc-docs/construction/u1-path-b-call-reduction/functional-design.md
DavidGBbr: Fair point. Those were AI-DLC Construction cycle artifacts (measurement-first process), not product runtime docs. I’ve removed `aidlc-docs/` from this PR so we keep the diff focused on the upload AI code + tests. Happy to park a short summary on the Jira ticket (ON-6003) / Linear if useful for audit trail.

app/tmp-import-fixtures/01-near-ecrf-BR-RIO-2022.csv
DavidGBbr: Agree. Those files were for the local OpenRouter measurement harness, not Jest CI hence the `tmp-` name. Removed the folder from the PR. Kept a single small PDF under `app/tests/fixtures/` for the `unpdf` unit test only.

aws-aidlc-rule-details/common/ascii-diagram-standards.md
DavidGBbr: Sure!

aws-aidlc-rules/core-workflow.md
DavidGBbr: It will only be necessary if we decide to start using AI DLC, but for now, I’ve removed it to avoid misunderstandings.

app/src/backend/PdfToTextService.ts
```diff
@@ -1,8 +1,14 @@
...
- * Uses pdf-parse; enforces max pages and size to avoid timeouts.
+ *
+ * Uses `unpdf` (serverless-friendly pdf.js) instead of `pdf-parse`. The latter's vendored
+ * pdf.js breaks when bundled by Next.js Turbopack with misleading "bad XRef entry" errors
+ * even on valid PDFs. `unpdf` is kept external via `serverExternalPackages` as a safeguard.
+ *
+ * Inputs: PDF `Buffer` (size already capped at upload via MAX_FILE_SIZE).
+ * Outputs: trimmed text (capped to maxPages), page count, truncated flag when pageCount exceeds maxPages.
  */
-import pdfParse from "pdf-parse";
```
cephaschapa, line 5: Lets ensure to remove this package also if now used anywhere in the codebase

app/src/backend/tableShapeChunking.ts
```diff
@@ -0,0 +1,70 @@
+/**
+ * Pure helpers for Path B table-shape chunk planning.
+ * Used by interpret route to size LLM shape batches without silent truncation.
+ */
+
+/** Default max chunks (legacy hard cap); adaptive planning never goes below this for small files. */
+export const BASE_MAX_TABLE_SHAPE_CHUNKS = 15;
+
+/** Hard ceiling to bound LLM cost/time on very large sheets. */
+export const ABSOLUTE_MAX_TABLE_SHAPE_CHUNKS = 40;
```
cephaschapa, line 10: I'm assuming this is the chunk window an LLM can take without timing out right?

cephaschapa (approved):
Great to have these improvements 👍. Do you think we should maintain the same LLM model or version for extraction?

app/src/backend/PdfToTextService.ts
DavidGBbr: Yes! Pdf-parse was removed from package.json and replaced by unpdf. 
Remaining mentions are only historical comments (e.g. why we switched / Turbopack note in next.config.mjs). No runtime dependency on pdf-parse remains.

app/src/backend/tableShapeChunking.ts
DavidGBbr: Not quite, `ABSOLUTE_MAX_TABLE_SHAPE_CHUNKS` is a hard ceiling on **how many shape batches** we allow for a sheet (each batch is up to `chunkSize` rows, typically 100). It’s a cost/time safety cap so large sheets don’t spawn unbounded LLM calls, and so we don’t silently drop rows like the old fixed cap of 15. It isn’t the model context window. Per-call timeouts are still governed by `LLM_TIMEOUT_MS`, and parallel in-flight calls by `LLM_CHUNK_CONCURRENCY`.

| Scenario | Rows on sheet | chunkSize | Chunks needed | **Before** (hard cap 15) | **After** (adaptive 15→40) |
|----------|---------------|-----------|---------------|--------------------------|----------------------------|
| Small (F3/F4) | 4 | 100 | 1 | 1 chunk — OK | 1 chunk — OK |
| Medium | 1 500 | 100 | 15 | 15 chunks — full coverage | 15 chunks — full coverage |
| Large | 2 000 | 100 | 20 | **15 chunks only** → rows 1501–2000 **silently dropped** | **20 chunks** → full coverage |
| Very large | 5 000 | 100 | 50 | **15 chunks** → rows 1501–5000 **silently dropped** | **40 chunks** → rows 1–4000 covered; **`shapeTruncated` + warning** for 4001–5000 |

**Summary:** before, the worst case cut everything beyond `15 × chunkSize` with no warning. Now the ceiling grows up to 40 when needed, and anything still beyond that is **explicit** (`shapeTruncated` / warning), not silently omitted.

DavidGBbr:
> Great to have these improvements 👍. Do you think we should maintain the same LLM model or version for extraction?

Yes, for this PR we intentionally keep a single shared model via LLM_MODEL (default gpt-4o-mini) for both Path B shape and Path C extract, so ops and before/after measurements stay comparable.

Choosing / splitting models (or upgrading versions) is out of scope here. We’re tracking that separately in [ON-6053](https://openearth.atlassian.net/browse/ON-6053) (Review and update models being used), which covers comparing newer models and deciding whether some tasks need a stronger/weaker model.



[ON-6053]: https://openearth.atlassian.net/browse/ON-6053?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2837, opened 2026-07-14
ON-6005 Separate CA/CC post-deploy smoke status

piotrnowakowski:
## Overview

This PR responds to feedback that CA/CC smoke failures should appear as a distinct GitHub Actions status instead of failing in Github actions.

## Milan's feedback addressed

- **Separate Actions status:** moves the CA/CC post-deploy smoke from the deployment step into a dedicated `CA/CC post-deploy smoke` job in all six web and Climate Advisor deploy workflows.
- **Deployment vs. smoke is immediately visible:** each smoke job depends on deployment. A failed rollout leaves Deployment red and smoke skipped; a successful rollout followed by an integration failure leaves Deployment green and Smoke red.
- **Kubernetes execution is preserved:** the new jobs authenticate to the same EKS cluster on a fresh runner, seed the existing fixture, wait for Climate Advisor readiness, and run the existing in-cluster smoke script.
- **Failure notifications identify the correct stage:** the web development notifier now reports `CA/CC post-deploy smoke` separately from `Deployment` when the existing Slack notification path runs.

## Changes

- Split smoke verification from deployment in the develop, test, and production workflows for both CityCatalyst web and Climate Advisor.
- Added the missing Climate Advisor development rollout wait so deployment only succeeds after the rollout is ready.
- Removed the deployment-contract test file; the workflow no longer maintains a test whose purpose is to test the test/deployment wiring.


climate-advisor/service/tests/test_deployment_contract.py
```diff
@@ -1,263 +0,0 @@
-from __future__ import annotations
```
piotrnowakowski, line 1: we no longer need a test for functioning test

lemilonkh (approved):
Awesome, thanks 👍 ✨

PR #2838, opened 2026-07-14
Fix globalapi timeout

amanda-eames:
## Summary

Fixes the cc-global-api crashing in a loop due to aggressive health check timeouts. The `/health` endpoint takes 1.5-2+ seconds to respond, but probe timeout was set to only 5 seconds, causing kubelet to mark the pod unhealthy and restart it repeatedly. This cascaded to the ALB marking all targets as unhealthy and returning 503 errors to users.

## Changes

- **livenessProbe**: Increase `timeoutSeconds` from 5 → 15 and `periodSeconds` from 10 → 15
- **readinessProbe**: Increase `timeoutSeconds` from 5 → 15, `periodSeconds` from 10 → 15, and `initialDelaySeconds` from 5 → 10
- **startupProbe**: Increase `timeoutSeconds` from 5 → 15

## How to test

1. Apply the deployment: `kubectl apply -f k8s/prod/cc-prod-global-api-deploy.yml`
2. Watch the rollout: `kubectl rollout status deployment/cc-global-api-deploy -n default`
3. Once ready, verify the API is responding: `curl -v https://api.citycatalyst.io/health`
4. Should see `200 OK` with response body `{"status":"ok"}`

## Files changed

- `k8s/prod/cc-prod-global-api-deploy.yml` - New deployment manifest with adjusted probe timeouts****


k8s/prod/cc-prod-global-api-deploy.yml
lemilonkh, line 1: Great, we might need to reference this file in `.github/workflows/global-api-tag.yml` so it gets applied when deploying to production

PR #2850, opened 2026-07-15
ON-6045: Fix AFOLU/IPPU empty columns and gas value mapping after import

cephaschapa:
## Summary

Fixes empty Activity Type/Source columns and incorrect gas value display for AFOLU (sector V) and IPPU (sector IV) after eCRF inventory import. Activity Type, Source, CO2, CH4, and N2O columns now correctly populate with data from the imported file.

## Changes

image

- **manual-input-hierarchy.json**: Added ctivityTypeField and source field definitions for sectors IV and V so the import service can map activity data correctly.
- **InventoryImportService.ts**: Added dynamic source field mapping logic for AFOLU/IPPU sectors that selects the correct source field from the schema hierarchy. Added per-gas storage in ctivityData (co2_amount, ch4_amount, n2o_amount) and debug logging for gas value tracing.
- **FormatAdapterService.ts**: Improved near-ecrf adapter gas column detection by adding exact eCRF header patterns (`ghgs (metric tonnes co2e) - co2/ch4/n2o`) to avoid ambiguous short-term matching where `co2` could match any header containing `co2e`.
- **direct-measure-table.tsx**: Removed incorrect `co2eq` fallback that was displaying totalCO2e (sum of all gases) in the CO2 column instead of actual CO2 emissions.
- **i18n locales**: Added missing `aggregate-sources-source` translation key across all locales (en, de, es, fr, pt).

## How to test

1. Upload an eCRF inventory file containing AFOLU/IPPU data
2. Approve the import
3. Navigate to a GHGI subsector page for AFOLU (e.g. V.3 Livestock)
4. Verify:
   - **Activity Type** column shows values (e.g. 'Livestock')
   - **Source** column shows values (e.g. 'Abadia de Goiás')
   - **CO2 EMISSIONS** shows 0 when no CO2 data exists in the file (not totalCO2e)
   - **CH4 EMISSIONS** shows correct values from the CH4 column
   - **N2O EMISSIONS** shows correct values from the N2O column
   - **Total Emissions** footer still shows the correct aggregate

## Ticket

ON-6045

## Notes

- Debug logging is included for gas value tracing during import (`[Import] GPC ...` and `[FormatAdapter near-ecrf] ...` console lines). These can be removed in a follow-up.
- The near-ecrf adapter previously used short search terms like `co2` which could match any header containing `co2e` (including CH4 and N2O headers). Now uses exact eCRF format patterns as highest priority.


app/src/backend/FormatAdapterService.ts
```diff
@@ -195,9 +195,12 @@ export default class FormatAdapterService {
...
     ]);
-    const co2Idx = this.col(h, ["co2 emissions", "co2"]);
-    const ch4Idx = this.col(h, ["ch4 emissions", "ch4"]);
-    const n2oIdx = this.col(h, ["n2o emissions", "n2o"]);
+    const co2Idx = this.col(h, ["ghgs (metric tonnes co2e) - co2", "co2 emissions", "- co2"]);
+    const ch4Idx = this.col(h, ["ghgs (metric tonnes co2e) - ch4", "ch4 emissions", "- ch4"]);
+    const n2oIdx = this.col(h, ["ghgs (metric tonnes co2e) - n2o", "n2o emissions", "- n2o"]);
+    console.log(
+      `[FormatAdapter near-ecrf] Gas column indices: co2=${co2Idx}, ch4=${ch4Idx}, n2o=${n2oIdx}, totalEm=${totalEmIdx}. Headers: ${h.join(" | ")}`,
+    );
```
lemilonkh: Can we use the Logger here (with info)?

lemilonkh (approved):
Awesome, thanks 👍 ✨

app/src/backend/FormatAdapterService.ts
cephaschapa: Yes yes

PR #2855, opened 2026-07-16
ON-6049: Fix third-party data disconnect not clearing activity values

tatimoreira:
## Summary

After disconnecting a third-party data source from a subsector, activity values created by that source were not being deleted (no DB-level cascade on `ActivityValue.inventoryValueId`). This left orphaned rows visible in the UI, making it appear as if the data was still connected. Additionally, the disconnect API threw a 404 on repeat attempts (non-idempotent), and errors were silently swallowed on the frontend.

## Changes

- **Backend**: DELETE endpoint now explicitly destroys all `ActivityValue` rows linked to the disconnected `InventoryValue` records before destroying the inventory values themselves, inside the same transaction
- **Backend**: Made DELETE endpoint idempotent — returns 200 instead of 404 when already disconnected
- **Frontend**: Added `"ActivityValue"` to `disconnectThirdPartyData` `invalidatesTags` so RTK Query refreshes the activity list in the UI
- **Frontend**: Added `.unwrap()` + try/catch with error toast in `ExternalDataSection` disconnect handler (errors were previously swallowed silently)
- **Frontend**: Added `showError()` call in `ActivityTab` disconnect catch block
- **i18n**: Added `disconnect-data-source-error` translation key to all 5 locales (en, de, es, fr, pt)
- **Backend**: Wrapped `applySource` and `connectAllForInventory` logic in try/catch for better error resilience

## How to test

1. Connect a third-party data source to a subsector (e.g. Stationary Energy > Fuel Consumption, Scope 1)
2. Verify the External Dataset card appears and emissions are shown
3. Click **Disconnect** on the data source card
4. Verify the success toast appears and the section switches back to the manual data entry view
5. Verify no activity rows remain visible in the emissions table
6. Click Disconnect a second time (repeat disconnect) — verify it returns success without an error

## Ticket

ON-6049

## Notes

The `ActivityValue` table has no `ON DELETE CASCADE` on `inventory_value_id`, so orphaned rows were silently left behind. The fix deletes them explicitly in the same transaction as the `InventoryValue` destroy.


app/src/components/Tabs/Activity/external-data-section.tsx
```diff
@@ -74,16 +74,24 @@ const ExternalDataSection = ({
...
+        datasourceId: inventoryValue.datasourceId,
+      }).unwrap();
+      toaster.create({
+        title: t("disconnected-data-source"),
+        type: "info",
+        duration: 5000,
+      });
+      onDisconnect?.(inventoryValue.datasourceId!);
+    } catch (_err) {
+      toaster.create({
```
lemilonkh: It might be good to log the error to the console here to make debugging easier.

lemilonkh (approved):
Amazing work, thanks 👍 ✨

PR #2858, opened 2026-07-17
fix(e2e): dismiss toast notifications before Continue button clicks

cephaschapa:
### Summary
Toast notifications rendered at bottom-end can intercept pointer events and block button clicks. Added dismissToasts helper that:
- Clicks close triggers on closable toasts
- Sets pointer-events: none on toast group containers as fallback Applied to both onboarding wizard flows where the 'inventory already exists' toast was blocking the Continue button.

### DIFF
--- a/app/e2e/helpers.ts
+++ b/app/e2e/helpers.ts
@@ -44,6 +44,28 @@ export async function dismissCookieConsent(page: Page) {
   }
 }
 
+/**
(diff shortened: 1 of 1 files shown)


lemilonkh (approved):
Great to have this utility for unit tests 👍 
We might also think about moving our toast notifications to the top of the page in the future perhaps?

PR #2859, opened 2026-07-17
Fix timeout globalapi k8s

amanda-eames:
# feat: create production-specific global-api deployment and use it in workflow

## Overview
Addressed Milan's feedback to use production-specific configuration for the global-api deployment instead of relying on dev defaults.

## Changes
- **New file:** `k8s/prod/cc-prod-global-api-deploy.yml` — Production deployment configuration with optimized probe timeout values (15s instead of default 5s)
- **Updated:** `.github/workflows/global-api-tag.yml` — Modified deployment step to reference the production config on line 82

## Why This Matters
When deploying to production, the workflow now uses the production-specific settings (extended probe timeouts) instead of the dev defaults, ensuring better reliability during production deployments.

## Testing
- Workflow will use the prod config on next tag deployment
- No functional changes to the deployment logic—only the configuration file reference

### DIFF
--- a/.github/workflows/global-api-tag.yml
+++ b/.github/workflows/global-api-tag.yml
@@ -11,6 +11,7 @@ on:
     paths:
       - global-api/**
       - k8s/cc-global-api-deploy.yml
+      - k8s/prod/cc-prod-global-api-deploy.yml
       - k8s/cc-global-api.yml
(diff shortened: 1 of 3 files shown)


lemilonkh (approved):
Great, thanks 👍 ✨

PR #2861, opened 2026-07-17
fix: cursor jump in FormattedNumberInput and layout adjustments

cephaschapa:
## Summary

Fixes a cursor jump bug in `FormattedNumberInput` where the caret would land at the second-to-last position while typing (e.g. `1,11|1` instead of `1,111|`), causing data entry errors across all numeric input fields.

## Changes

Screenshot 2026-07-17 183234
Screenshot 2026-07-17 183148

- Replaced Chakra `NumberInputRoot`/`NumberInputField` with a plain `Input` (`type="text"`, `inputMode="decimal"`) to avoid Chakra's internal cursor management overriding our position restoration
- Added `useLayoutEffect` + `pendingCursorRef` to synchronously restore cursor position after React commits DOM changes
- Added `countSeparatorsBefore()` helper to accurately compute cursor offset when formatting adds/removes thousand separators
- Fixed `min`/`max` validation silently skipping when value is `0` (`if (min && ...)` → `if (min != null && ...)`)
- Removed unused `isCharacterValid()` function
- Fixed layout widths in `ActivityDataSection` and `EmissionFactorsSection`

## How to test

1. Open any activity modal (e.g. Stationary Energy → add activity)
2. Type a number into any numeric field (activity data, emission factors)
3. Verify the cursor stays at the end while typing — no jumping to the middle
4. Verify thousand separators format correctly (e.g. `1234` → `1,234`)
5. Verify decimal input works (e.g. `1,234.56`)
6. Verify emission factor fields display correctly with their unit addons

## Notes

- Root cause: Chakra's `NumberInput` has internal value/cursor state management that fires after `requestAnimationFrame`, overriding any cursor restoration. Switching to a plain `Input` eliminates the conflict entirely.
- The `inputMode="decimal"` attribute preserves the numeric keyboard on mobile devices.

### DIFF
--- a/app/src/components/Modals/activity-modal/sections/ActivityDataSection.tsx
+++ b/app/src/components/Modals/activity-modal/sections/ActivityDataSection.tsx
@@ -56,14 +56,14 @@ export const ActivityDataSection = ({
         label={<Text truncate>{t(title)}</Text>}
         flex="2"
       >
-        <HStack>
+        <HStack w="full">
           <FormattedNumberInput
             control={control}
             name={`activity.${title}`}
             defaultValue="0"
             t={t}
             miniAddon
-            minWidth="300px"
+            w="full"
             flex={2}
           />
           {(units?.length as number) > 0 && (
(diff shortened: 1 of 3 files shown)


lemilonkh (approved):
This is really good - didn't know about useLayoutEffect yet, great job! 👍🏻✨

PR #2862, opened 2026-07-17
fix: show "No data yet" instead of 0kg when inventory has no emissions

tatimoreira:
## Summary
- When a user has an inventory but hasn't entered any emissions data yet, `totalEmissions` is `0`. Previously this was passed to `formatEmissions(0)` which rendered `0.0 kg CO2e` in the hero.
- Changed the condition from `!= null` to a truthy check so `0` falls through to the no-data fallback.
- Updated the fallback text from `N/A` to the existing `no-data-for-inventory-yet` translation key ("No data for inventory yet"), already translated in all 5 languages.
- Conditionally hides the `CO2e` unit label in the Hero when there's no real emissions data.

## Test plan
- [ ] Create an inventory with no data entered — hero should display "No data for inventory yet" (no "0kg CO2e")
- [ ] Verify that inventories with real emissions data still display correctly with value + unit + CO2e

### DIFF
--- a/app/src/components/GHGIHomePage/Hero.tsx
+++ b/app/src/components/GHGIHomePage/Hero.tsx
@@ -216,8 +216,10 @@ export function Hero({
                       >
                         <>
                           {value}{" "}
-                          {/* eslint-disable-next-line i18next/no-literal-string */}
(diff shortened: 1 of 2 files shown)


lemilonkh (approved):
Great, thanks 🙌✨

PR #2864, opened 2026-07-20
Implement CNB funder research pipeline

piotrnowakowski:
## Summary

- add a thin offline CNB funding-opportunity research CLI with input manifests and ignored per-run outputs
- add strict Pydantic request, agent-output, evidence, conflict, gap, financial-amount, review, and run-metadata contracts
- keep the research prompt in `climate-advisor/prompts/` and the Firecrawl integration in `service/app/tools/`
- run `gpt-5.6-terra` with medium reasoning, the current filled object, missing-data priorities, remaining-turn awareness, deep-project-first research, and a final no-tools gap audit
- retain Firecrawl source snapshots, field evidence, conflicts, agent trace, prompt hash, timings, and MLflow identifiers for human review
- correct the case-sensitive active MLflow experiment name and make Windows run closure encoding-safe
- document the pipeline and keep generated `climate-advisor/output/` artifacts out of Git

## Why

CNB needs a repeatable way to research one known funder and funding opportunity before approved records are imported. The pipeline must make incomplete coverage, conflicting sources, ambiguous monetary values, and provenance visible without allowing the agent to write directly to the CNB database.

The earlier loop could accept an incomplete structured response too early. This implementation treats it as a checkpoint, returns the validated current object and unresolved coverage to the model, and reserves a final turn for a systematic gap audit.

## Validation

- `uv run pytest service/tests/test_mlflow_logging.py service/tests/test_cnb_research.py service/tests/test_prompt_config.py -q` — 26 passed
- Python compilation and CLI help checks passed
- two live World Bank City Climate Finance Gap Fund runs completed as `coverage_complete`
- strongest live run captured 9 sources, 27 evidence records, one linked project/action/funding chain, and 4 explicitly typed financial amounts
- broader service suite reached 205 passed and 9 skipped; 9 existing vector-database integration tests require the unavailable local PostgreSQL instance

## Review boundary

The CLI writes local `pending_review` artifacts only. It has no CNB database credentials or write tool, and no automatic importer is included.


climate-advisor/output/cnb_research/ef602f2c-f47d-4384-b079-5fdfde085ad4/research_bundle.json
```diff
@@ -0,0 +1,701 @@
+{
```
piotrnowakowski, line 1: that is just for testing and to see output i will throw it out before merge

climate-advisor/service/app/services/cnb_research_agent.py
```diff
@@ -0,0 +1,517 @@
...
+    trace: list[AgentTurn],
+    openai_client: OpenAI,
+    model_name: str,
+    reasoning_effort: str,
+    prompt: str,
+) -> AgentLoopOutcome:
+    """Run model-selected Firecrawl turns followed by one explicit gap audit."""
+    # Initialize the validated working dossier and first-turn payload.
+    current_filled_object = request.current_filled_object or empty_result(request)
+    missing_data = find_missing_data(current_filled_object, request=request)
```
mircorudolph: Resumed dossiers calculate missing_data before reconciling prior evidence with sources captured in the current run. This can report coverage_complete, after which bundle assembly drops non-seed evidence and emits unsupported populated fields. Revalidate or recapture prior sources before determining coverage.

On a resumed run, current_filled_object may contain facts and evidence collected in an earlier run. The loop accepts that object immediately and calculates coverage from its populated fields.

Those fields satisfy coverage checks, the agent can finish as coverage_complete without re-scraping the sources that supported them.

But bundle creation later deliberately retains evidence only when its source_ref belongs to a source captured in the current run.

mircorudolph:
When running the script locally, I get a lot of 
- PydanticSerializationUnexpectedValue
and 
- WARNING mlflow.tracing.export.mlflow_v3 
issues also it does not log to mlflow at all.
I can see in mlflow the latest 2 runs (likely from you) failed/ aborted. Please check those warnings/ issues before merging ideally

mircorudolph (approved):

climate-advisor/service/app/services/cnb_research_agent.py
piotrnowakowski: fixed

piotrnowakowski:
Thanks, I reproduced this and checked it end-to-end.

There were two separate issues:

- The actual MLflow logging/context issue was fixed. We now use the MLflow 3.2-compatible trace metadata keys, and pytest tracing is disabled so tests no longer pollute the live `Clima` experiment.
- `PydanticSerializationUnexpectedValue` and the `mlflow.tracing.export.mlflow_v3` `_metrics` warning still appear with the local MLflow 3.2/OpenTelemetry combination. They are noisy compatibility warnings, but they no longer prevent persistence or truncate the trace.

I ran a completely new full research run after the fixes:

- [MLflow run](https://mlflow-dev.openearth.dev/#/experiments/4/runs/7d24a373220c4497bf9a1c73c837e077)
- [Combined trace](https://mlflow-dev.openearth.dev/#/experiments/4/traces/tr-5d1e002b2d88cac5bda971cb5356bd1d)

It finished successfully with `coverage_complete` after 14/15 turns. MLflow contains one `OK` workflow trace with all 27 spans: 14 model calls and 12 Firecrawl calls. All nine source artifacts are uploaded, all source hashes verify, and the session/request/run IDs match.

I also checked the data directly through the MLflow website and API. No required spans or artifacts are missing. The failed traces visible earlier were generated by pytest; the historical rows remain, but new test runs are now excluded from remote tracing.

So the blocking “does not log to MLflow” issue is resolved. I have not hidden the remaining warnings; they are dependency-level compatibility noise and can be handled through dependency alignment separately rather than masking them in the application.

PR #2869, opened 2026-07-21
fix: require at least one city when inviting collaborators

tatimoreira:
## Summary
- The "Send invites" button was enabled as soon as an email and project were selected, letting invites be sent with an empty city list (the email would go out with zero cities attached).
- Frontend: `invite-collaborators-step.tsx` now also requires `selectedCities.length > 0` for both the submit-button validity check and the `sendInvites` guard.
- Backend: `CreateUsersInvite.cityIds` now enforces `.min(1)`, so `POST /api/v1/user/invites` rejects an empty city list independent of the frontend state.

## Ticket
[CC-563](https://linear.app/openearth/issue/CC-563/invitation-can-be-sent-without-selecting-cities)

## Test plan
- [x] Added `app/tests/api/user_invites.jest.ts` asserting the API returns 400 when `cityIds` is empty.
- [ ] Manually verify in the invite-collaborators modal/onboarding step: with no city checked, "Enviar invitaciones" stays disabled; checking a city enables it.
- [ ] Confirm existing collaborator/admin invite flows with cities selected still work end-to-end.


app/src/util/validation.ts
```diff
@@ -193,7 +193,7 @@ export const AcceptOrganizationInvite = z.object({
 
 export const CreateUsersInvite = z.object({
   projectId: z.string().uuid(),
-  cityIds: z.array(z.string()),
+  cityIds: z.array(z.string()).min(1),
```
cephaschapa, line 196: Are we also preventing submission if cities are = 0 in the frontend?

cephaschapa (approved):
Great 👍

lemilonkh (approved):

app/src/util/validation.ts
lemilonkh: Yes 👍

lemilonkh (approved):
Awesome, thanks 👍 ✨

PR #2871, opened 2026-07-21
CC-569: Fix truncated module card titles on Dashboard

tatimoreira:
## Summary

The "Climate Risk and Vulnerability" and "Solar Project Builder" Dashboard cards truncated their titles with an ellipsis instead of showing the full name.

## Changes

- Removed the 2-line `lineClamp` on the module card title so it always renders in full.
- Increased the card's minimum height (280px → 310px) to fit longer, multi-line titles.

## How to test

1. Run `npm run dev` and open a city's Dashboard.
2. Confirm the "Climate Risk and Vulnerability Assessment" card shows the full title, not "Climate Risk and Vulnerability...".
3. Switch the app language to Spanish or Portuguese and confirm "Constructor de Proyectos Solares en Tejados" also renders in full.
4. Check that cards in the same row still line up visually.

## Ticket

CC-569

## Notes

Verified by rendering `ModuleCard` in isolation with the real long titles from the bug report; full titles now wrap onto 3 lines without clipping, and card rows remain aligned since row height is unconstrained (`minH`, not fixed `h`).


.linear.toml
```diff
@@ -0,0 +1 @@
+team_id = "CC"
```
cephaschapa: We can remove this file from linear.

cephaschapa (approved):
Nice, thanks!

PR #2882, opened 2026-07-22
[CC-512] Implement CNB similar-project search

piotrnowakowski:
## Summary

**Scope status:** This PR is the first technical overview and local prototype of similar-project matching. It demonstrates the proposed flow using reviewed artifacts and in-memory adapters. It does not implement production post-upload matching, reference-data access, workflow persistence, or user-facing integration. Those remain follow-up work.

- add a reproducible, project-anchored funded-project research CLI with required project context, optional canonical-funder enrichment, strict caller-owned batch inputs, and single-request recovery
- add reviewed research/import contracts plus a local similar-project matching prototype, prompt, reference-data interfaces, in-memory adapters, and provider-backed QA runner
- extend the static local CNB review workspace for corpus curation and prototype similar-project review
- allow non-approved review drafts to be saved before canonical-funder selection while preserving that requirement for approved reviews
- validate approved imports so selected decision values match the reviewed payload, excluded fields remain empty, and every populated database-bound project field has an affirmative decision
- harden research breadth with evidence-qualified monotonic project retention, exact target self-match exclusion, truthful completion, and bounded structured-output retries
- use OpenRouter for CNB research, local prototype matching, and funder identity while preserving the existing direct OpenAI configuration and OpenAI embedding path
- update CNB architecture and operating documentation while separating the target architecture from the prototype delivered here

## Why

The Concept Note Builder needs a reproducible way to discover and review comparable funded projects across multiple programmes. The previous local flow was centered on one known funder/program and did not provide the reviewed corpus or canonical-funder resolution needed to explore similar-project matching.

This PR establishes that reviewed corpus and demonstrates the proposed post-ingestion matching contract locally. It does not connect the matcher to production project uploads, production CNB reference data, production workflow persistence, or a user-facing product flow.

The review boundary also had two gaps: draft review states were blocked by an approval-only canonical-funder requirement, while approved import payloads were not cross-checked against their selected and excluded field decisions. The updated boundary permits incomplete drafts but rejects inconsistent approved imports.

CNB research, funder identity, and the local matching prototype use OpenRouter through the OpenAI-compatible SDK. Existing direct OpenAI compatibility and OpenAI embeddings remain unchanged.

## Impact and boundaries

- funded-project discovery always receives a substantive target-project profile; canonical-funder input is optional during discovery
- project and batch JSON inputs are caller-owned run inputs rather than checked-in project-specific data
- canonical IDs remain required at the approved reviewed import and local prototype-matching boundary
- pending, needs-changes, and rejected review artifacts can be saved without pretending canonical-funder review is complete
- `project_tags` are reviewer-curated on funded-project records; the prototype may report only supported `matched_tags` and does not invent a runtime tag vocabulary or weights
- the local matcher runs only with injected reviewed artifacts and in-memory adapters
- there is no production post-upload trigger, production reference-data client, workflow persistence adapter, or user-facing matching integration in this PR
- the browser remains local-only and never writes directly to the database
- the shipped default reference-data client safely returns no candidates

## Validation

- `uv run --directory service pytest -p no:cacheprovider tests -q -k cnb` — 43 passed
- `uv run --directory service pytest -p no:cacheprovider tests/test_cnb_review_import.py -q` — 4 passed
- `node --test scripts/cnb_research/review-support.test.js` — 3 passed
- full Climate Advisor service suite — 251 passed, 9 skipped
- one local prototype matching request completed through `https://openrouter.ai/api/v1/responses` with HTTP 200; this validates the prompt/model contract, not production integration


climate-advisor/scripts/cnb/review_samples/cbcb9402-535d-48fc-b027-344abaeab8c3.similar-projects.json
piotrnowakowski, line 1: just for testing i will throw it out

mircorudolph (commented):
Can we cluster this repo a little more?

E.g. it now contains general clima prompts and capabilities, stationary energy workflows and now CNB workflows.
CNB related things are prefixed with cnb_ mostly which is good.
However e.g. in service/app/services we have all cnb_ top level, at the same time we also have a folder stationary_energy. I suggest to also do this with CNB for services, prompts, models etc.

Otherwise with each additional agentic workflow we introduce, it will blow up quickly

climate-advisor/service/app/services/cnb/review_import.py
```diff
@@ -0,0 +1,558 @@
...
+                        row = cursor.fetchone()
+                    if row is None:
+                        raise RuntimeError("source insert did not return an ID")
+                    source_ids[source.source_ref] = UUID(str(row[0]))
+
+            # Insert reviewed funded-project rows and their source-grounded claims.
+            for project in payload.projects:
+                record = project.record
+                cursor.execute(
+                    "INSERT INTO funding_records "
```
mircorudolph, line 464: Re-running the approved importer always inserts new funding_records and evidence: run_id is never persisted or checked, and there is no natural-key/upsert guard. A retry after a timeout or an accidental second run will silently duplicate canonical reference projects, degrading every later matching result; make the import idempotent (for example, persist/import-check the source run and record refs) before inserting

climate-advisor/service/app/services/cnb/similar_project_search.py
mircorudolph, line 1: This fixed-field shortlist is the only candidate set the LLM can assess, but it compares mostly exact normalized strings before truncating the pool. Semantically equivalent inputs such as “solar PV”/“distributed renewable generation” or “municipality”/“local authority” can therefore be ranked behind weaker exact matches and never reach selection. I think this is what we try to evade with using tags however we need to make sure this appraoch here is consistent and stable. Could we introduce a controlled synonym/taxonomy mapping for these fields—or at minimum record excluded candidates and evaluate shortlist recall—before relying on this for production matching?

Also is this logic how we shortlist discussed with product? E.g. how do we decide the importance of e.g. same region etc.?
The fixed ordering implicitly decides the relative importance of category, geography, finance route, applicant type, hazards, interventions, and tags.

climate-advisor/service/app/services/cnb/similar_project_search.py
```diff
@@ -0,0 +1,623 @@
...
+        self,
+        *,
+        request: CnbSimilarProjectSearchRequest,
+        candidates: list[CnbSimilarProjectCandidate],
+    ) -> list[CnbSimilarProjectCandidate]:
+        """Keep eligible funded-project rows, defaulting to same-funder only."""
+        eligible: list[CnbSimilarProjectCandidate] = []
+        for candidate in candidates:
+            if (
+                request.funder_scope == "same_funder"
```
mircorudolph, line 211: Could we make the matching policy explicit here? At the moment, only funder scope, funded-award status, and evidence are hard eligibility gates; geography, finance route, instrument type, applicant type, hazards, and tags only influence shortlist order. That may be the right behavior, but it means a candidate with a material mismatch can still be selected by the LLM. 

Probably important to align this logic clearly with product so we are excluding/ mapping correctly. Like each field as either must match, preferred match, informational, or acceptable when unknown with a caveat, then enforce the hard gates before shortlisting.

climate-advisor/service/app/services/cnb/similar_project_search.py
```diff
@@ -0,0 +1,623 @@
...
+            "candidates": [
+                self._shortlist_payload(item) for item in shortlist
+            ],
+        }
+        logger.info(
+            "Running CNB similar-project selection for run %s with %s shortlist items.",
+            request.run_id,
+            len(shortlist),
+        )
+        response = self.openai_client.responses.parse(
```
mircorudolph, line 340: The grounded-output checks are a strong start, but I cannot see an evaluation plan showing whether the selected examples are actually useful to proposal writers. Before production integration, could we define a small expert-labelled benchmark: representative target projects, the genuinely useful funded-project comparisons for each, and measures for shortlist recall, LLM-selection precision, and reviewer agreement? We should also review plausible-but-wrong matches. This would let us calibrate the shortlist, tag taxonomy, and prompt against evidence rather than intuition.

This would be useful to check wheather our selection pipeline is actually picking up the right candidates.
Maybe something we cannot do yet if we are missing data but then lets add this as a follow up once we have enough data and decissions to assess this clearly

mircorudolph (approved):

climate-advisor/service/app/services/cnb/similar_project_review.py
```diff
@@ -0,0 +1,337 @@
...
+            )
+            grouped_candidates.setdefault(candidate.funding_record_id, []).append(
+                (candidate, research.run_id, project.record.funding_record_ref)
+            )
+            all_sources.extend(sources)
+
+    # Merge duplicate candidates while preferring the richest scalar record.
+    merged_candidates = []
+    for funding_record_id, entries in grouped_candidates.items():
+        merged_candidate = _merge_candidate_entries(entries)
```
DavidGBbr, line 340: When two records land on the same semantic key they get collapsed here with no warning. Could we at least log a caveat when `len(entries) > 1`? Silent merges are easy to miss later when the reviewed corpus grows.

climate-advisor/service/app/services/cnb_similar_project_search.py
```diff
@@ -0,0 +1,623 @@
...
+
+    def __init__(
+        self,
+        *,
+        openai_client: OpenAI,
+        workflow_store: CnbSimilarProjectsWorkflowStore,
+        model_name: str,
+        prompt: str,
+        reasoning_effort: str = "medium",
+        store_responses: bool = True,
```
DavidGBbr: `store_responses` defaults to `True` here (and in `from_settings`), but the CLI passes `False`. I'd flip the default to `False`,  same as funder identity — so we don't persist payloads to OpenRouter by accident if someone reuses the service later.

DavidGBbr (commented):
Reviewed locally too, overall this looks solid as a prototype. Two small follow-ups below (not blockers). Nice video walkthrough as well.

climate-advisor/service/app/services/cnb/similar_project_search.py
piotrnowakowski: that is a great idea i will try to do it when i will have an example from the city itself

climate-advisor/service/app/services/cnb/review_import.py
piotrnowakowski: Fixed in `749d80680`. The importer now persists the stable `(source_run_id, source_record_ref)` identity, uses a managed unique constraint plus `ON CONFLICT` to return the existing `funding_record_id` on retries, and skips duplicate source/evidence writes. Added a retry regression test and updated the managed schema/README contract; the focused CNB suite passes (`44 passed`). The externally managed CNB database still needs those two columns and the unique constraint because Climate Advisor does not own its migrations.

piotrnowakowski:
@mircorudolph Good point — I’ve made this change locally by reorganizing the CNB code into dedicated `cnb/` folders within each existing layer:

- `prompts/cnb/`
- `scripts/cnb/`
- `service/app/models/cnb/`
- `service/app/services/cnb/`
- `service/tests/cnb/`

Files inside those folders no longer need the repetitive `cnb_` prefix. This preserves the repository’s existing separation between models, services, prompts, scripts, and tests while preventing CNB modules from accumulating in the shared top-level directories.

I also updated all imports, prompt configuration, CLI paths, and documentation. No runtime behavior was intentionally changed; the focused CNB and prompt suite passes with 48 Python tests and 3 review-UI JavaScript tests.

The implementation is currently prepared locally and will appear in this PR after the changes are committed and pushed.

climate-advisor/service/app/services/cnb/similar_project_review.py
piotrnowakowski: Fixed in `f9510fd43`. When multiple reviewed records share the same semantic identity and are merged, we now emit a warning containing the merged candidate ID and every source `run_id:record_ref`. The same notice is added to the candidate's `known_gaps`, so it remains visible in the review artifact rather than existing only in logs. Added regression coverage for the warning, source references, preserved fields, and visible caveat. Focused validation passes: 50 Python tests and 3 review-UI JavaScript tests.

climate-advisor/service/app/services/cnb/similar_project_search.py
piotrnowakowski: Okey that is a valid point i will ask carlos about that similarity search what we want to focus on.
i will comment explicitly that this is first version

climate-advisor/service/app/services/cnb/similar_project_search.py
piotrnowakowski: Addressed in `cefbea357`. The V1 matching policy is now explicit in the service and README:

- hard eligibility: requested same-funder scope, a funded non-opportunity record, and retained evidence
- preferred ordering only: category, sector, geography, finance route, instrument type, applicant type, hazards, interventions, and reviewed tags
- unknown preferred fields: retain the candidate and surface a caveat

No new hard gates were introduced without Product agreement. Added regression coverage proving that preferred-field mismatches and unknown values remain eligible in V1, while missing values receive deterministic caveats. Product validation of field priority, taxonomy, and shortlist recall remains required before production reliance. Validation passes: 51 Python tests and 3 review-UI JavaScript tests.

climate-advisor/service/app/services/cnb_similar_project_search.py
piotrnowakowski: Fixed in `f9510fd43`. `store_responses` now defaults to `False` in both the constructor and `from_settings`, matching the funder-identity flow. Persistence remains available only through an explicit `store_responses=True`. Added regression coverage for both defaults and the explicit opt-in case.

PR #2883, opened 2026-07-22
docs(CC-553): native document storage architecture

DavidGBbr:
## Summary
- Adds `docs/NativeDocumentStorageArchitecture.md` as the CC-553 design deliverable.
- Maps the three city intakes (GHGI onboarding, HIAP/MEED preferences, Concept Note Builder uploads) through CityCatalyst ownership into Climate Advisor access.
- Documents Path A (structured capabilities) vs Path B (optional Markdown delivery), ownership rules, downstream decision influence, and no-breaking-change integration with the existing PDF OCR pipeline.

## Context
Addresses [CC-553](https://linear.app/openearth/issue/CC-553/create-architecture-for-agentic-native-document-storage). This is architecture only — implementation follow-ups (catalog, HIAP snapshot capability, CNB delivery, CA storage adapter / CC-570) come after review.

## Test plan
- [ ] Read the DoD Mermaid in `docs/NativeDocumentStorageArchitecture.md` (GitHub Mermaid preview)
- [ ] Confirm ownership + Path A/B match product intent with @Mirco / Piotr
- [ ] Confirm open questions section is enough to unblock follow-up tickets
- [ ] No runtime/code changes in this PR — docs only

### DIFF
--- a/docs/NativeDocumentStorageArchitecture.md
+++ b/docs/NativeDocumentStorageArchitecture.md
@@ -1,0 +1,306 @@
+# Native Document Storage Architecture
+
+> Draft for [CC-553](https://linear.app/openearth/issue/CC-553/create-architecture-for-agentic-native-document-storage) · 2026-07-22  
+> Status: **draft for stakeholder review**
+
+## One-line intent
+
+CityCatalyst owns city-provided documents and structured intakes. Climate Advisor (Clima) never holds S3 keys; it reads via **typed capabilities** and optional **Markdown delivery**, then uses that context to guide downstream suggestions.
+
+## Scope (from original product ask)
+
+Three intakes today / soon:
+
+1. **GHGI onboarding** — city uploads inventory file
+2. **HIAP / MEED** — city preferences / selections
+3. **Concept Note Builder** — city uploads supporting docs
+
+DoD: Mermaid + ownership + how Clima accesses + how inputs inform decisions → then follow-up tickets.
+
+**v1 services in diagram:** CityCatalyst app + Climate Advisor (+ hiap-meed as one compute path for preferences).  
+**`global-api`:** out of v1 unless someone shows active document traffic there (open confirm).
+
+**Related repo docs:** [ConceptNoteBuilderArchitecture.md](./ConceptNoteBuilderArchitecture.md) (CNB/OCR deep dive), [AgenticModuleScope.md](./AgenticModuleScope.md) (Stage-1 agentic scope).
+
+---
+
+## DoD diagram — target document flow
+
+```mermaid
+flowchart TB
+  subgraph Intakes["City inputs"]
+    GHGI["1. GHGI onboarding<br/>PDF / CSV / XLSX"]
+    HIAP["2. HIAP / MEED<br/>preferences + selections"]
+    CNB["3. Concept Note Builder<br/>CAP, budget, letters…"]
+  end
+
+  subgraph CC["CityCatalyst — owner of native inputs"]
+    Catalog["Native Document Catalog<br/>proposed · metadata + pointers only"]
+    S3["CC S3 bucket<br/>imports/… + pdf-ocr/results/…"]
+    OCR["PdfOcrJob queue<br/>CronJob → Mistral OCR"]
+    Prod["Product SoT<br/>inventory rows · HIAP tables<br/>MEED preference snapshots"]
+  end
+
+  subgraph Other["Other services"]
+    MEED["hiap-meed<br/>prioritize compute only<br/>not the only HIAP path"]
+    CNBDB["datateam CNB DB<br/>bundles · chapters · funder KB"]
+  end
+
+  subgraph CA["Climate Advisor / Clima"]
+    PathA["Path A — Capabilities<br/>GET structured JSON"]
+    PathB["Path B — Markdown delivery<br/>CC POSTs .md when opted in"]
+    Bundle["Context bundle<br/>selected facts + excerpts"]
+    Agents["Agents / suggestions<br/>SE · CNB · future recs"]
+  end
+
+  GHGI --> Catalog
+  HIAP --> Catalog
+  CNB --> Catalog
+
+  Catalog --> Prod
+  Catalog --> OCR
+  OCR --> S3
+  HIAP --> MEED
+  MEED -->|"ranking result"| Prod
+
+  Prod --> PathA
+  OCR -->|"CC reads result .md<br/>delivery_target = climate_advisor"| PathB
+
+  PathA --> Bundle
+  PathB --> Bundle
+  Bundle --> CNBDB
+  Bundle --> Agents
+```
+
+### How to read this (30 seconds)
+
+1. All city inputs enter **CityCatalyst** first.
+2. PDFs go S3 + OCR queue; prefs/rankings stay as **structured** product data.
+3. Clima has **two doors**: ask for JSON summaries (Path A) or receive Markdown bytes from CC (Path B). Clima does **not** read S3.
+4. Agents do not browse the bucket — they use a **context bundle** built from those doors.
+5. The **Native Document Catalog** is a proposed logical layer (API facade over existing tables first, or an explicit table later) — it is not implemented yet.
+
+---
+
+## Ownership
+
+| Input | What is stored | Owner | Clima access |
+| --- | --- | --- | --- |
+| GHGI PDF + OCR `.md` | S3 objects + `ImportedInventoryFile` + `PdfOcrJob` | CC | Path A (emissions/status) today; Path B optional later |
+| HIAP / MEED prefs | JSON snapshot + ranking/selection tables | CC (MEED = compute when used) | Path A (`hiap.summary` — to build) |
+| CNB uploads | S3 + `ConceptNoteUpload` + `PdfOcrJob` | CC | Path B → ingest → excerpts in bundle |
+| Context bundle snapshot | Run-scoped assembled context | datateam CNB DB (CA orchestrates) | Internal to CA workflows |
+| Funder / similar projects | Curated research corpus | datateam CNB DB | CNB tools (not city-native docs) |
+
+**Hard rules**
+
+1. Clima never gets S3 keys or signed URLs for source/OCR objects.
+2. Path B is **opt-in** per OCR job (`delivery_target`), not automatic for every inventory PDF.
+3. Re-upload = **new** immutable id; old row soft-deleted / superseded.
+4. No cross-DB foreign keys — only shared IDs over APIs.
+
+---
+
+## How Clima gets access
+
+```mermaid
+sequenceDiagram
+  participant City as City user
+  participant CC as CityCatalyst
+  participant OCR as PdfOcrJob / Mistral
+  participant CA as Climate Advisor
+  participant Bundle as Context bundle
+
+  Note over City,Bundle: Path A — structured capabilities default
+  City->>CC: Upload GHGI / set HIAP prefs
+  CC->>CC: Persist product SoT
+  CA->>CC: GET internal capability summary
+  CC-->>CA: Bounded JSON facts
+  CA->>Bundle: Store selected facts
+
+  Note over City,Bundle: Path B — Markdown delivery opt-in CNB
+  City->>CC: Upload CAP PDF for CNB run
+  CC->>OCR: Queue OCR
+  OCR->>CC: Write .md to S3
+  CC->>CA: POST markdown + sha256 + upload_id
+  CA->>Bundle: Register + excerpt for run
+```
+
+### Path A — capability payloads
+
+Clima calls CC internal APIs. Live GHGI examples already exist under `/api/v1/internal/ca/capabilities/ghgi/…` (for example `emissions-context`, `list-accessible`). The JSON below is **illustrative** for architecture discussion — field names may not match production responses 1:1.
+
+**GHGI emissions context (illustrative; live capability exists):**
+
+```json
+{
+  "capability": "ghgi.emissions_context",
+  "city_id": "city_msp_001",
+  "inventory_id": "inv_2024",
+  "year": 2024,
+  "status": "approved",
+  "total_emissions_tco2e": 4120000,
+  "sectors": {
+    "stationary_energy": 1800000,
+    "transportation": 1500000,
+    "waste": 820000
+  }
+}
+```
+
+Optional later enrichment (not live today) could advertise related native documents:
+
+```json
+{
+  "native_documents": [
+    {
+      "native_document_id": "ndoc_ghgi_pdf_01",
+      "source_kind": "inventory_import",
+      "label": "GHGI inventory PDF 2024",
+      "markdown_ready": true
+    }
+  ]
+}
+```
+
+**HIAP / MEED summary (target only — not wired yet):**
+
+```json
+{
+  "capability": "hiap.summary",
+  "city_id": "city_msp_001",
+  "selected_actions": [
+    {
+      "action_id": "hiap_sw_12",
+      "title": "Green stormwater infrastructure corridor",
+      "is_selected": true
+    }
+  ],
+  "strategic_preferences": {
+    "sectors": ["water", "infrastructure"],
+    "timeframes": ["near_term"],
+    "co_benefits": ["equity", "public_health"]
+  },
+  "preference_snapshot_id": "ndoc_hiap_prefs_01"
+}
+```
+
+### Path B — Markdown delivery
+
+After OCR succeeds for a CNB upload, **CC reads** the authoritative Markdown from S3 and **POSTs the bytes** to CA (no S3 key in the body). Endpoint shape already exists; production storage still returns `503 cnb_storage_unavailable` until the datateam adapter is wired.
+
+```http
+POST /v1/concept-notes/cnb_run_demo_001/uploads/upl_cap_001/markdown
+Authorization: Bearer <cc-to-ca-token>
+Content-Type: application/json
+```
+
+```json
+{
+  "markdown": "<!-- page: 1 -->\n# Minneapolis Climate Action Plan\n...\n<!-- page: 34 -->\nTarget: reduce CSO events 40% by 2030.\n",
+  "filename": "Minneapolis_CAP_2025.pdf",
+  "source_label": "Climate Action Plan",
+  "page_count": 120,
+  "sha256": "a3f1c9e8b7d64520123456789abcdef0123456789abcdef0123456789abcdef0"
+}
+```
+
+CA then keeps **excerpts** in the run bundle (not necessarily the full corpus in every prompt):
+
+```json
+{
+  "upload_id": "upl_cap_001",
+  "excerpts": [
+    {
+      "excerpt_id": "ex_34",
+      "page": 34,
+      "text": "Target: reduce CSO events 40% by 2030.",
+      "used_for": ["problem_statement"]
+    }
+  ]
+}
+```
+
+---
+
+## Downstream decisions (why this storage matters)
+
+| Input | Informs today | Informs with this architecture |
+| --- | --- | --- |
+| GHGI structured inventory | Inventory UI, HIAP inputs, CA GHGI tools | Same + SE agentic prefilling + CNB emissions context |
+| GHGI OCR Markdown | Row extraction only | Optional excerpts if product enables Path B / excerpt capability |
+| HIAP / MEED prefs | HIAP UI / prioritizer request | Any Clima skill via `hiap.summary` |
+| CNB uploads | — (not wired) | Concept note draft, evidence links, gaps |
+| Funder KB / similar projects | Research pipeline | CNB examples (curated, not city-native) |
+
+```mermaid
+flowchart LR
+  Inputs["GHGI · HIAP · CNB uploads"] --> Access["Clima Path A / B"]
+  Access --> Bundle["Context bundle"]
+  Bundle --> SE["Stationary Energy agent"]
+  Bundle --> CNBw["Concept Note Builder"]
+  Bundle --> Recs["Future CC / Clima suggestions"]
+```
+
+---
+
+## Current state vs target (short)
+
+| Intake | Now | Target |
+| --- | --- | --- |
+| GHGI PDF | S3 + `PdfOcrJob` + row extract; CA = structured Path A only | Same ownership; proposed catalog registers it; Path A first |
+| HIAP / MEED | Rankings in CC; MEED prefs often request-scoped; classic HIAP API path also exists | Durable preference snapshot in CC + Path A |
+| CNB uploads | Ingest endpoint exists; `503 cnb_storage_unavailable` | CC upload + OCR + Path B delivery + CA storage adapter |
+
+---
+
+## Integration — no breaking changes
+
+| Keep | Extend later (follow-up tickets) |
+| --- | --- |
+| `PdfOcrJob` + cron + Mistral | `concept_note_upload` resolver; inventory stays no-delivery by default |
+| GHGI capability routes | Add `hiap.summary` (+ optional markdown excerpts) |
+| CA `POST .../markdown` | Replace unavailable repo with datateam adapter |
+| HIAP / hiap-meed split | Persist MEED snapshots in **CC**, not in hiap-meed |
+
+Suggested follow-ups after approval: catalog facade/API · HIAP snapshot + capability · CNB upload + delivery · CA storage adapter ([CC-570](https://linear.app/openearth/issue/CC-570/placeholder-implementation)) · optional inventory Markdown excerpts.
+
+---
+
+## Constraints
+
+| Constraint | Note |
+| --- | --- |
+| ~20 MB PDF working cap | Ops plan needed for larger files |
+| S3 required for PDF OCR | Missing bucket → 503 |
+| Permissions | Same city/project checks as import routes |
+| Versioning | Immutable sources; new upload = new id |
+| Compliance | Follow CC file lifecycle until product/legal say otherwise |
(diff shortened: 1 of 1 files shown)


DavidGBbr:
## Note for @mirco (review ask)

Architecture draft for **CC-553** — docs only.

**Ask:** please sanity-check ownership + the two Clima access paths (Path A capabilities vs Path B Markdown delivery), especially:
1. Is `global-api` correctly out of v1?
2. Should HIAP/MEED preference snapshots live in CityCatalyst?
3. Is per-run context bundle (v1) + later city-wide reuse via a catalog the right sequencing?

Open questions are listed at the bottom of `docs/NativeDocumentStorageArchitecture.md`.

Linear: https://linear.app/openearth/issue/CC-553/create-architecture-for-agentic-native-document-storage

docs/NativeDocumentStorageArchitecture.md
```diff
@@ -0,0 +1,333 @@
...
+## Ownership
+
+| Input | What is stored | Owner | Clima access |
+| --- | --- | --- | --- |
+| GHGI PDF + OCR `.md` | S3 objects + `ImportedInventoryFile` + `PdfOcrJob` (source artifact) | CC | **Path A only today** (emissions/status from inventory). Path B / excerpts = product opt-in later, not default |
+| HIAP / MEED prefs | Selection / preference SoT in CC (exact snapshot shape **open** — see below) | CC (MEED = compute when used) | **Path A only** (`hiap.summary` — to build). Never Path B |
+| CNB uploads | S3 + `ConceptNoteUpload` + `PdfOcrJob` | CC | **Path B** → ingest → excerpts in bundle |
+| Context bundle snapshot | Run-scoped assembled context | datateam CNB DB (CA orchestrates) | Internal to CA workflows |
+| Funder / similar projects | Curated research corpus | datateam CNB DB | CNB tools (not city-native docs) |
+
```
mircorudolph, line 181: | CNB uploads | S3 + ConceptNoteUpload + PdfOcrJob | CC | Path B → ingest → excerpts in bundle |

If owner "CC" means a database inside CC for CNB then yes. CC is modular and has modules like HIAP, CNB, GHGI. Each are within CC and having their own database

docs/NativeDocumentStorageArchitecture.md
```diff
@@ -0,0 +1,333 @@
...
+    "timeframes": ["near_term"],
+    "co_benefits": ["equity", "public_health"]
+  },
+  "preference_snapshot_id": "ndoc_hiap_prefs_01"
+}
+```
+
+### Path B — Markdown delivery
+
+After OCR succeeds for a **CNB** upload, **CC reads** the authoritative Markdown from S3 and **POSTs the bytes** to CA (no S3 key in the body). Endpoint shape already exists; production storage still returns `503 cnb_storage_unavailable` until the datateam adapter is wired.
```
mircorudolph: Just a few comments:
We mention here specifically CNB for file upload. Just want to make sure that forward looking, also files could be uploaded in other modules like GHGI or so. This architecture should be extendable to those future possibilities. 

Important: If any service has been uploaded a file and its stored in S3, then this file **could** be made accesible for **all** services. However this needs scoping e.g. via the CA capability layer and context. So CA still owns which tools, context (including files), it will load.

docs/NativeDocumentStorageArchitecture.md
```diff
@@ -0,0 +1,333 @@
...
+| Versioning | Immutable sources; new upload = new id |
+| Compliance | Follow CC file lifecycle until product/legal say otherwise |
+| Cross-DB | API IDs only — no FK CC ↔ CNB DB |
+
+---
+
+## Open questions (review)
+
+1. Catalog: lean facade over existing tables first, or new `NativeDocument` table now?
+2. Should agents ever read GHGI OCR Markdown, or only structured inventory? (Default proposal: **inventory only**; PDF = source artifact / optional later reference.)
```
mircorudolph: As a default for our current state: we read the structured inventory. However this architecture should enable us to pass e.g. a pointer to the file markdown into the CA context and capability wrapper if we wanted to. This means we can later easily decide if a certain step inside CA gets access to the structured input only, or also additional information to the markdown files

docs/NativeDocumentStorageArchitecture.md
```diff
@@ -0,0 +1,333 @@
...
+3. **HIAP / MEED preference SoT in CC — how light?** Options under discussion:
+   - **A (lighter v1):** persist only on **explicit user commit / selection save**; ephemeral ranking stays re-request from MEED/HIAP APIs (avoids caching every upstream answer + heavy schema).
+   - **B:** fuller durable snapshots / history in CC for audit and city-wide reuse.
+   - Goal either way: Clima can `GET hiap.summary` without depending on request-scoped prefs alone — **not** “reproduce every MEED API payload in CC.”
```
mircorudolph: I would lean towards version A.
What is happening today:
HIAP:
A user can click "generate ranking" and after this can select up to 3 chosen actions. Also the user can generate and output plan. 
What we need to save is: 
1. the last ranking (already saved in CC)
2. The selected 3 actions (already saved in CC)
3. The generated output plan (already saved in CC)
All of these required a dedicated user action and can be seen as "finished" state that we need to capture.

For HIAP-MEED its very similar but in addition:
0a. City inputs of preferences (like I prefer short term actions, I prefer actions for stationary Energy, ....)
0b. excluded actions (a user can say which actions the user wants to exclude entirely or that are blocked by e.g. legal contraints
1-3 same as above.

This is structured data (path A) that needs to be made accesible to Clima

docs/NativeDocumentStorageArchitecture.md
```diff
@@ -0,0 +1,333 @@
...
+4. Bundle stays **per-run** in v1; catalog is the path to later city-wide reuse?
+5. Is `UserFile` BYTEA in-scope for agentic native docs?
+6. Confirm `global-api` out of v1.
```
mircorudolph: Ideally we can already include globalAPI. The globalAPI does not contain city specific data in most cases and we need to actually discuss with product and data team how and which data to use. But to me it seems like a useful ressource to have inside Clima (always scoped via capability and context layer)

docs/NativeDocumentStorageArchitecture.md
```diff
@@ -0,0 +1,333 @@
...
+7. Extra retention/audit rules beyond current CC lifecycle?
+8. If GHGI (or other) source files become visible as “we already received this,” what **product / UX** changes are needed so users are not confused (current GHGI flow was built for align-into-inventory, not future chat context)?
```
mircorudolph: This would require a new frontend page/ view. 
For CA, we could just have a tool like "summarize all my context you have" and it would list all the things we have stored and where.

docs/NativeDocumentStorageArchitecture.md
```diff
@@ -0,0 +1,333 @@
...
+
+  subgraph CA["Climate Advisor / Clima"]
+    PathA["Path A — Capabilities<br/>GET structured JSON<br/>default for GHGI + HIAP"]
+    PathB["Path B — Markdown delivery<br/>CC POSTs .md · mainly CNB"]
+    Bundle["Context bundle<br/>consumes Path A/B · not the store"]
+    Agents["Agents / suggestions<br/>SE · CNB · future recs"]
+  end
+
+  GHGI -->|"register pointer"| Catalog
+  HIAP -->|"register prefs id"| Catalog
```
mircorudolph: A pointer-only catalog can usefully include durable HIAP inputs, but the boundary should be explicit: register only committed snapshots/selections with stable domain IDs, keep the HIAP tables authoritative, and resolve content and permissions through hiap.summary. 

Since the catalog includes structured inputs as well as documents, consider naming it NativeInputCatalog rather than NativeDocumentCatalog since we also store structured data here.

This would be a reference table basically correct? That tells Clima (or other services): these are all the inputs we have received and the original data storage is at: .... ?

docs/NativeDocumentStorageArchitecture.md
```diff
@@ -0,0 +1,333 @@
...
+    GHGI["1. GHGI onboarding<br/>source file PDF / CSV / XLSX<br/>SoT for agents = inventory rows"]
+    HIAP["2. HIAP / MEED<br/>preferences + selections<br/>structured only"]
+    CNB["3. Concept Note Builder<br/>CAP, budget, letters…<br/>forward-looking"]
+  end
+
+  subgraph CC["CityCatalyst — owner of native inputs"]
+    Catalog["Native Document Catalog<br/>proposed · not implemented<br/>metadata + pointers only"]
+    S3["CC S3 bucket<br/>imports/… + pdf-ocr/results/…"]
+    OCR["PdfOcrJob queue<br/>CronJob → Mistral OCR"]
+    Prod["Product SoT<br/>inventory rows · HIAP tables<br/>preference commit snapshot TBD"]
```
mircorudolph: This box I find confusing. 

The SoT (source of truth) is the respective already existing databases inside CC like where we store GHGIs, city preferences, HIAP rankings, ...

The `NativeDocumentCatalog` or `NativeInputCatalog` (see comment below) is just a pointer reference to those. 

The workflow I would see is:
The catalog tells us what is available and where. Clima then can get access to the SoT via that catalog if enabled via capability wrappers. Is that correct?

docs/NativeDocumentStorageArchitecture.md
```diff
@@ -0,0 +1,333 @@
...
+| GHGI structured inventory | Inventory UI, HIAP inputs, CA GHGI tools | Same + SE agentic prefilling + CNB emissions context |
+| GHGI source PDF / OCR Markdown | Row extraction only (current user flow) | Catalog pointer / user reference; optional excerpts only if product enables Path B later |
+| HIAP / MEED prefs | HIAP UI / prioritizer request | Any Clima skill via `hiap.summary` (once wired) |
+| CNB uploads | — (not wired) | Concept note draft, evidence links, gaps |
+| Funder KB / similar projects | Research pipeline | CNB examples (curated, not city-native) |
+
+```mermaid
+flowchart LR
+  Inputs["GHGI · HIAP · CNB uploads"] --> Access["Clima Path A / B"]
+  Access --> Bundle["Context bundle"]
```
mircorudolph: This is exactly the feature we already have. Where CA creates a dedicated context bundle per service. 
In this step in the capability/ context wrapper, the agent decides what is being loaded into this context. This gets informed e.g. by the new pointer catalog. We need to ensure that all services stay self contained and are not getting coupled. This means e.g. CNB must not depend on e.g. inputs of other services but they can enrich that service if available. The agent via capability wrapper will build the context on what is available, scoped for that service

docs/NativeDocumentStorageArchitecture.md
```diff
@@ -0,0 +1,333 @@
...
+```
+
+---
+
+## Current state vs target (short)
+
+| Intake | Now | Target |
+| --- | --- | --- |
+| GHGI PDF | S3 + `PdfOcrJob` + row extract; CA = structured Path A only | Same ownership; proposed catalog may register the source file; **Clima stays on inventory Path A** unless product later opts into excerpts |
+| HIAP / MEED | Rankings in CC; MEED prefs often request-scoped; classic HIAP API path also exists | Clima can read prefs via Path A `hiap.summary`. **How durable the SoT is (commit-only snapshot vs re-request) is an open question** — see below |
```
mircorudolph: Just putting visibility on this that HIAP and HIAP-MEED are different services with different data storage and data schemas. So when implementing lets keep in mind they are not the same. Conceptually however they can be treated almost identical

docs/NativeDocumentStorageArchitecture.md
```diff
@@ -0,0 +1,333 @@
...
+
+  Note over City,Bundle: Path A — structured capabilities default (GHGI + HIAP)
+  City->>CC: Upload GHGI / commit HIAP prefs
+  CC->>CC: Persist product SoT (inventory rows / prefs)
+  CA->>CC: GET internal capability summary
+  CC-->>CA: Bounded JSON facts
+  CA->>Bundle: Store selected facts
+
+  Note over City,Bundle: Path B — Markdown delivery opt-in (mainly CNB; not HIAP)
+  City->>CC: Upload CAP PDF for CNB run
```
mircorudolph: Here we need to distinguish to pathes and ownerships.

If a city uploads a CAP e.g. during GHGI onboarding in the GHGI module, then this is owned by that module GHGI. Via our catalog, CNB can later access this document (e.g. reading the markdown from S3). Owner is GHGI module.

If the document gets uploaded in the CNB module later, CNB owns this document and e.g. GHGI module could use this potentially to fill our emissions etc. Owner is CNB.

This is also important to store as meta data. 

Owner should always be the module where the user uploaded it. User can be any other module that is allowed to read it (e.g. via capability wrapper).

docs/NativeDocumentStorageArchitecture.md
```diff
@@ -0,0 +1,333 @@
...
+  },
+  "preference_snapshot_id": "ndoc_hiap_prefs_01"
+}
+```
+
+### Path B — Markdown delivery
+
+After OCR succeeds for a **CNB** upload, **CC reads** the authoritative Markdown from S3 and **POSTs the bytes** to CA (no S3 key in the body). Endpoint shape already exists; production storage still returns `503 cnb_storage_unavailable` until the datateam adapter is wired.
+
+**HIAP / MEED does not use Path B** — preferences are structured Path A.
```
mircorudolph: That is correct for now. But lets keep in mind that HIAP as some point might read data from path B (currently not needed - just a future possibility)

docs/NativeDocumentStorageArchitecture.md
mircorudolph: What do we mean with `and **POSTs the bytes** to CA`?

If the document gets uploaded via CNB, then it owns the file (same as if docuemnt gets uploaded in e.g. GHGI, that module owns it). The file get stored either way in S3 (OCRd). 
CA can then get access to this markdown at any place via capability wrapper and catalog poiter. 

So wouldn't CA then rather `request` that file instead of CC posting it? Maybe just terminology in my head.

mircorudolph (approved):

docs/NativeDocumentStorageArchitecture.md
DavidGBbr: Yep, that was too vague. I've just updated:
owner = the module where the user uploaded/committed (GHGI, CNB, HIAP, …), not a generic “CC” bag. CNB owning its own DB/rows inside the modular CC setup is the mental model now.

docs/NativeDocumentStorageArchitecture.md
DavidGBbr: Makes sense, I agree. 
Doc now treats CNB as today’s main file path, but the pattern should extend to GHGI (and others). If a file is in S3, other services can read it later only via catalog discovery + CA capability/context scope. 

OBS: CA still decides what gets loaded.

docs/NativeDocumentStorageArchitecture.md
DavidGBbr: Got it. Default stays structured inventory. Architecture now explicitly allows a later step to also pull a markdown pointer into CA context when we want that.

docs/NativeDocumentStorageArchitecture.md
DavidGBbr: Went with option A. Doc now lists the finished states to expose via Path A:

HIAP: last ranking, up to 3 selected actions, output plan (already in CC).
HIAP-MEED: same + city prefs + exclusions.

That’s the SoT Clima should read, not every ephemeral API round-trip.

docs/NativeDocumentStorageArchitecture.md
DavidGBbr: Updated global-api is in scope as an optional Clima resource, always capability-scoped. Which datasets first still needs product + data; left that open.

docs/NativeDocumentStorageArchitecture.md
DavidGBbr: Prefer that for v1. Doc now leans toward a CA tool like “summarize all my context” instead of a new frontend page, unless product asks for one later.

docs/NativeDocumentStorageArchitecture.md
DavidGBbr: Yes pointer/reference table: what we got, where the real data lives, owning module. 
Renamed to NativeInputCatalog. 
For HIAP, only register committed snapshots/selections with stable IDs; HIAP tables stay SoT; resolve via hiap.summary.

docs/NativeDocumentStorageArchitecture.md
DavidGBbr: Fair, that box was muddy. Reworked the Mermaid: 
module SoTs = existing DBs; NativeInputCatalog only points at them. 
Flow: catalog says what’s available → Clima hits the SoT via capabilities when the wrapper allows.

docs/NativeDocumentStorageArchitecture.md
DavidGBbr: Yep addressed in 62ea338. Framed as the existing per-service context bundle; catalog informs availability, capability wrapper decides what loads.

docs/NativeDocumentStorageArchitecture.md
DavidGBbr: Called out in the doc: same idea for Path A, separate services/schemas/storage in implementation.

docs/NativeDocumentStorageArchitecture.md
DavidGBbr: Added owner = upload module (on the catalog row). 
Other modules can read later via capability; they don’t become owner.

docs/NativeDocumentStorageArchitecture.md
DavidGBbr: Softened that. Prefs stay Path A for now; HIAP on Path B is just a future option, not v1.

docs/NativeDocumentStorageArchitecture.md
DavidGBbr: Agreed on the mental model. Path B is now CA requesting markdown via catalog + capability; existing POST can stay as transport.

docs/NativeDocumentStorageArchitecture.md
```diff
@@ -0,0 +1,359 @@
...
+| With `NativeInputCatalog`                                                          | Without it                                                        |
+| ---------------------------------------------------------------------------------- | ----------------------------------------------------------------- |
+| One **lookup table** in CC Core: what exists, who owns it, scope, where to resolve | CA must ask each module individually                              |
+| Then request only the needed items via existing CC auth + capabilities             | No single place to list “all inputs we hold” for settings / Clima |
+
+
+Secondary benefit: the same lookup can power a future CC frontend view (e.g. settings — “all data we have about this city/org”), including manage/delete flows later.
+
+Today many flows already know the target module; the catalog matters as modules and uploads grow (e.g. CNB).
+
```
mircorudolph, line 24: I like this new clear why we need it/ goal table. Very clear

docs/NativeDocumentStorageArchitecture.md
```diff
@@ -0,0 +1,359 @@
...
+| --------------------------------------------- | -------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| **CC Core lookup table (chosen)**             | ✅        | Auth already lives in CC (JWT / module access). One register/lookup place for discovery. Modules keep SoTs; catalog stores pointers + owner + scope. Schema/migrations with Milan / Amanda. |
+| Catalog owned by Climate Advisor              | ❌        | CA should not become the system of record for “what the city uploaded.” Frontend/settings also need this list.                                                                              |
+| Separate catalog microservice with its own DB | ❌ for v1 | Risks bypassing CC auth (“query Mirco’s files without CC checks”).                                                                                                                          |
```
mircorudolph: Not sure my name should be in here as an example for files :) Better "user" etc

docs/NativeDocumentStorageArchitecture.md
```diff
@@ -0,0 +1,359 @@
...
+**Auth reminder:** A row in the catalog means “this input exists.” It does **not** grant access. CC still verifies that the CA request (on behalf of a user) may read that module/file. Capability wrappers still decide what enters a given context bundle.
+
+---
+
+
+
+## Two kinds of city data (both register)
+
+Everything discoverable by Clima should appear in `NativeInputCatalog`. We split conceptually. The product modules in play are the original three intakes (**GHGI**, **HIAP / HIAP-MEED**, **CNB**) — same story as Mirco’s ask, without a separate “scope” section.
+
```
mircorudolph, line 64: What is Mirco's ask?

docs/NativeDocumentStorageArchitecture.md
```diff
@@ -0,0 +1,359 @@
...
+**Outside this catalog:** `global-api` remains an optional Clima resource (mostly shared / non-city-specific), capability-scoped — confirm datasets with product + data.
+
+**Related:** [ConceptNoteBuilderArchitecture.md](./ConceptNoteBuilderArchitecture.md), [AgenticModuleScope.md](./AgenticModuleScope.md).
+
+---
+
+
+
+## DoD diagrams
```
mircorudolph, line 86: When you view the diagrams rendered, they cut of text. Please check the rendered diagrams 
1) Write path — persist + register
2) Read path — who consumes the catalog

docs/NativeDocumentStorageArchitecture.md
```diff
@@ -0,0 +1,359 @@
...
+  GHGI -->|"register pointer"| Catalog
+  HIAP -->|"register pointer"| Catalog
+  CNB -->|"register pointer"| Catalog
+```
+
+
+
+
+
+### 2) Read path — who consumes the catalog
```
mircorudolph, line 124: Nice addition and makes the flow clear

docs/NativeDocumentStorageArchitecture.md
```diff
@@ -0,0 +1,359 @@
...
+1. **Something durable happens in a module** → module saves its SoT **and** inserts a row in `NativeInputCatalog`.
+2. **Clima needs context** → queries the catalog, picks allowed rows (capability wrapper), then **requests** either structured JSON (**Path A**), markdown (**Path B**), or both.
+
+Clima never receives S3 keys. A catalog row only answers “what exists / where / who owns it.”
+
+### Example 1 — GHGI inventory (Path A)
+
+1. City uploads a 2024 inventory PDF in **GHGI**.
+2. GHGI runs OCR / row extract, stores the approved inventory numbers in its DB (and the file in S3 as the source artifact).
+3. GHGI **registers** a catalog row, e.g. kind=`inventory_import`, owner=`ghgi`, pointer to inventory id (+ optional file id).
```
mircorudolph: This workflow is generally correct, I would add however also a user scope. E.g. is this file scoped to a user, an inventory, a city (which could be e.g. for multiple inventories), an organization?
It becomes important that we e.g. don't make a GHGI for 2024 accessible to CNB builder for 2025 (or at least make this visible that the data is not fully matching) -> needs clarification with product team (also applied to Path B example - basically while storing any pointer, we need to store the scope of this document as well - not just the module)

docs/NativeDocumentStorageArchitecture.md
```diff
@@ -0,0 +1,359 @@
...
+
+
+Same idea for Path A; **separate schemas** in implementation.
+
+### Example 3 — CNB supporting PDF (Path B)
+
+1. City uploads a Climate Action Plan inside **CNB**.
+2. CNB queues OCR, stores markdown in S3, keeps upload metadata in its SoT.
+3. CNB **registers** a catalog row (owner=`cnb`, `markdown_ready=true`, pointer to upload id).
+4. Clima for that CNB run queries the catalog, sees the CAP row, and **requests markdown** from the CNB module.
```
mircorudolph: Not sure about this flow if then CNB reads S3 or e.g. if Clima gets a tool to read/ request that file via CC. it feels it should be a core capability of Clima (Clima still does not need to own any direct S3 keys/ connections but CC would expose anyways an endpoint to read the S3 which could be used by other services like CNB, Clima, ....) S3 upload is a core capability of CC (afaik) so it should have all those endpoints and Clima, CNB, etc would use those I assume)

mircorudolph (commented):
Now a lot clearer thanks. 

I left some additions (nothing blocking but needs some updated)

docs/NativeDocumentStorageArchitecture.md
DavidGBbr: Adjusted!

docs/NativeDocumentStorageArchitecture.md
DavidGBbr: That was sloppy wording. Meant the original CC-553 product ask (GHGI + HIAP/MEED + CNB). Updated the sentence to link CC-553 instead.

docs/NativeDocumentStorageArchitecture.md
DavidGBbr: Good catch. Shortened the Mermaid node labels and moved the detail into bullets under each diagram so GitHub doesn’t clip the text, now it's easier to read.

docs/NativeDocumentStorageArchitecture.md
DavidGBbr: Agreed. Catalog rows now explicitly carry scope (user / inventory / city / org), not only owning module including the GHGI 2024 vs CNB 2025 case. Left the exact rules as an open point for product + Milan.

docs/NativeDocumentStorageArchitecture.md
DavidGBbr: Updated. Path B is a CC core file/markdown read capability (S3 stays behind CC). Clima/CNB call that they don’t hold S3 keys or own the bucket connection. Owning module still owns the upload metadata/pointer

lemilonkh (approved):
Great approach, thanks 👍🏻

PR #2885, opened 2026-07-23
feat(ui): CC-539 Account Settings General Improvements

lemilonkh:
## Summary

Refactors account settings into general settings page

## Changes

Moves account settings to path /LNG/settings
Adjusts style and contents of user menu in NavigationBar
Hide API tokens tab for non-admin users
Adds dropdowns to team and projects tabs as temporary fix for not having the organization ID in the URL
Splits off preferences section from account details on account tab in settings
Adds apps tab to unified settings page to complete move of all functionality from old settings page
Removes old settings page and related components :tada: 

Major refactoring of settings related components and pages

## How to test

Navigate to settings in user menu and verify functionality on all tabs still works correctly

## Ticket

https://linear.app/openearth/issue/CC-539/account-settings-general-improvements


app/src/app/[lng]/settings/OrganizationSelector.tsx
```diff
@@ -0,0 +1,65 @@
...
+      onValueChange(organizations?.[0].organizationId);
+    }
+  }, [isLoading, onValueChange, value, organizations]);
+
+  return (
+    <NativeSelectRoot
+      mb={4}
+      width="300px"
+      style={{
+        fontFamily: "Poppins",
```
Comment (cephaschapa) [inline app/src/app/[lng]/settings/OrganizationSelector.tsx]: Might be worth adding this font family value into chakra. I know we have font Family of heading to represent Poppins. I wonder if we can add for another text that would use poppins I.e in this case.

cephaschapa (approved):
Awesome, great work 👍 👏 .

PR #2890, opened 2026-07-23
fix: cascade default_city_id to NULL when a city is deleted (CC-568)

tatimoreira:
## Summary
Fixes [CC-568](https://linear.app/openearth/issue/CC-568/error-popup-appears-when-removing-a-city-from-settings) — deleting a city from Settings → Projects showed a generic "Ocurrió un error" toast instead of succeeding.

**Root cause:** `User.default_city_id` has a foreign key to `City`, added in `20250724120000-add-cityid-to-user.cjs` without an `ON DELETE` clause — Postgres defaulted to `NO ACTION`. The Sequelize model (`src/models/User.ts`) already declares `onDelete: "SET NULL"` on this field, but that annotation only takes effect via `sequelize.sync()`; since the column was created through a raw-SQL migration, the real DB constraint never matched what the model claimed. So whenever any user (not just the one clicking delete) had `default_city_id` pointing at the target city, `city.destroy()` threw a `SequelizeForeignKeyConstraintError`, which fell through the API's generic error handler as a 500.

- Added a migration to set `User_default_city_id_fkey` to `ON DELETE SET NULL ON UPDATE CASCADE`, matching the sibling `default_inventory_id` / `default_project_id` columns, which already had this correctly set.
- Added a `ForeignKeyConstraintError` branch to the API error handler (`src/util/api.ts`) returning a clear 409 instead of a generic 500, so any similar gap fails loudly instead of silently in the future.

## Test plan
- [x] Ran the migration locally; confirmed `\d "User"` shows `ON DELETE SET NULL` on `User_default_city_id_fkey`
- [x] Reproduced the original bug locally with a test org/project/city where a user's `default_city_id` pointed at the target city and `default_inventory_id` was null; confirmed the delete previously failed with a FK violation
- [x] Confirmed (in a rolled-back transaction) that deleting the city post-fix succeeds and nulls out `default_city_id` instead of throwing
- [x] Manual verification through the actual UI (Settings → Projects → delete city) — confirmed working, city deletes successfully with no error toast
- [ ] Migration applied to staging as part of normal deploy pipeline


lemilonkh (approved):
Amazing work figuring this out, thanks 🙌✨

PR #2892, opened 2026-07-23
CC-588: Fix GHGI hero population display and wording

tatimoreira:
## Summary

Fixes the GHGI home page hero: total population now shows a real value instead of "N/A" when the local DB has no population row for the inventory's year, "No data for inventory yet"/"N/A" wording is unified to "No data" across all locales, and the gap between the hero content and the map is fixed at 32px.

## Changes

- `Hero.tsx`: falls back to OpenClimate's population data (already fetched for the tooltip source) via `findClosestYearToInventory` when the DB has no population entry for the city/year.
- `Hero.tsx`: population empty-state text now uses the `no-data-for-inventory-yet` key instead of `N/A`.
- `Hero.tsx`: added `gap="32px"` between the hero content column and the map.
- `dashboard.json` (en/de/es/fr/pt): `no-data-for-inventory-yet` wording changed to "No data" and its translated equivalents.

## How to test

1. Open a city's GHGI home page for a city/inventory year with no population row in the local DB but with population data available from OpenClimate (e.g. the CityCatalyst Demo org's Buenos Aires city).
2. Confirm "Total Population" shows a real number instead of "N/A", and the emissions/population empty states both read "No data".
3. Confirm there's visible spacing (32px) between the hero text content and the map, and it doesn't look cramped.

## Ticket

CC-588


lemilonkh (approved):
Great fix, thanks 👍

PR #2893, opened 2026-07-23
CC-562: Fix 404 page nav icon link to use full locale-aware path

tatimoreira:
## Summary

The 404 page's nav bar logo/title links pointed to `//cities/` (rendered by browsers as `/cities`) instead of the full locale-aware path like `/en/cities/`, because the `[...not_found]` layout hardcoded `lng=""` when rendering `NavigationBar`.

## Changes

- `[...not_found]/layout.tsx` now derives the real locale from the catch-all route's first path segment (falling back to the default language if it isn't a supported one) and passes it into `NavigationBar`, and uses it for the `<html lang>`/`dir` attributes as well.
- Added `.linear.toml` to `.gitignore` to prevent the local Linear CLI config from being committed again.

## How to test

1. Run `npm run dev` and visit `/en/some-page-that-does-not-exist`.
2. Hover/click the CityCatalyst logo or title in the nav bar — it should link to `/en/cities/` instead of `/cities`.
3. Try another locale, e.g. `/es/does-not-exist`, and confirm the link becomes `/es/cities/`.

## Ticket

CC-562


lemilonkh (approved):
Awesome, thanks 👍 ✨

PR #2894, opened 2026-07-23
CC-556: Add CityCatalyst data into CNB  context bundle

piotrnowakowski:
## Summary

Adds the run-scoped CityCatalyst context flow used by Concept Note Builder:

- adds `POST /v1/concept-notes/{run_id}/cc-context` in Climate Advisor
- validates the CC-issued user identity, run ownership, city access, and immutable run/city binding
- selects the newest accessible GHGI inventory deterministically
- returns GPC sectors I-V, sector-local completion/data-state counts, and at most five top sources
- converts CityCatalyst inventory storage from kilograms to tonnes and calculates two-decimal shares
- persists only `context_bundle.cc_context.ghgi`, preserving separately supplied MEED and the rest of the bundle under the write lock
- omits MEED by default and includes it only when the request passes `include_meed: true`
- accepts an externally supplied compact MEED snapshot with at most ten ranked actions without triggering or deleting MEED during context assembly
- rejects malformed GHGI capability payloads without persisting zero-filled snapshots
- removes the GHGI read dependency on the Stationary Energy agentic feature flag
- documents current manual city selection and the future frontend selector

Legacy HIAP is not part of this contract.

## Why

CNB needs a compact, stable city-context snapshot that can be reused throughout a concept-note run. The existing GHGI capabilities were tied to the Stationary Energy agentic flag and exposed raw database emissions as tonnes even though CityCatalyst stores them in kilograms. Climate Advisor also lacked a run/city-bound endpoint and targeted persistence adapter for this context.

## Compact local execution example

This is an **example**, not production city truth. It comes from a local Iquique execution using `include_meed: true` and a 2022 inventory populated from the checked-in MEED fixture. The inventory is correctly marked `partial` because it contains 8 of 48 required GHGI values. MEED used live dev Global API inputs and the checked-in Chile legal fixture.

```json
{
  "run_id": "af6430b9-cfd7-4009-aed3-5f545dff960a",
  "city_id": "b6a15059-ddfa-42d8-9daf-450713a86b0d",
  "context_bundle": {
    "cc_context": {
      "ghgi": {
        "availability": "partial",
        "inventory": {
          "id": "2edd677c-1ec6-4bc6-a052-a634b195f4df",
          "year": 2022,
          "type": "gpc_basic_plus",
          "gwp": "ar6"
        },
        "emissions": {
          "total_tco2e": 9076427.28,
          "sectors": ["I", "II", "III", "IV", "V"],
          "top_sources": 5
        }
      },
      "meed": {
        "availability": "available",
        "city": {
          "name": "Iquique",
          "locode": "CL IQQ"
        },
        "counts": {
          "total_actions": 102,
          "valid_actions": 82,
          "discarded_excluded": 1,
          "discarded_legal": 19,
          "ranked_actions": 10
        },
        "actions": [
          "icare_0016",
          "icare_0028",
          "c40_0010",
          "icare_0121",
          "icare_0002",
          "c40_0035",
          "ipcc_0105",
          "icare_0040",
          "c40_0012",
          "icare_0120"
        ]
      }
    }
  }
}
```

With `include_meed: false` or when the field is omitted, the response has no `meed` key. With `include_meed: true`, the local endpoint returned HTTP 200 with five sectors, five top sources, and ten MEED actions. No response contains `source_mix`, `limitations`, `removed_summary`, or `hiap`. The persisted bundle retained its existing city, project, and separately stored MEED sections.


app/tests/agentic-inventory-capabilities.jest.ts
```diff
@@ -179,8 +180,7 @@ describe("GHGI inventory internal CA capability routes", () => {
   beforeEach(() => {
     setupTests();
     process.env.CC_SERVICE_API_KEY = serviceKey;
-    process.env.NEXT_PUBLIC_FEATURE_FLAGS =
```
piotrnowakowski, line 182: we change it to make sure the agentic featre and this one are seperated

app/src/backend/agentic/ghgi/inventory/context.ts
```diff
@@ -657,6 +709,25 @@ function valueToString(value: unknown): string | null {
   return String(value);
 }
 
+function kilogramsToTonnesString(value: unknown): string | null {
```
DavidGBbr: This kg to tonnes conversion is the right fix for the CNB contract, but heads up that anything else consuming `emissions-context` (including inventory agentic tools) will see a 1000x drop in magnitude. Can we confirm those paths still make sense with real tonnes, or track a quick follow-up if prompts/evals still assume the old scale?

climate-advisor/service/app/routes/concept_note_city_context.py
```diff
@@ -0,0 +1,219 @@
...
+    try:
+        run_context = await repository.load_run_context(
+            user_id=user_id,
+            run_id=run_id,
+            city_id=payload.city_id,
+        )
+    except ConceptNoteCityContextRepositoryError as exc:
+        return repository_problem(exc)
+
+    cached_context = cached_cc_context(
```
DavidGBbr: Once GHGI is persisted we reuse it for the rest of the run. That seems intentional for CNB. Just confirming there's no planned force-refresh in this iteration, and that stale inventory mid-run is an accepted tradeoff.

climate-advisor/service/app/repositories/concept_note_city_context.py
```diff
@@ -0,0 +1,305 @@
...
+            async with self._session_factory() as session:
+                row = (
+                    await session.execute(
+                        text(
+                            """
+                            SELECT
+                                runs.user_id,
+                                runs.city_id,
+                                bundles.context_bundle
+                            FROM concept_note_runs AS runs
```
DavidGBbr, line 104: Quick ops check: this adapter expects the datateam `concept_note_runs` / `concept_note_context_bundles` schema in the DB CA talks to. Do we already have that in the target env, or should we call out the dependency in the merge notes?

DavidGBbr (commented):
Hey Piotr, overall this looks solid. Auth, run/city binding, GHGI merge under lock, and the MEED opt-in behavior all read clearly, and the tests cover the important paths.

One thing I want to double-check before merge: the kg to tonnes fix in `emissions-context` is correct for CNB, but that capability is also used by the inventory agentic tools. Those callers aren't updated in this PR, so they'll start seeing values ~1000x smaller. Have you (or anyone) sanity-checked Stationary Energy / `inventory_context_tools` against the new scale, or is that intentionally follow-up?

Also worth confirming on the ops side that the datateam CNB tables are actually present wherever we expect this endpoint to work. Without them we'll just get `503 cnb_storage_unavailable`, which is fine as a contract, but easy to miss if someone assumes merge equals ready in the environment.

On product: GHGI is snapshotted for the run and not refreshed later. That matches the docs. Just flagging so we're aligned that inventory edits mid-run won't show up in the bundle.

CodeQL's exception-exposure note on the route looks like noise given the public error mapping you already added.

mircorudolph:
> accepts an externally supplied compact MEED snapshot with at most ten ranked actions without triggering or deleting MEED during context assembly

why do we use MEED and not HIAP? MEED is the more specific Chile one. 
I suggest we use the HIAP one for now. 
In the future the MEED will likly be closer to the future HIAP version so if this is temporary I think that would be ok as well.
Also what is the reasoning behind at most 10 actions? We rank 20 actions and it could theoretically be possible that a city selects action 11 to implement.
Also a city has the possibility to select its own 3 top actions so maybe that would be the better way to a) send 3 top selected actions if available or otherwise all 20 actions?

mircorudolph:
> converts CityCatalyst inventory storage from kilograms to tonnes and calculates two-decimal shares

Is this aligned with product?
In CC our standard is kg as a base unit so all operations are easier and no accidental mismatch of units. Just flagging for visibility that we need to pay attention to not introduce any sort transformations especially if later we persist data in CNB and would store tonnes back to our database

climate-advisor/service/app/routes/concept_note_city_context.py
mircorudolph: Also cached responses bypass the CityCatalyst city-access check performed by load_ghgi_context. A run owner whose city/project access was revoked can therefore continue retrieving persisted GHGI and MEED data; revalidate city access before returning cached context.

climate-advisor/service/app/services/concept_note_city_context.py
```diff
@@ -0,0 +1,413 @@
...
+        raise ConceptNoteCityContextDataError(
+            "Inventory emissions total is missing"
+        )
+
+    status_by_sector = records_by_reference(status_data.get("by_sector"))
+    emissions_by_sector = records_by_reference(emissions_data.get("by_sector"))
+
+    sectors: list[GhgiSector] = []
+    for reference, name in SECTORS:
+        status = status_by_sector.get(reference, {})
```
mircorudolph: Missing GPC sector rows are replaced with empty mappings and persisted as zero-filled sectors. Validate that both capability responses contain each sector I–V exactly once so malformed upstream data returns invalid_cc_context instead of silently corrupting the CNB snapshot.

mircorudolph:
What is the video: `three-speaker-naming-demo.mp4`
about?

mircorudolph:
We use the MEED rankings, (just 10 see other comment). But we are not using e.g. city preferences. Makes sense because those are currently not persisted in CC itself (will be implemented in HIAP-MEED frontend migration).
Maybe we can add in the docs that this needs a later fix/update

mircorudolph:
We store the GHGI emissions in a persisted snapshot.
if at a later stage the emissions change/ update, the snapshot doesn't get updated.

Could we integrate some staleness check in CNB that would e.g. check a hash or similar (assuming CC already provides this) to see if data changed? This needs likely some discussion with product and full stack

mircorudolph (approved):
Great thanks!

piotrnowakowski:
> What is the video: `three-speaker-naming-demo.mp4` about?

codex attached wrong video XD cons of making late night PRs
ignore that one

climate-advisor/service/app/repositories/concept_note_city_context.py
piotrnowakowski: Not yet—the CNB database has not been set up in the target environment. The `concept_note_runs` and `concept_note_context_bundles` tables are expected to be provisioned there later. Until then, this endpoint will return `503 cnb_storage_unavailable`; we should treat it as an explicit deployment dependency.

piotrnowakowski:
Re: [HIAP, city selections, and the action cap](https://github.com/Open-Earth-Foundation/CityCatalyst/pull/2894#issuecomment-5108218056)

Good call. I switched the context integration from MEED to persisted HIAP data for the same inventory. It now returns every city-selected action when selections exist; otherwise it returns all persisted ranked actions, with no 10-action cap. The capability remains read-only and does not trigger or repair prioritization.

piotrnowakowski:
Re: [keeping emissions in kilograms](https://github.com/Open-Earth-Foundation/CityCatalyst/pull/2894#issuecomment-5108271640)

Agreed. I removed the kilogram-to-tonne conversion from the shared emissions contract. CityCatalyst and CNB now keep emissions in the CC base unit end to end, using explicit `kgco2e` field names, so transformed values cannot accidentally be written back as kilograms.

piotrnowakowski:
Re: [using city preferences](https://github.com/Open-Earth-Foundation/CityCatalyst/pull/2894#issuecomment-5108391605)

Good point. This is now handled by the HIAP integration instead of being left as a future documentation note. Persisted city selections take precedence, including selected unranked actions; when there is no selection, we return all persisted ranked actions. CNB does not trigger or modify HIAP.

app/src/backend/agentic/ghgi/inventory/context.ts
piotrnowakowski: Good catch. I removed the conversion at the shared capability level, so `emissions-context` continues to expose kilogram values and existing inventory-agentic consumers do not see a 1000x magnitude change. CNB now also uses explicit `kgco2e` fields, and the relevant inventory-context tests were updated.

climate-advisor/service/app/routes/concept_note_city_context.py
piotrnowakowski: @mircorudolph Fixed in the latest push. Before using cached GHGI or HIAP context, the route now revalidates live access to the city through the accessible-inventory capability. If access has been revoked, it returns `403` and does not serve the cached data. I also added regression coverage for revoked access.

climate-advisor/service/app/services/concept_note_city_context.py
piotrnowakowski: Fixed. Both the status and emissions payloads must now contain the canonical GPC sectors I–V exactly once. Missing, duplicate, or unexpected sector references return `503 invalid_cc_context`, and the malformed snapshot is not persisted. Regression coverage verifies rejection of incomplete upstream sector data.

PR #2900, opened 2026-07-27
Improve multilingual report generation reliability

mircorudolph:
## Summary

Speeds up multilingual City Action Reports while making English the canonical source for both reports and prioritization explanations. The change also prevents translated URL provenance drift, keeps report admission from blocking FastAPI's shared thread pool, removes the temporary MLflow trace-test endpoint, and documents every public API route in Swagger.

## Timing improvement

| Bilingual report comparison | Before | After | Improvement |
| --- | --- | --- | --- |
| LLM topology | 16 isolated chapter calls: 8 English + 8 Spanish | 8 concurrent English chapter calls + 1 batched Spanish translation | 44% fewer LLM calls |
| Observed wall time | About 1.7 minutes per report | **53.8s average** across 5 runs (range: 53.0-55.4s) | **47% less time / about 1.9x faster** |

Per-pod admission allows three active output-plan requests by default. Additional requests wait asynchronously for up to the configured timeout, without occupying FastAPI's shared synchronous worker pool, and receive HTTP 429 if capacity remains unavailable. In a local burst test, three reports were admitted and completed while a fourth correctly returned HTTP 429 after the configured queue timeout.

## Changes

- Generate eight canonical English report chapters concurrently, then translate the complete report once for all requested target languages.
- Generate prioritization explanations in English once, translate them with the shared vocabulary, retry invalid translation output once, and preserve English explanations with a warning if translation still fails.
- Protect URLs in chapter Markdown and limitations with placeholders before translation, restore only canonical URLs afterward, and validate both fields for provenance drift.
- Distinguish exhausted translation validation from transient provider failures: validation failures return non-retryable HTTP 502 with `error_code=report_translation_validation_failed`; transient connectivity, rate-limit, and provider-server failures return retryable HTTP 502 with `error_code=report_translation_provider_unavailable` and `Retry-After: 5`.
- Use strict JSON-schema completion handling for explanations to avoid MLflow Pydantic serialization warnings.
- Add complete Swagger/OpenAPI descriptions and HTTP response guidance for every public route.
- Remove the temporary MLflow tool-trace endpoint, configuration, deployment values, documentation, and tests.

## Frontend caller updates

- For `POST /v1/prioritize`, send `requestData.requestedLanguages` with English plus each selected non-English language. For `POST /v1/reports/output-plan`, send the same selection in `requestData.language`, with one `locode`, one ranked `actionId`, and the original prioritization snapshot.
- Treat English as canonical. The API defensively prepends `en`; therefore, a request containing only Spanish returns English and Spanish, and report chapter `title`, `markdown`, and `limitations` maps contain both language keys.
- Keep the user's selected display language as UI state. Render `rankedAction.explanations[selectedLanguage]` and `chapter.markdown[selectedLanguage]`; do not assume the submitted language list exactly matches the returned keys or use the first response language as the display selection.
- Do not request independent report or explanation generation per language; render the returned translations.
- Handle output-plan HTTP `429` as queue saturation and retry later.
- Do not automatically retry `502` responses with `error_code=report_translation_validation_failed` because the backend has already exhausted its corrected validation retry.
- Retry `502` responses with `error_code=report_translation_provider_unavailable` only after the returned `Retry-After: 5` delay.
- `/v1/mlflow/trace-test/tool-calls` has been removed; only diagnostic callers are affected.

## Verification

- `uv run pytest -q` - 246 passed
- `git diff --check` - passed
- OpenAPI inspection confirms six public routes, documented response codes, and no remaining MLflow trace-test endpoint.


hiap-meed/app/modules/prioritizer/services/report_concurrency.py
```diff
@@ -0,0 +1,43 @@
...
+
+@contextmanager
+def reserve_report_generation_slot() -> Iterator[None]:
+    """
+    Reserve one per-pod report slot or fail after the configured queue timeout.
+
+    Waiting callers hold no chapter workers or report-specific LLM resources.
+    The slot is always released when report generation finishes or raises.
+    """
+    acquired = _REPORT_SLOTS.acquire(timeout=QUEUE_TIMEOUT_SECONDS)
```
DavidGBbr: **Blocking / high:** this admission queue blocks a FastAPI/Starlette sync worker for up to `OUTPUT_PLAN_QUEUE_TIMEOUT_SECONDS` (default 120s) while waiting for a slot.

Even though waiting callers don't create chapter LLM workers, they still occupy a thread-pool worker. Under burst load that can starve sibling sync routes on the same pod (`/v1/prioritize`, exclusion preview, etc.).

Suggestion: fail fast with `429` when no slot is free (`acquire(blocking=False)`), or move waiting off the request thread pool (async semaphore / dedicated waiter executor). Keeping the concurrency cap itself looks good; the long blocking wait is the risky part.

DavidGBbr: **Blocking / high:** this admission queue blocks a FastAPI/Starlette sync worker for up to `OUTPUT_PLAN_QUEUE_TIMEOUT_SECONDS` (default 120s) while waiting for a slot.

Even though waiting callers don't create chapter LLM workers, they still occupy a thread-pool worker. Under burst load that can starve sibling sync routes on the same pod (`/v1/prioritize`, exclusion preview, etc.).

Suggestion: fail fast with `429` when no slot is free (`acquire(blocking=False)`), or move waiting off the request thread pool (async semaphore / dedicated waiter executor). Keeping the concurrency cap itself looks good; the long blocking wait is the risky part.

hiap-meed/app/modules/prioritizer/api.py
```diff
@@ -83,6 +92,18 @@
 router = APIRouter(tags=["prioritization"])
 
 
+def get_output_plan_generation_slot() -> Iterator[None]:
+    """Hold one per-pod report slot for the lifetime of an output-plan request."""
+    try:
+        with reserve_report_generation_slot():
```
DavidGBbr: Same concern as in `report_concurrency.py`: because this `Depends` wraps the blocking semaphore wait, each queued output-plan request holds a sync worker for the lifetime of the wait (up to the queue timeout) before generation even starts.

Prefer failing fast with `429` when no slot is available, or waiting outside the shared sync request thread pool.

hiap-meed/app/modules/prioritizer/services/report_translation.py
```diff
@@ -0,0 +1,349 @@
...
+    for chapter in canonical_chapters:
+        placeholders: dict[str, str] = {}
+
+        def replace_url(match: re.Match[str]) -> str:
+            """Return the next stable placeholder for one canonical URL."""
+            placeholder = f"[[URL_{chapter.key.upper()}_{len(placeholders) + 1}]]"
+            placeholders[placeholder] = match.group(0)
+            return placeholder
+
+        protected_markdown = URL_PATTERN.sub(replace_url, chapter.markdown)
```
DavidGBbr, line 200: **Request change:** URL protect/restore/validate currently covers `markdown` only.

`limitations` are still sent to the translation LLM as plain text, and `_validate_translated_content` only compares URLs found in markdown. A translated `limitations[]` entry could therefore introduce or alter `https://...` links without failing validation, which undercuts the "URLs are not model-generated" guarantee in the README.

Please either:
1. run the same placeholder protect/restore on each limitation string, and validate URL multisets across `markdown + limitations`, or
2. explicitly reject any URL that appears in translated limitations but not in the canonical chapter.

hiap-meed/app/modules/prioritizer/services/report_translation.py
```diff
@@ -0,0 +1,349 @@
...
+    if len(translated_limitations) != len(canonical.limitations):
+        raise ValueError(
+            f"Translated chapter `{canonical.key}` changed limitation coverage"
+        )
+    validate_generated_language(
+        "\n".join([translated_markdown, *translated_limitations]),
+        language,
+        content_label=f"Translated chapter `{canonical.key}`",
+    )
+    canonical_urls = URL_PATTERN.findall(canonical.markdown)
```
DavidGBbr: Related to the protect/restore gap above: this check only looks at markdown URLs, so provenance drift inside `translated_limitations` would still pass validation.

Please include `limitations` in the URL multiset comparison (or forbid unexpected URLs there).

hiap-meed/app/modules/prioritizer/orchestrator.py
```diff
@@ -1201,13 +1202,45 @@ def run_prioritization(
...
+                    for language in explanation_languages
+                    if language != "en"
+                ]
+                translation_llm_io: dict[str, object] = {
+                    "status": "skipped",
+                    "reason": "no_target_languages",
+                }
+                if target_languages:
+                    translations_by_action_id, explanation_warnings, translation_llm_io = (
+                        translate_explanations(
```
DavidGBbr: **Nit / follow-up:** report translation retries once on validation failure, but explanation translation is a single shot.

If `translate_explanations` fails transiently (schema/language drift), the orchestrator clears `explanations_by_language` and the prioritize response comes back with ranking but no explanations even though the canonical English batch already succeeded.

Consider one validation retry here (same pattern as the report path), or failing open by returning English-only explanations with a warning instead of dropping all explanation text.

hiap-meed/app/modules/prioritizer/api.py
```diff
@@ -754,6 +823,27 @@ def generate_output_plan(
...
+                error_message=str(error),
+                status_code=502,
+            )
+            raise HTTPException(
+                status_code=502,
+                detail={
+                    **_error_payload(request_trace_id, str(error)),
+                    "retryable": True,
+                },
+                headers={"Retry-After": "1"},
```
DavidGBbr: Minor question: marking translation validation failures as `retryable: true` with `Retry-After: 1` may cause clients to immediately replay the same request and hit the same structural failure again (wasted LLM spend + queue pressure).

Is the intent only for rare model flukes? If so, a slightly larger backoff or distinguishing flaky vs persistent validation errors might be safer for frontend retry logic.

hiap-meed/tests/unit/test_report_translation.py
```diff
@@ -0,0 +1,184 @@
...
+        target_chapter_inputs=target_inputs,
+    )
+
+    assert list(translated) == ["es"]
+    assert [chapter.title for chapter in translated["es"]] == [
+        "Resumen",
+        "La acción",
+    ]
+    assert translated["es"][0].source_refs == ["city"]
+    assert "https://correct.example" in translated["es"][0].markdown
```
DavidGBbr: FYI: CodeQL flagged this as "Incomplete URL substring sanitization". Looks like a false positive on a unit-test assertion rather than sanitization logic. Worth dismissing/justifying so the check doesn't block merge.

DavidGBbr (changes requested):
Nice Job! I loved the documentation and the results

Just a few points:
1. **Blocking:** the per-pod admission wait can hold a sync HTTP worker for up to 120s and risk starving other routes on the same pod.
2. **Must-fix:** URL provenance protection doesn't cover `limitations`, only chapter markdown.
3. **Follow-up:** explanation translation has no retry, unlike the report translation path.

hiap-meed/tests/unit/test_report_translation.py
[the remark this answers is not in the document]
mircorudolph: This is a unit-test assertion checking whether the expected URL occurs in generated Markdown; it is not URL sanitization.

hiap-meed/app/modules/prioritizer/services/report_concurrency.py
mircorudolph: Fixed by moving report admission to an async bounded semaphore. Queued output-plan requests now suspend without consuming FastAPI’s shared thread pool, while preserving the existing three-active-report limit, 120-second timeout, and HTTP 429 behavior. Added coverage for both timeout/reuse and admission when a slot becomes available.

mircorudolph: fixed

hiap-meed/app/modules/prioritizer/api.py
mircorudolph: fixed

hiap-meed/app/modules/prioritizer/services/report_translation.py
mircorudolph: Fixed by extending URL placeholder protection, restoration, and validation to both chapter Markdown and limitations. Canonical URLs are now removed from all model-visible chapter fields, each placeholder must be preserved exactly once, and restored URLs are validated across markdown + limitations. Added coverage confirming limitation URLs are restored and model-generated limitation URLs are rejected. Updated the translation prompts and README accordingly

hiap-meed/app/modules/prioritizer/services/report_translation.py
mircorudolph: fixed

hiap-meed/app/modules/prioritizer/orchestrator.py
mircorudolph: Fixed both failure paths. Explanation translation now retries invalid schema, action/language coverage, or wrong-language output once with a focused correction. If translation still fails, the orchestrator preserves the successful canonical English explanations and returns them with generated_languages: ["en"] plus a translation warning instead of clearing all explanation text. Added unit and integration coverage.

hiap-meed/app/modules/prioritizer/api.py
mircorudolph: Fixed by distinguishing exhausted translation validation from transient provider failures. Validation failures already receive one internal corrected retry; if both attempts fail, the API now returns 502 with error_code: report_translation_validation_failed, retryable: false, and no Retry-After, preventing clients from replaying the full report request. Transient provider connectivity, rate-limit, and server failures return error_code: report_translation_provider_unavailable, retryable: true, and Retry-After: 5. Service and API contract tests were added.

hiap-meed/tests/unit/test_report_translation.py
mircorudolph: Yea its a false positive. Thanks!

DavidGBbr (approved):
Approved, good job!

PR #2904, opened 2026-07-28
docs: Terra UI ↔ code parity audit

carlosgraffi:
Compares the **CC Terra UI Guidelines** Figma library against the Chakra v3 theme in `app/src/lib/theme/`. No such comparison existed before.

**Docs only. No application code touched.**

## What's here

| File | What it is |
|---|---|
| `docs/terra-ui-parity/index.html` | Self-contained page comparing Figma to code. Open it directly — no build, no server. |
| `docs/terra-ui-parity/README.md` | Findings, plus a proposal for a guarded in-app version. |

## Good news first

All **25** Terra UI colour tokens exist in the theme under corresponding names. The vocabulary is genuinely shared — alignment is bookkeeping, not a rewrite. Colour sits at 93% token adoption, typography at 94%.

## Findings worth a look

**19 references point at font-size tokens that don't exist** — and most are Figma's style names typed verbatim. Figma names its styles `Body/large`, `Body/extra-large`, `Label/large`; the theme abbreviates to `body.lg`, `body.xl`, `label.lg`. `fontSize="body.large"` appears 6 times (all six delete modals). Chakra drops an unresolvable token silently, so the text renders at browser default — no build error, no lint rule, nothing in review catches it.

**`content.tertiary` is defined twice with different values.** A raw token at `#7A7B9A` and a semantic token resolving to `#4B4C63`. Semantic wins, so `#4B4C63` renders — but the raw line is what you find when you search the file, and it's annotated `// validated`. It's the most-used colour in the product (377 refs), and `#7A7B9A` is hardcoded ~80 times in `icons.tsx`, so no icon follows the token if it moves.

**`semantic.*` duplicates `sentiment.*`.** Six of seven values are byte-identical. The seventh disagrees: `semantic.warning` is `#C98300`, `sentiment.warningDefault` is `#F9A200` — two warning colours depending on which name you reach for. 44 usages on the namespace design has never seen.

**Breakpoints were implemented, then commented out.** `app-theme.ts:694–702` holds a five-step scale whose boundaries — 600, 905, 1240, 1440 — match Terra UI's grid styles exactly, inside a block comment. No history explains why.

**Spacing sits at 5% adoption** — 63 token uses against 1,117 hardcoded `px` literals. The scale is agreed on both sides and used by neither.

**Terra UI ships duplicate token collections.** Spacing and radius each exist twice in Figma: a semantic set that maps 1:1 to the theme, and a numeric `Primitives` set that maps to nothing. One is presumably legacy.

**Part of the component library is Chakra v2.** Switch and Checkbox carry `colorScheme=blue|cyan|green|pink|purple|teal` — a prop that doesn't exist in v3, where the theme uses `colorPalette: "brand"`. Alert uses `variant=left-accent|solid|subtle|top-accent` against a codebase with no Alert component. Table uses `variant=Simple|Striped|Unstyled` against an in-house `data-table`. These read as inherited from a Chakra v2 Figma kit and can't be handed off as specified — that needs a design decision, not a code one.

**Design specifies disabled everywhere; code implements it nowhere.** Figma models `State=Disabled` across tabs, inputs, checkboxes and switches. Every recipe in the theme has 10 `_hover`, 6 `_active`, 5 `_loading` and zero `_disabled`.

## Scope and limits

- Scoped to the **CityCatalyst brand theme**. The five other `data-theme` palettes are per-organization white-labelling, deliberately out of scope.
- Parity is **structural** — names and shape, not values. The Figma REST search returns style names and descriptions but not resolved hex, so a green row means the token exists on both sides, not that the hex agrees. Confirming values needs a Desktop Bridge extraction. Code-side values are read from `app-theme.ts` and are accurate.
- The component section is a structural read of variant APIs sampled from the guidelines page, not a visual diff.
- This page is hand-authored and does not track the codebase. It will go stale — see the proposal in the README for a guarded in-app route that wouldn't.

## Review notes

Nothing here changes behaviour, so there's nothing to test. The useful review is whether the findings match your understanding, and whether the Chakra v2 components in Terra UI should be rebuilt or retired.

### DIFF
--- a/docs/terra-ui-parity/README.md
+++ b/docs/terra-ui-parity/README.md
@@ -2,6 +2,8 @@
 
 Open [`index.html`](./index.html) in a browser. No build, no server — it's self-contained.
 
+There's also a [hosted copy](https://claude.ai/code/artifact/3de11a51-9f25-4bf5-b73c-0e87849826f0) for anyone who'd rather not clone the repo. It's access-controlled and shared with the Open Earth team — if it asks you to sign in and you're outside the team, use the local file instead. It's a point-in-time snapshot, so treat `index.html` in this repo as the source of truth.
+
 ## What this is
 
 A structural comparison between the **CC Terra UI Guidelines** Figma library and the Chakra v3 theme in `app/src/lib/theme/`. It exists to answer three questions we couldn't answer before:
--- a/docs/terra-ui-parity/README.md
+++ b/docs/terra-ui-parity/README.md
@@ -1,0 +1,69 @@
+# Terra UI ↔ CityCatalyst parity
+
+Open [`index.html`](./index.html) in a browser. No build, no server — it's self-contained.
+
+## What this is
+
+A structural comparison between the **CC Terra UI Guidelines** Figma library and the Chakra v3 theme in `app/src/lib/theme/`. It exists to answer three questions we couldn't answer before:
+
+1. Which design tokens exist on both sides, and under what names?
+2. Which are defined and never used — in either direction?
+3. Where does the handoff actually break?
+
+Scoped to the **CityCatalyst brand theme**. The five other `data-theme` palettes are per-organization white-labelling — a product feature, not a design-system concern.
+
+## What it found
+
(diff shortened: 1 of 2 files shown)


lemilonkh (approved):
Amazing, thanks 👍 ✨

PR #2906, opened 2026-07-28
CC-516: update account settings and add organization plan badges

cephaschapa:
## Summary

Refresh the account settings experience with improved layout, save-state behavior, and a profession dropdown, and add end-to-end organization plan type support with plan badges and usage details.

## Changes

image

* Update the settings page layout and account tab copy, and disable save until account details actually change.
* Replace the free-text position field with a localized profession dropdown, including admin-only visibility for the Open Earth Foundation Admin option.
* Add organization plan_type and 	rial_ends_at support across the migration, model, API, admin forms, and settings/org profile plan details UI.
* Add trial/demo/full badges, usage calculations, translations, and disabled button theme styling.

## How to test

1. Run cd app && npm run db:migrate.
2. Open the settings page and verify the updated account layout, descriptive copy, and preferences info banner behavior.
3. Confirm the account details save button is disabled on load, enables after editing name or profession, and disables again after a successful save.
4. Verify the profession dropdown shows the correct localized options and that only OEF admins can see the Open Earth Foundation Admin option.
5. Create a new organization as an admin and confirm a plan type can be selected.
6. Edit an existing organization and confirm the plan type can be updated from the organization profile page.
7. Verify Trial plans show remaining days, while Demo and Full show the correct badge styling.
8. Verify plan details show projects, cities in use, and remaining city slots correctly.

## Ticket

[CC-516](https://linear.app/openearth/issue/CC-516/account-settings-update-styles-and-content-of-the-account-tab)

## Notes

* This PR includes a database migration: pp/migrations/20260727140000-add-organization-plan-type.cjs.


app/migrations/20260727140000-add-organization-plan-type.cjs
```diff
@@ -0,0 +1,29 @@
...
+    await queryInterface.addColumn("Organization", "trial_ends_at", {
+      type: Sequelize.DATE,
+      allowNull: true,
+      comment: "When the trial plan expires (trial plan only)",
+    });
+
+    await queryInterface.bulkUpdate("Organization", {
+      plan_type: "full",
+      trial_ends_at: null,
+    });
```
lemilonkh: Aren't these the default values set by the above addColumn calls? Do we need this bulkUpdate step?

app/src/app/[lng]/settings/account/AccountDetailsTab.tsx
```diff
@@ -61,7 +62,14 @@ const AccountDetailsTab: FC<AccountDetailsFormProps> = ({
...
       payload.title = data.title;
     }
-    await setCurrentUserData(payload).then(() => showSuccessToast());
+    await setCurrentUserData(payload).then(() => {
+      showSuccessToast();
+      reset({
+        name: data.name,
+        email: data.email,
+        title: data.title ?? "",
+      });
```
Comment (lemilonkh) [inline app/src/app/[lng]/settings/account/AccountDetailsTab.tsx:71]: Do we need to reset here? These are the current values of the inputs, right?

app/src/components/PlanBadge.tsx
```diff
@@ -0,0 +1,88 @@
...
+  [OrganizationPlanType.TRIAL]: {
+    background: "#24BE00",
+    labelKey: "trial-plan",
+  },
+  [OrganizationPlanType.DEMO]: {
+    background: "#232640",
+    labelKey: "demo-plan",
+  },
+  [OrganizationPlanType.FULL]: {
+    background: "#2351DC",
```
lemilonkh, line 26: Do we have these colors in the theme? If not, no problem.

lemilonkh (approved):
Amazing work, thanks 🎉

app/src/app/[lng]/settings/account/AccountDetailsTab.tsx
Comment (cephaschapa) [reply app/src/app/[lng]/settings/account/AccountDetailsTab.tsx:71]: Yes they are.

app/migrations/20260727140000-add-organization-plan-type.cjs
cephaschapa: Good catch. The bulkUpdate is unnecessary here since plan_type already gets the default "full" from the column definition, and trial_ends_at is nullable so existing rows are already NULL. I’ll remove that backfill step.

PR #2907, opened 2026-07-28
fix: docs page not having API routes in built Docker image

lemilonkh:
Fixes docs page being empty in deployed builds
Also fixes title being simply "Next.js"


cephaschapa (approved):
Great work 👍 👍👍

PR #2908, opened 2026-07-28
feat(ui): CC-541 rework projects tab on settings

lemilonkh:
Updated header:

Now instead of {project} > {city} we separate project and city in a different line.

Add inventory: Organization admins view only

Updated layout:

Removed tabs since collaborators are already in Teams.

Table label: Should be Inventories

Status label content is now secondary color

Added search projects feature and created card around projects tab content

Allow scrolling project list


cephaschapa (approved):
Amazing work 👏,  refreshing to see new changes.

PR #2921, opened 2026-07-29
CC-514: replace next-runtime-env with custom runtime env

cephaschapa:
## Summary

Replace 
ext-runtime-env with an in-repo runtime env implementation so NEXT_PUBLIC_* values can be injected at request/start time instead of being locked to build-time values.

## Changes

- add a small @/lib/runtime-env module with an allowlist, env() helper, and injected window.__ENV script
- swap existing 
ext-runtime-env imports to the new helper and render RuntimeEnvScript from the root layout
- remove the unused authorize-layout import and drop 
ext-runtime-env from app dependencies

## How to test

1. In app, run 
pm run dev and open the app.
2. Confirm window.__ENV exists in the browser console and includes the allowlisted NEXT_PUBLIC_* keys.
3. Build with one NEXT_PUBLIC_FEATURE_FLAGS value, then run 
pm run start with a different value and confirm the running app reflects the start-time value.
4. Verify feature-flagged UI, analytics initialization, and OpenClimate API calls still read the expected runtime values.

## Ticket

CC-514

## Notes

This keeps runtime public env behavior compatible with the k8s deployment model without relying on next-runtime-env, which is stale for the current Next.js/React stack.


app/src/lib/runtime-env/keys.ts
```diff
@@ -0,0 +1,23 @@
+/** NEXT_PUBLIC_* keys exposed to the browser at runtime via window.__ENV */
+export const PUBLIC_RUNTIME_ENV_KEYS = [
+  "NEXT_PUBLIC_FEATURE_FLAGS",
+  "NEXT_PUBLIC_OPENCLIMATE_API_URL",
+  "NEXT_PUBLIC_POSTHOG_KEY",
+  "NEXT_PUBLIC_POSTHOG_HOST",
+  "NEXT_PUBLIC_DEPLOYMENT_ENV",
```
lemilonkh, line 7: I think we are using more NEXT_PUBLIC env vars, like e.g. NEXT_PUBLIC_SUPPORT_EMAILS. Would be good to search through the codebase and our GitHub actions workflows to see which are used.

app/src/lib/runtime-env/keys.ts
```diff
@@ -0,0 +1,23 @@
...
+] as const;
+
+export type PublicRuntimeEnvKey = (typeof PUBLIC_RUNTIME_ENV_KEYS)[number];
+
+export function getPublicRuntimeEnv(): Record<PublicRuntimeEnvKey, string> {
```
lemilonkh, line 22: It seems this is running in the frontend from the RuntimeEnvScript.tsx file? Does it provide any advantage over using the env directly then?

If it is indeed running as a server component (where the env vars would work as we want), maybe having a "use server" directive at the top would ensure that it's working correctly as intended?

app/src/lib/runtime-env/RuntimeEnvScript.tsx
```diff
@@ -0,0 +1,14 @@
+import { getPublicRuntimeEnv } from "./keys";
```
lemilonkh, line 1: Maybe add a "use server" directive at the top of this file to ensure it works correctly?

app/src/lib/runtime-env/keys.ts
cephaschapa: Good catch — I searched the app + GitHub Actions and expanded the allowlist. Client-facing keys that need runtime injection are now:

NEXT_PUBLIC_FEATURE_FLAGS
NEXT_PUBLIC_OPENCLIMATE_API_URL
NEXT_PUBLIC_POSTHOG_KEY / NEXT_PUBLIC_POSTHOG_HOST
NEXT_PUBLIC_DEPLOYMENT_ENV
NEXT_PUBLIC_SUPPORT_EMAILS
NEXT_PUBLIC_CC_CCRA_REPLIT_URL
NEXT_PUBLIC_HIGHLIGHT_PROJECT_ID

app/src/lib/runtime-env/RuntimeEnvScript.tsx
cephaschapa: RuntimeEnvScript is a Server Component (no "use client") that returns a <script> tag. Adding "use server" would treat its exports as Server Actions

app/src/app/[lng]/auth/layout.tsx
```diff
@@ -30,7 +30,7 @@ export default function AuthLayout(props: {
         <Box w="full">
           <Box pt={148} pb={4} w="480px" maxW="full" mx="auto" px={4}>
             <Toaster />
-            {props.children}
+            <Suspense fallback={null}>{props.children}</Suspense>
```
Comment (lemilonkh) [inline app/src/app/[lng]/auth/layout.tsx:33]: In some places we could potentially remove feature flags that are always on now (e.g. go to market, OAuth) so we don't need Suspense (in all of the places at least) since it prevents server side pre-rendering of components.

Fine for now though.

lemilonkh (approved):
Amazing, thanks! 🎉🎊✨

PR #2922, opened 2026-07-29
feat(ci): send Slack notifications for successful deployments

lemilonkh:
Send Slack notifications in #releases channel for deployments to both test and prod


cephaschapa (approved):
Great, thanks 🚀

PR #2923, opened 2026-07-29
feat(ui): CC-517 hide brand settings section for non-org-admin users

lemilonkh:
Hide brand settings for normal users
Also allows choosing organization (using the dropdown for now)
Fixes some hydration errors

### DIFF
--- a/app/src/app/[lng]/settings/account/BrandSettingsTab.tsx
+++ b/app/src/app/[lng]/settings/account/BrandSettingsTab.tsx
@@ -29,6 +29,7 @@ import { trackEvent } from "@/lib/analytics";
 import { api } from "@/services/api";
 import { logger } from "@/services/logger";
 import LogoUploadCard from "./logo-file-upload";
+import { OrganizationResponse } from "@/util/types";
 
 const KeyColorMapping = {
   blue_theme: "#001EA7",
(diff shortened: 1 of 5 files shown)


tatimoreira:
🔴 **Security: UI-only gating — backend endpoint has no authorization check**

The whole point of this PR is access control, but the endpoint it protects, `branding/route.ts`, has **no authorization check at all**. Any authenticated user can `PATCH /api/v1/organizations/{id}/branding` for *any* organization ID directly, bypassing the new UI restriction entirely.

Sibling routes already have the fix pattern available:
- `organizations/[organization]/route.ts` PATCH calls `UserService.validateIsAdmin(session)`
- `organizations/[organization]/users/route.ts`, `invitations/route.ts`, etc. call `UserService.validateIsAdminOrOrgAdmin(session, organizationId)` — exactly the semantics needed here.

This isn't code touched by this PR, but since the PR's premise is "hide brand settings for non-admins," it's worth flagging that the real access control is still missing server-side. Recommend adding `validateIsAdminOrOrgAdmin` to the branding route as a follow-up (ideally before/alongside this PR).

tatimoreira (changes requested):
I let a comment on the PR (backend endpoint has no authorization check)

tatimoreira (approved):

PR #2924, opened 2026-07-29
CC-544: Redesign sidebar navigation drawer

tatimoreira:
## Summary

Redesigns the JN sidebar drawer navigation: adds an organization switcher, restyles the project/city filters, adds a capped "All tools" module list, and restores the "Learning hub" link, with several alignment and styling fixes along the way.

## Changes

- Add an organization switcher header (with dropdown for multi-org users) to the JN drawer and legacy project drawer, reusing the navbar's briefcase icon treatment
- Restyle the Project/City selects as bordered pill inputs; add "All projects", "Add a city", and "Dashboard" pill actions
- Add an "All tools" heading above the module accordions with a capped, independently scrollable list
- Extract `CustomSelect` into `components/ui/custom-select.tsx` for reuse
- Fix the Plan stage icon color (was ignoring color due to a hardcoded fill and missing prop forwarding) and accordion sub-item indentation
- Restore the "Learning hub" link removed from the navbar under a new "Useful links" section
- Fix section alignment (search input, selects, "All tools" heading, "Useful links") to share a left edge, and refine drawer dividers/icon/link styling

## How to test

1. Run the app locally and open a city/organization with a GHGI project.
2. Open the JN sidebar drawer and confirm the organization switcher appears (and its dropdown works when the user belongs to multiple organizations).
3. Confirm the Project/City filters render as bordered pill selects and function correctly.
4. Confirm the "All tools" module list is capped in height and scrolls independently of the rest of the drawer.
5. Confirm "Useful links" > "Learning hub" opens in a new tab with the correct text styling.

## Ticket

CC-544


app/src/components/HomePage/JNDrawer.tsx
```diff
@@ -626,67 +693,108 @@ const JNDrawer = ({
...
+                  >
+                    {t("all-tools")}
+                  </Text>
+                  <Box maxH="500px" overflowY="auto">
+                    {stageOrder.map((stage) => {
+                      const modules = projectModules.filter((mod) => {
+                        // Filter out CCRA module
+                        if (mod.id === Modules.CCRA.id) {
+                          return false;
+                        }
```
lemilonkh: Do we need to manually filter these modules out here or can we just disable them for projects that don't need them?

lemilonkh (approved):
Great work, thanks 👍 ✨

app/src/lib/theme/recipes/app-theme.ts
```diff
@@ -707,6 +711,7 @@ export const appTheme = createSystem(defaultConfig, {
         rounded: { value: "8px" },
         "rounded-xl": { value: "16px" },
         "rounded-xxl": { value: "20px" },
+        "40": { value: "40px" },
```
lemilonkh: Might be better to have a name here?

lemilonkh (approved):

app/src/components/HomePage/JNDrawer.tsx
tatimoreira: CCRA was backfilled into ProjectModules for every existing project, so the API currently reports it as enabled everywhere. The manual filter here is mirroring the same temporary gate used on the home page (FeatureFlags.CCRA_MODULE) to hide the still-in-progress module from general users until it's ready.
 I've updated this file to check that flag too, since it was missing here and made the drawer inconsistent with the home page.

app/src/lib/theme/recipes/app-theme.ts
tatimoreira: agree!

PR #2925, opened 2026-07-29
[1/3] CC-608: Add CA Concept Note persistence

piotrnowakowski:
## Stack

1. **#2925 — CA persistence foundation** (this PR)
2. #2930 — CA run API
3. #2931 — CityCatalyst proxy and cross-service documentation

## Summary

- provision `concept_note_runs`, `concept_note_context_bundles`, and `concept_note_uploads` through the Climate Advisor Alembic chain using `CA_DATABASE_URL`
- add SQLAlchemy models and Concept Note persistence adapters
- group Concept Note persistence under `app/persistence/concept_notes`
- preserve existing city-context and Markdown-ingest behavior while moving those adapters

## Scope

Climate Advisor database and persistence only. The CA run API and CityCatalyst proxy are isolated in the two upper layers.

## Validation

- `33 passed` for `test_concept_note_city_context.py` and `test_concept_note_markdown_ingest.py`
- `alembic heads` — `20260729_120000 (head)`
- `git diff --check`

## History preservation

The original #2925 snapshot remains available at `codex/CC-608-pre-stack-backup-8ea15b2`. After rebasing the stack onto current `develop`, all three layer patch hashes remained unchanged and all 26 paths touched by the original PR retained identical Git blob hashes.


climate-advisor/service/app/models/db/concept_note.py
```diff
@@ -0,0 +1,237 @@
...
+    thread_id: Mapped[UUID | None] = mapped_column(
+        PGUUID(as_uuid=True),
+        nullable=True,
+    )
+    user_id: Mapped[str] = mapped_column(String(255), nullable=False)
+    name: Mapped[str] = mapped_column(String(120), nullable=False)
+    city_id: Mapped[str] = mapped_column(String(255), nullable=False)
+    project_id: Mapped[str | None] = mapped_column(String(255), nullable=True)
+    funder_id: Mapped[UUID | None] = mapped_column(
+        PGUUID(as_uuid=True),
```
DavidGBbr, line 45: `funder_id` / `selected_funding_record_id` have no FK, guessing that's because those entities live in the CityCatalyst core DB, so cross-DB FK isn't possible. Where does the existence check happen? In #2930/#2931?

climate-advisor/service/app/persistence/concept_notes/markdown.py
```diff
@@ -0,0 +1,609 @@
...
+    async def get_delivery_context(
+        self,
+        *,
+        upload_id: UUID,
+    ) -> ConceptNoteUploadSnapshot:
+        raise ConceptNoteStorageUnavailable()
+
+
+class SqlAlchemyConceptNoteMarkdownRepository(ConceptNoteMarkdownRepository):
+    """Persist CNB upload lifecycle state in the CA workflow database."""
```
DavidGBbr: If `get_session_factory()` fails here, it silently falls back to `UnavailableConceptNoteMarkdownRepository` (503) with just a log line. Is that the intended fail-safe, or should this have its own health check so a broken config in prod doesn't get masked?

climate-advisor/service/app/persistence/concept_notes/runs.py
```diff
@@ -0,0 +1,117 @@
...
+
+    async def thread_belongs_to_user(
+        self,
+        *,
+        thread_id: UUID,
+        user_id: str,
+    ) -> bool:
+        """Return whether a durable chat thread belongs to the user."""
+        result = await self.session.execute(
+            select(Thread.thread_id).where(
```
DavidGBbr, line 43: Since `Thread` lives in the same DB (`app.models.db.thread`), why check ownership manually instead of a real FK?

DavidGBbr (commented):
Really clean structure on this one, the split between `models/db`, `persistence/concept_notes`, and the migration is easy to follow, and the stack description made it simple to scope what to review. A few questions below, nothing blocking.

**General:** `ConceptNoteRunRepository.create_or_get` is the trickiest part of this diff (idempotency under concurrency) and has no test yet. Is that covered in #2930 once the route exists, or worth adding a concurrency test here already?

climate-advisor/service/app/models/db/concept_note.py
```diff
@@ -0,0 +1,237 @@
...
+        String(255),
+        nullable=False,
+    )
+    filename: Mapped[str] = mapped_column(String(255), nullable=False)
+    source_label: Mapped[str | None] = mapped_column(
+        String(255),
+        nullable=True,
+    )
+    markdown_s3_key: Mapped[str | None] = mapped_column(
+        String(1024),
```
DavidGBbr: Is there a file size limit enforced upstream? If not, a big PDF turns into one huge row here.

piotrnowakowski: yes there is 20mb limit upstream

climate-advisor/service/app/models/db/concept_note.py
piotrnowakowski: Correct about the cross-database FK limitation, with one clarification: these records live in the externally managed CNB reference database rather than the CityCatalyst core database. #2930 now validates both references before `create_or_get` using `CNB_DATABASE_URL`. Unknown funders, unknown records, and records belonging to another funder are rejected with 422; an unavailable reference database returns 503. Tests cover the valid and invalid cases.

climate-advisor/service/app/persistence/concept_notes/markdown.py
piotrnowakowski: Agreed. The unavailable repository remains the fail-safe for requests, but the problem is no longer hidden from deployment health checks. #2925 adds `GET /ready`, which executes a database query and returns 503 for missing configuration or an unreachable database. Kubernetes readiness probes now use `/ready`, while `/health` remains a pure liveness endpoint. Tests cover all three readiness outcomes.

climate-advisor/service/app/persistence/concept_notes/runs.py
piotrnowakowski: You're right that the current CA `Thread` model is queryable in this deployment. We intentionally kept `concept_note_runs.thread_id` without an FK because the CNB architecture treats it as a CityCatalyst-owned integration identifier; adding an FK would couple the new workflow schema to CA's transitional thread table. We retain the ownership check while that table exists and added coverage for missing and wrong-user threads. The model, migration, and architecture documentation now state this boundary explicitly.

climate-advisor/service/app/models/db/concept_note.py
piotrnowakowski: Yes, CityCatalyst enforces a 20 MiB source-PDF limit. Additionally, #2937 removes the large-row concern: Climate Advisor now stores only the immutable S3 key, digest, and metadata rather than inline Markdown content.

DavidGBbr (approved):

PR #2926, opened 2026-07-30
Fix "No data" state not showing for empty GHG inventories

tatimoreira:
## Summary

Two related bugs made the home page lie about GHG inventory status: an inventory with no real activity data showed "0kg" instead of "No data", and the action-card CTA below it picked collaborator-vs-admin messaging using the wrong signal, so some non-admin accounts saw an admin "get started" prompt instead of the "contact your administrator" empty state.

## Changes

- `InventoryService.getInventoryWithTotalEmissions`: stop coalescing a NULL emissions sum to 0. NULL now propagates whenever an inventory has no activity data (no `InventoryValue` rows, or all `co2eq` are null, e.g. notation-key-only prefill), letting the existing "No data" UI path fire correctly instead of showing a misleading 0kg.
- `models/Inventory.ts`: widen `totalEmissions` to `number | null` to match.
- `EmissionsWidget.tsx`: adjust one downstream consumer for the widened type.
- `HomePageJN/ActionCards.tsx` + `HomePage.tsx`: the action card decided collaborator-vs-admin messaging from per-city `CityUser` presence (`useGetUserAccessStatusQuery`) instead of the account's actual role. Now uses the same `isCollaboratorRole` (`userInfo.role === Roles.User`) the parent `HomePage.tsx` already computes, so the messaging is consistent with the rest of the page and correct for non-admin accounts with no city association yet.

## How to test

1. Find or seed a city with an inventory that has no real activity data (rows exist but `co2eq` is null, or zero `ActivityValue` rows).
2. Load that city's home page — the emissions stat should read "No data", not "0kg".
3. As a non-admin account with no `CityUser`/`ProjectAdmin`/`OrganizationAdmin` row, load the home page for a city with no data — the action card should show "No greenhouse gas inventory has been created yet / Contact your administrator...", not "Get started with GHG Inventory".

## Notes

Verified the emissions-sum fix directly against a seeded inventory (query returned `null` instead of `0` post-fix). `tsc --noEmit` shows only pre-existing, unrelated errors (confirmed identical on `develop`).

### DIFF
--- a/app/src/backend/InventoryService.ts
+++ b/app/src/backend/InventoryService.ts
@@ -60,11 +60,11 @@ export class InventoryService {
     }
 
     // TODO [ON-2429]: Save total emissions for inventory every time activity data is modified
-    // Returns 0 when inventory values exist but all co2eq are null (e.g. notation-key-only prefill),
-    // and NULL only when no inventory values exist at all.
+    // NULL when no activity data has been recorded yet (no inventory values, or all
+    // co2eq are null, e.g. notation-key-only prefill) so the UI can show "No data"
+    // instead of a misleading 0kg.
     const rawQuery = `
-    SELECT
-      CASE WHEN COUNT(*) > 0 THEN COALESCE(SUM(co2eq), 0) ELSE NULL END AS sum
+    SELECT SUM(co2eq) AS sum
     FROM "InventoryValue"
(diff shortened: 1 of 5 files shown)


lemilonkh (approved):
Great fix, thanks 👍

PR #2928, opened 2026-07-30
CC-545: Always show JN drawer, fix cities table links

tatimoreira:
## Summary
- Always render `JNDrawer` in the nav bar instead of gating it behind the `JN_ENABLED` flag with `ProjectDrawer` as the fallback.
- Fix the "View full list" link in the project cities table, which pointed to a nonexistent `/organization/{id}/account-settings` route; it now goes to `/settings?tab=team` with the project/city preselected via query params.
- Match the cities table row hover background to `border.neutral` (#D7D8FA) per design.

## Test plan
- [ ] Verify the JN drawer opens from the nav menu on desktop.
- [ ] From a project's cities table, click "View full list" and confirm it lands on the team tab of account settings with the right project/city preselected.
- [ ] Hover a cities table row and confirm the background matches `border.neutral`.

### DIFF
--- a/app/src/components/navigation-bar.tsx
+++ b/app/src/components/navigation-bar.tsx
@@ -42,9 +42,7 @@ import { Avatar } from "@/components/ui/avatar";
 
 import { Button } from "@/components/ui/button";
 import { Roles } from "@/util/types";
-import ProjectDrawer from "@/components/GHGIHomePage/ProjectDrawer";
 import { useTheme } from "next-themes";
-import { FeatureFlags, hasFeatureFlag } from "@/util/feature-flags";
 import { useOrganizationContext } from "@/hooks/organization-context-provider/use-organizational-context";
 import { Trans } from "react-i18next";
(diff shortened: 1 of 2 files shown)


app/src/components/navigation-bar.tsx
```diff
@@ -42,9 +42,7 @@ import { Avatar } from "@/components/ui/avatar";
 
 import { Button } from "@/components/ui/button";
 import { Roles } from "@/util/types";
-import ProjectDrawer from "@/components/GHGIHomePage/ProjectDrawer";
```
lemilonkh, line 45: Let's also delete the ProjectDrawer component as it's no longer used

lemilonkh (approved):
Awesome, thanks 👍 ✨

app/src/components/navigation-bar.tsx
lemilonkh: Will do that in a separate PR

PR #2930, opened 2026-07-30
[2/3] CC-608: Add CA Concept Note run API

piotrnowakowski:
## Stack

1. #2925 — CA persistence foundation
2. **#2930 — CA run API** (this PR)
3. #2931 — CityCatalyst proxy and cross-service documentation

## Summary

- add the Climate Advisor request and response contracts for Concept Note runs
- add authenticated start and read routes with idempotent run creation
- register the routes and connect the service to CityCatalyst authorization

## Scope

Climate Advisor API and service only. CityCatalyst proxy routes are intentionally delivered in layer 3.

## Validation

- imported the Climate Advisor app and verified `/v1/concept-notes/start` and `/v1/concept-notes/{run_id}` are registered
- `33 passed` for the existing Concept Note city-context and Markdown-ingest suites
- `git diff --check`

### DIFF
--- a/climate-advisor/service/app/models/concept_note_runs.py
+++ b/climate-advisor/service/app/models/concept_note_runs.py
@@ -1,0 +1,71 @@
+from __future__ import annotations
+
+from datetime import datetime
+from typing import Literal
+from uuid import UUID
+
+from pydantic import BaseModel, ConfigDict, Field, field_validator, model_validator
+
+
+class ConceptNoteStartRequest(BaseModel):
+    """Authenticated request to create one Concept Note Builder run."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    user_id: str = Field(min_length=1, max_length=255)
+    name: str = Field(min_length=1, max_length=120)
+    city_id: UUID
+    project_id: str | None = Field(default=None, min_length=1, max_length=255)
+    funder_id: UUID | None = None
+    selected_funding_record_id: UUID | None = None
+    thread_id: UUID | None = None
+    idempotency_key: UUID
+
+    @field_validator("name")
+    @classmethod
+    def normalize_name(cls, value: str) -> str:
+        """Trim the display name and reject whitespace-only values."""
+        normalized = value.strip()
+        if not normalized:
+            raise ValueError("name must not be blank")
+        return normalized
+
+    @field_validator("project_id")
+    @classmethod
+    def normalize_project_id(cls, value: str | None) -> str | None:
+        """Trim an optional external project identifier."""
+        if value is None:
+            return None
+        normalized = value.strip()
+        if not normalized:
+            raise ValueError("project_id must not be blank")
+        return normalized
+
+    @model_validator(mode="after")
+    def validate_scope_references(self) -> "ConceptNoteStartRequest":
+        """Require a funder whenever a funding opportunity is supplied."""
+        if self.selected_funding_record_id is not None and self.funder_id is None:
+            raise ValueError(
+                "funder_id is required when selected_funding_record_id is provided"
+            )
+        return self
+
+
+class ConceptNoteRunResponse(BaseModel):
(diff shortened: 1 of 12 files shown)


piotrnowakowski:
Restored: PR #2930 is open again as layer 2 of the original three-PR CC-608 stack. Its original head, base, title, and ready-for-review state are restored, and native stack #2936 now links #2925 -> #2930 -> #2931.

climate-advisor/service/app/models/concept_note_runs.py
```diff
@@ -0,0 +1,71 @@
...
+        if not normalized:
+            raise ValueError("project_id must not be blank")
+        return normalized
+
+    @model_validator(mode="after")
+    def validate_scope_references(self) -> "ConceptNoteStartRequest":
+        """Require a funder whenever a funding opportunity is supplied."""
+        if self.selected_funding_record_id is not None and self.funder_id is None:
+            raise ValueError(
+                "funder_id is required when selected_funding_record_id is provided"
```
DavidGBbr, line 49: `funder_id` / `selected_funding_record_id` only get a presence check here, not an existence check. Confirmed this isn't resolved in #2931 either, is it deferred to the datateam reference DB integration, or missing?

climate-advisor/service/app/routes/concept_note_runs.py
```diff
@@ -0,0 +1,63 @@
...
+    payload: ConceptNoteStartRequest,
+    authorization: str | None = Header(default=None),
+    session: AsyncSession = Depends(get_session),
+) -> JSONResponse:
+    """Create a concept-note run or replay an identical idempotent request."""
+    service = ConceptNoteRunService(session)
+    try:
+        response = await service.start_run(payload, authorization=authorization)
+        await session.commit()
+    except Exception:
```
DavidGBbr: Does `get_session` already handle commit/rollback in its own `finally`? Just checking this manual commit/rollback here isn't duplicating that.

DavidGBbr (commented):
No tests for `start_run`/`get_run` in this PR, no fingerprint mismatch (409) or ownership (403/404) coverage. Worth adding before merge?

climate-advisor/service/app/services/concept_note_runs.py
```diff
@@ -0,0 +1,229 @@
...
+            )
+
+        run = await self.repository.get_for_user(
+            run_id=run_id,
+            user_id=canonical_user_id,
+        )
+        if run is None:
+            raise HTTPException(status_code=404, detail="Concept Note run not found")
+
+        await self._validate_city_access(
```
DavidGBbr, line 108: `get_run` re-validates city access on every read via the CityCatalyst API. Any caching planned, or is that intentional to avoid stale access?

climate-advisor/service/app/models/concept_note_runs.py
piotrnowakowski: Fixed in 99d6fdd6a. start_run now validates funder_id and selected_funding_record_id against the managed CNB reference database, including record-to-funder ownership, with focused coverage for unknown and mismatched references.

climate-advisor/service/app/services/concept_note_runs.py
piotrnowakowski: Intentional. CityCatalyst owns city access, so Climate Advisor neither stores nor caches that authorization state. get_run revalidates against CityCatalyst on every read to avoid serving a stored run after access is revoked; c4cf8a9db adds regression coverage for the revoked-access 403.

climate-advisor/service/app/routes/concept_note_runs.py
piotrnowakowski: Fixed in c4cf8a9db. `get_session` handles rollback when exceptions propagate but does not commit, so the route now keeps the explicit commit and removes the duplicate `try`/`except` rollback.

DavidGBbr (approved):

PR #2933, opened 2026-07-31
fix(ui): CC-525 resolve mechanical ESLint issues in frontend

tatimoreira:
## Summary

| Stage | Total problems | `no-explicit-any` | Notes |
|---|---:|---:|---|
| Ticket baseline | 1726 | 714 | `npm run lint` was crashing outright (ESLint 10 incompatibility) |
| After mechanical pass | 858 | 714 | `no-unused-vars`, `no-require-imports`, etc. resolved |
| After first `no-explicit-any` spot-fix | 685 | 541 | 2 Modal components typed as a small follow-up |
| **This PR (current)** | **77** | **1** *(documented exception)* | Remaining 77 = 75 `react-hooks/*` (→ [CC-653](https://linear.app/openearth/issue/CC-653/fix-react-hooks-react-compiler-lint-issues-in-frontend)) + 1 `no-explicit-any` + 1 `jsx-a11y/alt-text` (both documented, permanent) |

- Fixes ESLint itself crashing on `npm run lint`: PR #2914 bumped `eslint` to 10.7.0, but `eslint-plugin-react` (bundled via `eslint-config-next`) and `eslint-plugin-i18next` still called rule-context APIs ESLint 10 removed. Originally patched both via `patch-package` + a `postinstall` script; **removed that entirely in review** (see below) — bumped `eslint-plugin-i18next` to a version that fixes it upstream, and added an `eslint.config.mjs` settings override that avoids the `eslint-plugin-react` crash without patching it.
- Resolves all mechanical ESLint findings across the `app/` package: `no-unused-vars`, `no-unsafe-function-type`, `no-require-imports` (minus documented CommonJS-required exceptions), and a handful of smaller rules (`no-img-element`, `alt-text`, `display-name`, `no-empty-object-type`, `ban-ts-comment`, `no-unused-expressions`, `no-page-custom-font`), plus `no-non-null-asserted-optional-chain`.
- Resolves essentially all of `@typescript-eslint/no-explicit-any` (was 714 at the start of this PR): every site was traced to its real type — model/API/library types where they exist, `Path<T>`/`PathValue<T, Path<T>>` casts for react-hook-form's templated field paths, local type extensions for undeclared Sequelize associations, and `object` (not `Record<string, unknown>`) as the constraint for generics that need to accept real model instances without index signatures. A few shared components (`PasswordInput`, `CustomSelectableButton`, `PercentageBreakdownInput`) were made generic over the caller's form type instead of hardcoding `FieldValues`, since callers pass their own concrete `Inputs` types.
  - **1 site intentionally left**, with a code comment: `activity-form-modal.tsx`'s `methodology: any` is genuinely `Methodology | DirectMeasure` at runtime — fixing it means reconciling both shapes across this file, `use-activity-form.ts`, and `emission-data-section.tsx`, not a one-line annotation. Investigated further in review; now tracked as [CC-659](https://linear.app/openearth/issue/CC-659/reconcile-methodology-vs-directmeasure-shape-mismatch-in-activity).
- Merged `develop` in multiple times over the life of this PR to resolve drift conflicts; picked up several independent `no-explicit-any`/dead-code fixes `develop` had already made in the process, plus fixed a handful of unused-import/var lint issues that came in with `develop`'s own recent commits.
- Fixed several real bugs surfaced by removing `any`/investigating "unused" warnings along the way:
  - A comma-instead-of-semicolon typo in `lib/mcp/tools/inventories.ts` that silently merged two statements.
  - A dropped debug-log field in `AIInterpretationService.ts`.
  - Two dead parameters (`ignoreExisting`, `session`) on `ModuleDashboardService.getHIAPDashboardData` that every caller passed but the function never used.
  - `emissions-factors-sync.ts` null/`datasetUrl` handling, `SectorTabs.tsx`'s dead/wrong `sectorName` field, an `onChange`-vs-`onCheckedChange` checkbox bug plus an unjoined `scopes` array in `add-file-data-dialog.tsx`, a response-type mismatch in `delete-city/inventory-modal.tsx`'s `requestVerification`, field-name bugs in `useActivityForm`'s emission-factor lookups, `co2eq_100yr` string-vs-bigint comparisons in 3 files, an array-push-instead-of-spread bug in `actions/route.ts`, an `InventoryUserFileAttributes.data` field that was typed as the backend's `Buffer` but the frontend actually stores a base64 string there, an implicitly-`any` `units` value that was silently mismatched against a stricter component prop, and `CitiesTable.tsx`'s `organizationId` prop that was threaded through but never actually used in its settings-page link.

**Left out of scope for this pass** — tracked as a new ticket, [CC-653](https://linear.app/openearth/issue/CC-653/fix-react-hooks-react-compiler-lint-issues-in-frontend):
- `react-hooks/*` React Compiler rules, 75 sites (`set-state-in-effect` 34, `incompatible-library` 12, `static-components` 11, `exhaustive-deps` 8, `preserve-manual-memoization` 4, `set-state-in-render` 3, `immutability` 2, `refs` 1). These are runtime/behavioral rules, not type issues — need per-site verification against the running app, not just `tsc`, which is why they were deferred out of this PR rather than rushed.

**Intentional permanent skips**, documented in code: `require()` in `.cjs` migration/seeder files (need CommonJS), a conditionally-guarded `require()` in `highlight.ts`, and the `react-pdf` `Image` `alt` prop.

## Review feedback addressed

Went through every open review thread individually — traced each one back through git history/call sites rather than applying blanket fixes, since several turned out not to be simple "remove the unused var" cases.

**Confirmed dead and removed** (prop/state declared in a type but never read by the component, or never passed by any caller — verified for each): `finalFocusRef` (`SourceDrawer.tsx`), `totalEmissionsData` (`ByScopeViewSourceDrawer.tsx` wrapper only — the underlying `SourceDrawer` prop of the same name is real and stays), `icon` (`EmissionsWidget.tsx`, plus its `EmissionsData` entries and now-unused `FiHeart`/`IconType` imports), `onImport` (`review-confirm-step.tsx` — the real "confirm & import" trigger lives in a separate `ImportButton` in the parent page since a prior button-placement refactor), `selectedMethodology` state and the `isSelected` prop it fed (`select-methodology.tsx` / `methodology-card.tsx` — the card manages its own selection highlight internally and never read the prop), `watch`/`register` (`dropdown-select-input.tsx`, which cascaded to removing `watch` from its caller `add-file-data-dialog.tsx` too, where it also became unused), `fallbackRoute` (`useResourceValidation.ts` — redirect logic is entirely derived from `userInfo` defaults), the `fetchUserProjects` stub on `UserService` (empty body, never called; distinct from the real, used `ProjectService.fetchUserProjects`), `id`/`isSelected` (`suggested-activities-card.tsx` and its caller), a dead `useGetOCCityDataQuery` call in `HomePageJN/Hero.tsx` (copy-paste residue from `GHGIHomePage/Hero.tsx` — this component already gets population via its own prop).
- Restored one Google Fonts `<link>` in `providers.tsx` that the original mechanical pass incorrectly removed as an unused-font-tag lint fix — it's a deliberate fallback for locally-hosted font load failures (ON-4761), not dead code. Added an inline eslint-disable with a comment instead.

**`t` prop type consistency**: switched several components' ad-hoc `t: (key: string) => string` (or `Function`) signatures to `TFunction` from `i18next`, matching the rest of the codebase — `JNDrawer.tsx`, `formatted-number-input.tsx`, `SegmentedProgressWithNames.tsx`.

**Type/structure cleanup per review**:
- Extracted a duplicated inline RTK Query response shape into `ProjectBoundary` in `util/types.ts`, used for both the query generic and `transformResponse` in `api.ts`.
- `HighImpactActionRanking` now declares its `inventory` association directly on the model class (matching `InventoryValue`'s existing pattern) — the runtime `belongsTo` association already existed, it just wasn't reflected in the TS type. This let `BulkHiapPrioritizationService.ts` drop its manual `RankingWithInventoryAndCity` intersection-type workaround entirely.
- Moved `MergedRankedAction` out of `HiapService.ts` into the module's shared `hiap/types.ts`, alongside its sibling `PrioritizerRankedAction`.
- Fixed `VersionEntry.tsx`'s imports, which were split across two blocks with an interface wedged in between.
- Corrected a misleading `@ts-expect-error` comment in `datasource.jest.ts`: `url` **is** present in `DataSource`'s typed attributes, it's just typed as `string | undefined` rather than nullable, unlike the underlying DB column.
- `CityLimitIcon` dropped its unused `color: string` prop — traced back to the component's first commit and confirmed the SVG fill was hardcoded from day one; `color` was never wired to anything.

**`patch-package` removed entirely**, per review question ("do we need this or can we use stock versions?"):
- `eslint-plugin-i18next`'s patch is fixed upstream in 6.1.5 (bumped to it).
- `eslint-plugin-react`'s crash (still present in its latest release, `7.37.5`, nested inside `eslint-config-next`) is avoided by overriding `settings.react.version` in `eslint.config.mjs` instead of leaving it on `eslint-config-next`'s default `"detect"`, which is what triggers the crashing auto-detection path under ESLint 9's flat config. Verified by reproducing the exact crash with the override removed, then confirming a clean run on a fully fresh `node_modules` install.
- Bumping `eslint-plugin-i18next` did surface one new, genuine `i18next/no-literal-string` finding (an untranslated `"Fetching Cities..."` string the older/patched plugin version wasn't catching correctly) — fixed by adding a proper translation key across all 5 locales.

**New follow-up tickets filed from this review** (each is either a design decision or a larger fix than fit in this PR):
- [CC-658](https://linear.app/openearth/issue/CC-658/restore-or-remove-the-dead-by-activityby-scope-table-toggle-in) — a table-view toggle in `ReportResults.tsx` has been fully disabled since ON-3126 (Dec 2024); needs a decision to restore or fully remove the dead toggle.
- [CC-659](https://linear.app/openearth/issue/CC-659/reconcile-methodology-vs-directmeasure-shape-mismatch-in-activity) — `methodology: any` in `activity-form-modal.tsx` isn't a simple typing fix: `Methodology`/`DirectMeasure` use different key names for the same data, and two divergent code paths in `activity-tab.tsx` populate the same state with inconsistent shapes.
- [CC-660](https://linear.app/openearth/issue/CC-660/finish-or-remove-unwired-selectedproject-state-in-ghgi-onboarding) — `selectedProject` state in the GHGI onboarding setup page is populated but never read anywhere, including by the `addInventory` call; looks like unfinished Enterprise Mode project-selection wiring.

## CI build fix

CI's build job (`npm run build`'s own `tsc` type-check, distinct from the ESLint work above) had been failing on every single commit on this branch going back weeks — confirmed identical on `develop` (passes) vs. this branch (44 `tsc --noEmit` errors, present all the way back to before this review round started). Traced it to fallout from this PR's own earlier `no-explicit-any` typing pass: several shared form components had their loosely-typed `register`/`t` props tightened without being made generic, breaking assignability for callers using different concrete `Inputs` types. Fixed all 44, `tsc --noEmit` now passes clean:

- `email-input.tsx` / `form-select-input.tsx`: made generic over `TFieldValues`, matching `PasswordInput`'s existing pattern — their `register` props required `name: string` (any field name), but callers pass react-hook-form's real `UseFormRegister<Inputs>`, which only accepts `Inputs`' literal keys. Resolved 6 files' worth of errors at once (4 `EmailInput` callers + `update-user-modal.tsx` via `FormSelectInput`). `form-select-input.tsx` also had a `disabled` prop on `NativeSelectField`, which Chakra deliberately omits from that component's type — moved to `NativeSelectRoot`, which has it.
- `branding/route.ts`: a real bug — `session` was referenced but never destructured from `apiHandler`'s params; would have thrown a `ReferenceError` at runtime if this endpoint were ever hit.
- `file-data-card.tsx`: removed a `size` prop on `ExcelFileIcon` that's been silently ignored since the icon never accepted it (confirmed against its only other 2 call sites, neither of which pass it).
- Both onboarding setup pages: removed a dead `register` prop passed to `SetPopulationDataStep` (never read anywhere in that component), and added a `setValue` cast consistent with the same file's existing `UseFormSetValue<GHGIFormInputs>` pattern used for sibling steps.
- **Deleted `ActivityDataTab.tsx` and `DirectMeasureForm.tsx` entirely**: confirmed zero importers anywhere in the codebase (including their other named exports), already orphaned before a pre-PR file move (`45dac205c`), and using a broken/outdated Chakra `Select.Root` API pattern (bare `<option>` children + `register` spread) that doesn't match Chakra v3's Ark-based `Select` at all — not worth fixing UI code nobody renders.
- `activity-modal-body.tsx`: narrowed a validation helper's `fieldName` parameter from `string` to the 3 literal values it's ever called with, so its templated `` `activity.${fieldName}` `` paths type-check against react-hook-form's strict `FieldPath` union.

Linear: CC-525

## Test plan
- [x] `npm run lint` runs cleanly (no crash) and reports the reduced problem count (77, down from 1726)
- [x] `npx tsc --noEmit` passes with **zero errors** (was 44, all pre-existing on this branch — see CI build fix section)
- [x] `npm run build` compiles successfully and generates all 175 pages
- [x] Verified `patch-package` removal: reproduced the `eslint-plugin-react` crash with the config override removed, then confirmed a clean `eslint .` run against a fully fresh `node_modules` install with the override restored
- [x] CI green — both checks passing on `b110408` (first green run on this branch)


app/patches/eslint-config-next++eslint-plugin-react+7.37.5.patch
lemilonkh, line 1: Can we avoid needing to patch the eslint-config-next package?

app/patches/eslint-plugin-i18next+6.1.3.patch
lemilonkh, line 1: Can we avoid needing to patch eslint-plugin-i18next?

app/src/app/[lng]/admin/bulk-inventory-actions/CityAutoCompleteInput.tsx
```diff
@@ -1,17 +1,7 @@
...
-  Tag,
-  TagLabel,
-  ListItem,
-  Spinner,
-  List,
-  Text,
-} from "@chakra-ui/react";
+import { Box, Input, Tag, Spinner, List, Text } from "@chakra-ui/react";
 import { useGetOCCityQuery } from "@/services/api";
-import { useOutsideClick } from "@/lib/use-outside-click";
```
Comment (lemilonkh) [inline app/src/app/[lng]/admin/bulk-inventory-actions/CityAutoCompleteInput.tsx:14]: Is this file still used somewhere else? Otherwise it could be removed as well.

app/src/app/[lng]/cities/[cityId]/GHGI/[inventory]/manage-sectors/SectorTabs.tsx
```diff
@@ -135,7 +136,7 @@ const SectorTabs: FC<SectorTabsProps> = ({ t, inventoryId }) => {
   const [showDialog, setShowDialog] = useState(false);
   const pathname = usePathname();
   const [prevPathname, setPrevPathname] = useState(pathname);
-  const [nextRoute, setNextRoute] = useState<string | null>(null);
+  const [, setNextRoute] = useState<string | null>(null);
```
Comment (lemilonkh) [inline app/src/app/[lng]/cities/[cityId]/GHGI/[inventory]/manage-sectors/SectorTabs.tsx]: I think the state can be removed if it's value is not being used anywhere? We can probably get rid of the setter call as well then.

app/src/app/[lng]/cities/onboarding/layout.tsx
```diff
@@ -8,7 +8,7 @@ export default function OnboardingLayout(props: {
   children: React.ReactNode;
   params: Promise<{ lng: string }>;
 }) {
-  const { lng } = use(props.params);
+  use(props.params);
```
Comment (lemilonkh) [inline app/src/app/[lng]/cities/onboarding/layout.tsx]: This call and the use import can likely be removed as well.

app/src/app/[lng]/methodologies/page.tsx
```diff
@@ -12,7 +12,7 @@ import Footer from "@/components/Sections/Footer";
 
 export default function MethodologiesPage() {
   const lng = i18next.language as LANGUAGES;
-  const { t } = useTranslation("methodologies");
+  useTranslation("methodologies");
```
Comment (lemilonkh) [inline app/src/app/[lng]/methodologies/page.tsx]: Call and import can be removed if we don't use `t`

app/src/app/[lng]/settings/project/projectDetails.tsx
```diff
@@ -89,7 +89,7 @@ const ProjectDetails: React.FC<ProjectDetailsProps> = ({
     setSelectedInventory(null);
   }, [selectedCityData]);
 
-  const [userData, setUserData] = useState<UserAttributes>({
+  const [userData] = useState<UserAttributes>({
```
Comment (lemilonkh) [inline app/src/app/[lng]/settings/project/projectDetails.tsx:96]: Can maybe be removed if we never set actual values for this?

app/src/app/api/v1/city/[city]/modules/hiap/dashboard/route.ts
```diff
@@ -89,7 +89,6 @@ export const GET = apiHandler(async (req: Request, context) => {
   const {
     inventoryId,
     lng = "en",
-    ignoreExisting,
```
Comment (lemilonkh) [inline app/src/app/api/v1/city/[city]/modules/hiap/dashboard/route.ts:92]: Why was this query parameter removed? It is being passed into the ModuleDashboardService further below...

app/src/app/api/v1/city/[city]/modules/hiap/dashboard/route.ts
```diff
@@ -121,8 +120,6 @@ export const GET = apiHandler(async (req: Request, context) => {
     cityId,
     inventory,
     lng,
-    session ?? undefined,
-    ignoreExisting,
```
Comment (lemilonkh) [inline app/src/app/api/v1/city/[city]/modules/hiap/dashboard/route.ts:125]: I think we might need these arguments?

app/src/app/api/v1/mcp/route.ts
```diff
@@ -325,23 +325,10 @@ async function handleCallTool(request: any, session: AppSession) {
...
-        error.message.includes("Unauthorized") ||
-        error.message.includes("Authentication")
-      ) {
-        errorCode = ErrorCode.InvalidRequest;
-      } else if (
-        error.message.includes("not found") ||
-        error.message.includes("Not found")
-      ) {
-        errorCode = ErrorCode.InvalidParams;
-      }
```
lemilonkh, line 344: This seems necessary for the MCP API to provide correct error feedback?

app/src/app/providers.tsx
```diff
@@ -53,10 +53,6 @@ export function Providers({ children }: { children: React.ReactNode }) {
         href="https://fonts.gstatic.com"
         crossOrigin="anonymous"
       />
-      <link
-        href="https://fonts.googleapis.com/css2?family=Open+Sans:wght@400;600&family=Poppins:wght@400;500;600&display=swap"
-        rel="stylesheet"
-      />
```
lemilonkh, line 59: Are we sure this isn't being used? We had trouble with the locally hosted fonts previously, so this was added to make sure they load correctly I think.

app/src/backend/ModuleDashboardService.ts
```diff
@@ -82,8 +81,6 @@ export class ModuleDashboardService {
     cityId: string,
     inventory: Inventory,
     lng: string = "en",
-    session?: AppSession,
-    ignoreExisting: boolean = false,
```
lemilonkh, line 86: I see, these weren't being used in the function?
Then feel free my earlier comment about the usage of this function 😅

app/src/backend/UserService.ts
```diff
@@ -1101,7 +1101,7 @@ export default class UserService {
     return responseObject;
   }
 
-  public async fetchUserProjects(userId: string) {}
+  public async fetchUserProjects() {}
```
lemilonkh: This function isn't being used anywhere and can be removed I'd say.

app/src/components/Cards/suggested-activities-card.tsx
```diff
@@ -15,8 +15,6 @@ interface SuggestedActivityCardProps {
 
 const SuggestedActivityCard: FC<SuggestedActivityCardProps> = ({
   t,
-  isSelected,
-  id,
```
lemilonkh, line 19: Can' we remove these from the type definition of the component and the component usages as well?

app/src/components/GHGI/data-step/SourceDrawer.tsx
```diff
@@ -37,7 +37,6 @@ export function SourceDrawer({
   isOpen,
   onClose,
   onConnectClick,
-  finalFocusRef,
```
lemilonkh, line 40: Can we remove this from the type definition and component usages as well?

app/src/components/GHGI/inventory-pages/add-data-steps-page.tsx
```diff
@@ -298,7 +296,7 @@ export default function AddDataSteps() {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [inventoryProgress]);
 
-  const { value: activeStep, goToNextStep: goToNext } = useSteps({
+  const { value: activeStep } = useSteps({
```
lemilonkh, line 300: If we don't use goToNextStep, are these steps even being used? Or can we get rid of the hook?

app/src/components/GHGI/inventory-result/ByScopeViewSourceDrawer.tsx
```diff
@@ -20,7 +20,6 @@ const ByScopeViewSourceDrawer: React.FC<ByScopeViewProps> = ({
   isOpen,
   onClose,
   t,
-  totalEmissionsData,
```
lemilonkh, line 23: Should be removed from type definition and usages as well.

app/src/components/GHGI/inventory-result/EmissionsWidget.tsx
```diff
@@ -23,7 +23,6 @@ import { FiHeart } from "react-icons/fi";
 import { api } from "@/services/api";
 
 const EmissionsWidgetCard = ({
-  icon,
```
lemilonkh, line 26: Should be removed from type definition and usages

app/src/components/GHGI/ReportResults.tsx
```diff
@@ -55,9 +55,7 @@ function SectorTabs({
 }) {
   const { t: tData } = useTranslation(lng, "data");
   const [selectedTab, setSelectedTab] = useState(SECTORS[0].name);
-  const [selectedTableView, setSelectedTableView] = useState<TableView>(
-    TableView.BY_ACTIVITY,
-  );
+  const [selectedTableView] = useState<TableView>(TableView.BY_ACTIVITY);
```
lemilonkh, line 58: If we don't use the setter, can this be removed?
I think this feature still exists, so would be good to look into why it wasn't being used.

app/src/components/GHGI/ReportResults.tsx
```diff
@@ -96,10 +94,6 @@ function SectorTabs({
     refetch().finally(() => setIsLoadingNewData(false));
   }, [selectedTab, refetch]);
 
-  const handleViewChange = (event: ChangeEvent<HTMLSelectElement>) => {
-    setSelectedTableView(event.target.value as TableView);
-  };
```
lemilonkh, line 101: If this isn't being called, let's look into it because I think we still need this feature.

app/src/components/GHGIHomePage/Hero.tsx
```diff
@@ -51,7 +50,6 @@ interface HeroProps {
 }
 
 export function Hero({
-  currentInventoryId,
```
lemilonkh, line 54: Remove from type definition and usages as well

app/src/components/GHGIHomePage/InventoryPreferencesCard.tsx
```diff
@@ -13,7 +13,6 @@ import NextLink from "next/link";
 
 export function InventoryPreferencesCard({
   t,
-  isPublic,
```
lemilonkh, line 16: Remove from type definition and usages as well

app/src/components/GHGIHomePage/ProjectDrawer.tsx
```diff
@@ -58,7 +51,7 @@ const ProjectList = ({
   data,
   selectProject,
 }: {
-  t: Function;
+  t: (key: string) => string;
```
lemilonkh: Could also use TFunction for this

app/src/components/GHGIHomePage/ProjectDrawer.tsx
```diff
@@ -120,7 +113,7 @@ const SingleProjectView = ({
 }: {
   project: ProjectWithCities;
   backToProjects: () => void;
-  t: Function;
+  t: (key: string) => string;
```
lemilonkh: Could also use TFunction for this

app/src/components/HomePage/JNDrawer.tsx
```diff
@@ -57,7 +54,7 @@ const ProjectFilterSection = ({
   currentCityId,
   organizationId,
 }: {
-  t: Function;
+  t: (key: string) => string;
```
lemilonkh: Could be TFunction

app/src/components/HomePageJN/Hero.tsx
```diff
@@ -41,7 +40,7 @@ export function Hero({
   hideMap = false,
   numberFormat,
 }: HeroProps) {
-  const { data: cityData } = useGetOCCityDataQuery(city?.locode!, {
+  useGetOCCityDataQuery(city?.locode!, {
```
lemilonkh: Query and import can be removed if we don't use the variable.

app/src/components/Modals/activity-modal/activity-form-modal.tsx
```diff
@@ -32,19 +33,33 @@ interface AddActivityModalProps {
...
   t: TFunction;
   defaultCityId?: string;
-  setHasActivityData: Function;
+  setHasActivityData: Dispatch<SetStateAction<boolean>>;
   hasActivityData: boolean;
   inventoryId: string;
+  // Can be Methodology | DirectMeasure at runtime (see activity-tab.tsx),
+  // but internal usage here assumes Methodology-only fields (activitySelectionField,
+  // fields, etc.) without narrowing -- needs a proper fix reconciling the two shapes
+  // across this file, use-activity-form.ts, and emission-data-section.tsx.
```
lemilonkh, line 42: Can this be done as part of the PR? Otherwise feel free to create a follow-up ticket. Seems like an actual source of bugs otherwise.

app/src/components/steps/GHGI/import/review-confirm-step.tsx
```diff
@@ -35,7 +34,6 @@ export default function ReviewConfirmStep({
   cityId,
   cityName,
   importedFileId,
-  onImport,
```
lemilonkh, line 43: Remove from type definition and usages as well

app/src/components/Tabs/Activity/scope-unavailable.tsx
```diff
@@ -54,7 +53,7 @@ const ScopeUnavailable: FC<ScopeUnavailableProps> = ({
   const [markAsUnavailable, { isLoading }] =
     api.useUpdateOrCreateInventoryValueMutation();
 
-  const { field } = useController({
+  useController({
```
lemilonkh, line 56: I think this hook might be removed if it's not used? Or does it register something?

app/src/components/Tabs/Activity/select-methodology.tsx
```diff
@@ -17,7 +17,7 @@ const SelectMethodology = ({
   handleMethodologySelected: (methodology: Methodology | DirectMeasure) => void;
   directMeasure?: DirectMeasure;
 }) => {
-  const [selectedMethodology, setSelectedMethodology] = useState("");
+  const [selectedMethodology] = useState("");
```
lemilonkh: If we don't use the setter, the state can likely be removed as well.

app/src/components/ui/navigation-links.tsx
```diff
@@ -12,7 +10,7 @@ interface NavigationItem {
 
 interface NavigationLinksProps {
   items: NavigationItem[];
-  t: Function;
+  t: (key: string) => string;
```
lemilonkh, line 13: Could be TFunction

app/src/components/dropdown-select-input.tsx
```diff
@@ -24,8 +24,6 @@ const DropdownSelectInput: React.FC<DropdownSelectProps> = ({
   subsectors,
   setValue,
   t,
-  watch,
-  register,
```
lemilonkh, line 28: Remove from type definition and usages as well

app/src/components/formatted-number-input.tsx
```diff
@@ -25,7 +27,7 @@ interface FormattedNumberInputProps<T extends FieldValues>
   miniAddon?: boolean;
   testId?: string;
   localization?: string;
-  t: Function;
+  t: (key: string, options?: Record<string, unknown>) => string;
```
lemilonkh: Could be TFunction

app/src/components/icons.tsx
```diff
@@ -2035,7 +2035,7 @@ export const CitiesBuildingIcon = () => {
   );
 };
 
-export const CityLimitIcon = ({ color }: { color: string }) => {
+export const CityLimitIcon = ({}: { color: string }) => {
```
lemilonkh: Might be good to use the color? Or remove the argument and update usages...

app/src/components/password-input.tsx
```diff
@@ -22,7 +22,7 @@ export default function PasswordInput({
 }: {
   children?: React.ReactNode;
   error: FieldError | undefined;
-  register: Function;
+  register: UseFormRegister<any>;
```
lemilonkh: Might need to use `<T extends FieldValues>` for the component so we can use `<T>` instead of `<any>` here.

app/src/components/RiskCard.tsx
```diff
@@ -18,7 +18,6 @@ export const RiskCard = ({
   risk,
   resilienceScore = null,
   t,
-  onSeeMoreClick,
```
lemilonkh, line 21: Remove from type definition and usages as well

app/src/components/SegmentedProgressWithNames.tsx
```diff
@@ -15,7 +15,7 @@ export function SegmentedProgressWithNames({
   colors?: string[];
   max?: number;
   height?: number;
-  t: Function;
+  t: (key: string) => string;
```
lemilonkh: Could be TFunction

app/src/hooks/useResourceValidation.ts
```diff
@@ -24,7 +24,6 @@ export function useResourceValidation({
   resourceQuery,
   lng,
   resourceType,
-  fallbackRoute
```
lemilonkh, line 27: Remove from type definition and usages as well

app/tests/api/datasource.jest.ts
```diff
@@ -215,7 +215,7 @@ describe("DataSource API", () => {
 
   it("should delete an inventory value", async () => {
     const datasource = await db.models.DataSource.findOne({
-      // @ts-expect-error URL can be null
+      // @ts-expect-error -- DataSource model attributes don't include `url` in its typed WhereOptions
```
lemilonkh: It does include it, the type definition just doens't allow for null here (even though the column does I think).

app/package.json
```diff
@@ -4,6 +4,7 @@
   "private": true,
   "type": "module",
   "scripts": {
+    "postinstall": "patch-package",
```
lemilonkh: Let's figure out if this is necessary for us or if we can use the stock versions of the packages. Always comes with a bit of maintenance burden when we need to update the packages.

lemilonkh (approved):
Amazing work, tysm! 🎉
Let's figure out if `patch-package` is absolutely necessary or if we can live with the stock versions of the `eslint` related packages.
If we need to downgrade eslint to 9 again that's fine, we can wait until our dependencies update.

app/src/app/[lng]/cities/[cityId]/GHGI/onboarding/setup/page.tsx
```diff
@@ -308,7 +312,9 @@ export default function OnboardingSetup(props: {
     }
   }, [activeStep, isUploadMode]);
 
-  const [selectedProject, setSelectedProject] = useState<string[]>([]);
+  // Tracked but never read/sent to addInventory yet — looks like unfinished
+  // wiring rather than dead code, left as-is pending a follow-up ticket.
+  const [_selectedProject, setSelectedProject] = useState<string[]>([]);
```
Comment (lemilonkh) [inline app/src/app/[lng]/cities/[cityId]/GHGI/onboarding/setup/page.tsx:311]: Can you create a follow up ticket for this? Thanks.

app/src/backend/hiap/BulkHiapPrioritizationService.ts
```diff
@@ -5,11 +5,20 @@ import {
...
+import { hiapServiceWrapper } from "./HiapService";
 import { hiapApiWrapper } from "./HiapApiService";
 import { Op, QueryTypes } from "sequelize";
+import { HighImpactActionRanking } from "@/models/HighImpactActionRanking";
+import { Inventory } from "@/models/Inventory";
+import { City } from "@/models/City";
 
-interface CityInventoryData {
+// The `inventory`/`city` associations are eager-loaded via `include` below
+// but aren't declared on the HighImpactActionRanking model class itself.
```
lemilonkh: We can declare these on the models as we have done with joined models in others like InventoryValue. Fine like it is for now though.

app/src/backend/hiap/HiapService.ts
```diff
@@ -603,21 +590,47 @@ export function normalizeRankedActionForLang(action: any, lang: LANGUAGES) {
...
+  GHGReductionPotential: { [k: string]: string };
+  adaptationEffectiveness: string | null;
+  costInvestmentNeeded: string | null;
+  timelineForImplementation: string | null;
+  dependencies: string[];
+  keyPerformanceIndicators: string[];
+  powersAndMandates: string[] | null;
+  adaptationEffectivenessPerHazard: { [k: string]: string };
+  biome: string | null;
+}
```
lemilonkh: Can we move this to the top of the file or our shared types.ts file?

app/src/components/GHGI/inventory-versions/VersionEntry.tsx
```diff
@@ -4,6 +4,27 @@ import { UseErrorToast } from "@/hooks/Toasts";
...
+  totalEmissionsChangeSign?: number;
+  source?: string;
+  hasDataSource?: boolean;
+  previousSource?: string;
+  hasPreviousDataSource?: boolean;
+  // HIAP-specific fields
+  name?: string;
+  isSelected?: boolean;
+  rank?: number;
+}
```
lemilonkh, line 47: Can we move this below the imports?

app/src/components/HomePageJN/Hero.tsx
```diff
@@ -41,7 +40,7 @@ export function Hero({
   hideMap = false,
   numberFormat,
 }: HeroProps) {
-  const { data: cityData } = useGetOCCityDataQuery(city?.locode!, {
+  useGetOCCityDataQuery(city?.locode ?? "", {
     skip: !city?.locode,
   });
```
lemilonkh: No need to call the API if the value isn't being used.

app/src/services/api.ts
```diff
@@ -1407,15 +1410,38 @@ export const api = createApi({
...
+          result: Array<{
+            city: {
+              id: string;
+              name: string;
+              locode: string;
+              latestInventoryId: string;
+            };
+            boundingBox: BoundingBox;
+            data: GeoJSON;
+          }>;
```
lemilonkh: Can perhaps be extracted into a type definition in types.ts

lemilonkh (approved):
Wow, awesome work! 🎉 ✨ 🥳 

Please make sure to also consider the feedback from the previous review.

app/src/app/[lng]/admin/bulk-inventory-actions/CityAutoCompleteInput.tsx
Comment (tatimoreira) [reply app/src/app/[lng]/admin/bulk-inventory-actions/CityAutoCompleteInput.tsx:14]: Yes, it's still used in BulkInventoryCreationTabContent.tsx

app/src/app/api/v1/city/[city]/modules/hiap/dashboard/route.ts
Comment (tatimoreira) [reply app/src/app/api/v1/city/[city]/modules/hiap/dashboard/route.ts:92]: ignoreExisting is still validated by querySchema.parse() above (still a valid, documented query param) — it's just no longer destructured into a local variable, because it's no longer passed to ModuleDashboardService.getHIAPDashboardData(). That call only takes (cityId, inventory, lng) now.

app/src/app/api/v1/city/[city]/modules/hiap/dashboard/route.ts
Comment (tatimoreira) [reply app/src/app/api/v1/city/[city]/modules/hiap/dashboard/route.ts:125]: I checked — getHIAPDashboardData never used either argument internally, so this is safe. Its full body only does: look up the city, check ModuleService.hasModuleAccess, then call getHIAPStatusData twice (which only takes inventoryId, type, lng). Neither session nor ignoreExisting was read anywhere in there — no recalculation branch, no permission check inside the service (auth is already enforced earlier in this route via PermissionService.canAccessInventory).

They were purely dead arguments being threaded through from every caller for no effect. The real ignoreExisting-gated recalculation logic lives in a separate service path (HiapService, used by the inventory-level /hiap reprioritize endpoint), not this dashboard read.

app/src/app/api/v1/mcp/route.ts
tatimoreira: This was dead code before this change too — no error feedback is lost. The client still gets the specific message (error.message, e.g. "Unauthorized...", "not found..."), just never got a structured code in the first place.

Separately: this does point at a pre-existing gap — a spec-correct JSON-RPC 2.0 error response should include a code field (and probably use error: {code, message} rather than result.isError) so clients can branch on error type instead of string-matching messages. Worth its own ticket if we want that; out of scope for this cleanup pass.

app/src/app/providers.tsx
tatimoreira: You're right, I revert this.

app/src/components/GHGI/inventory-pages/add-data-steps-page.tsx
tatimoreira: Good catch on `goToNextStep` — checked the history and it was dead from the moment it was introduced. It was destructured in the June refactor (`41fb15b88`, "move inventory page logic to shared components") but never wired to any next/prev-step control in this file, before or after. Removing it here is correct.

The rest of the hook does still matter though: `activeStep` (the `value` returned by `useSteps`) drives `currentStep = steps[activeStep]`, which is used throughout — sector heading, progress bars, subsector list, data sources, everything. So we can't drop `useSteps` outright without losing that.

That said, you're onto something — I traced every usage and nothing in this file ever calls `goToNextStep`, `goToPrevStep`, or `setStep`. `activeStep` is set once from `defaultStep: Number(step) - 1` and never changes for the life of the component; the only way it changes is a full route navigation to a new `/data/[step]` URL, which remounts the component and re-derives `defaultStep` from scratch anyway. So `useSteps` is currently just an indirect way of computing `Number(step) - 1` — we could replace the whole thing with:

```ts
const activeStep = Number(step) - 1;
```

...and drop the `useSteps` import and the `history.replaceState` sync effect along with it, since that effect only exists to keep the URL in sync with a value that never diverges from the URL in the first place.

I'd treat that as a follow-up rather than bundling it into this ESLint-cleanup PR, since it's a step beyond removing an unused variable — happy to open a quick follow-up PR for it if that sounds right.

app/src/components/GHGI/ReportResults.tsx
tatimoreira: Traced this — it's not something this PR broke, it's the tail end of a year-old feature flag.

The by-activity/by-scope table toggle was deliberately disabled in `d41b36ae1` (Dec 2024, **ON-3126** "hide breakdown by activity"). That commit:
- Forced `shouldShowTableByActivity` to always `false` via a hardcoded `false &&` short-circuit
- Commented out the entire `<Selector>` control (`onChange={handleViewChange}`) that was the *only* caller of the setter
- Left a comment, `// ON-3126 restore view by activity`, signaling it was meant to be temporary

That disabled state was later copy-pasted into this file when the dashboard's full GHGI report view was built (`db65c8ecd`), carrying the dead toggle forward. So the setter has had zero callers for well over a year — this cleanup just correctly flagged code that's been inert since ON-3126 landed.

Worth noting: it's not only the setter that's dead — `shouldShowTableByScope` doesn't check `selectedTableView` anymore either (that condition is commented out too), so the state's *value* has no effect on anything now. Removing the setter here is safe.

Since the feature has been flagged "temporary" for 1.5+ years, I think this is worth its own ticket to either restore the by-activity view or rip out the dead toggle entirely (enum, state, commented-out `Selector`, both `should*` conditionals) — will file one rather than scope-creeping this ESLint PR.

tatimoreira: Follow-up ticket created: [CC-658](https://linear.app/openearth/issue/CC-658/restore-or-remove-the-dead-by-activityby-scope-table-toggle-in) — tracks deciding whether to restore the by-activity view or fully remove the dead toggle code.

app/src/components/GHGI/ReportResults.tsx
tatimoreira: Same root cause as the `setSelectedTableView` thread below — `handleViewChange` was only ever wired to the `<Selector onChange={handleViewChange}>` control, which was commented out as part of **ON-3126** ("hide breakdown by activity", Dec 2024, `d41b36ae1`). With that JSX commented out, this function has had zero callers for 1.5+ years, same as the setter. See the fuller writeup [here](https://github.com/Open-Earth-Foundation/CityCatalyst/pull/2933#discussion_r3732340122).

Filed [CC-658](https://linear.app/openearth/issue/CC-658/restore-or-remove-the-dead-by-activityby-scope-table-toggle-in) to decide whether to restore the by-activity view (which would bring `handleViewChange` back) or fully remove the dead toggle. Safe to leave removed here in the meantime.

app/src/components/Modals/activity-modal/activity-form-modal.tsx
tatimoreira: Traced this — it's more than a type annotation, so I'd rather not rush it into this PR. `Methodology` and `DirectMeasure` (`util/form-schema/index.ts`) aren't structurally compatible: `Methodology` stores extra fields under `fields?: ExtraField[] | Activity[]`, while `DirectMeasure` stores the equivalent data under a differently-named key, `"extra-fields"?: ExtraField[]`.

In `activity-tab.tsx`, the source state is honestly typed `useState<Methodology | DirectMeasure>()`, but there are two separate paths that populate it: a `useEffect` that normalizes by remapping `.activities`/`["extra-fields"]` into a synthetic `fields` prop when deriving from a saved activity value, and a separate `onMethodologyChanged` callback (used when the user actively picks a methodology) that sets the raw union **without** that normalization. Downstream, `use-activity-form.ts` already claims `methodology: Methodology` but internally does `id.includes("direct-measure")` runtime checks and `as` casts to cope with DirectMeasure-shaped data anyway — the declared type doesn't match what it can actually receive.

So a real fix means reconciling two shapes with different key names across two divergent state-update paths, then propagating a corrected type through the hook and this modal — a design decision, not a mechanical typing pass, and there's real regression risk doing it quickly here. Filed [CC-659](https://linear.app/openearth/issue/CC-659/reconcile-methodology-vs-directmeasure-shape-mismatch-in-activity) to track it properly.

app/package.json
tatimoreira: Dug into both patches — turns out we don't need `patch-package` at all anymore.

**`eslint-plugin-i18next` patch** — fixed upstream. It patched 3 calls to `context.getSourceCode()`, which was removed/deprecated under ESLint 9's flat config. `eslint-plugin-i18next@6.1.5` already ships a compat fix (`context.sourceCode || context.getSourceCode()`), and it's within our existing `^6.1.3` range — just needed the lockfile to actually pick it up. Bumped to `^6.1.5`.

**`eslint-config-next`/`eslint-plugin-react` patch** — this one's on a nested dependency, and `7.37.5` is still the latest published `eslint-plugin-react`, so a version bump doesn't fix it. But I traced the actual trigger: `eslint-config-next`'s bundled config sets `settings.react.version: "detect"`, and it's specifically that `"detect"` value that runs the crashing auto-detection path (`contextOrFilename.getFilename()`, which doesn't exist on ESLint 9's flat-config context). Adding our own override — `settings: { react: { version: "19.2.8" } }` in `eslint.config.mjs`, placed after we extend Next's config — skips that code path entirely, no patch required.

Verified this properly: reverted the config override and reproduced the exact crash (`contextOrFilename.getFilename is not a function`) on a clean `eslint-plugin-react@7.37.5` install, then restored the override and ran `eslint .` against a fully fresh `rm -rf node_modules && npm install` — clean both times, only the pre-existing (out-of-scope) react-compiler findings remain.

So: removed `postinstall: "patch-package"`, bumped `eslint-plugin-i18next` to `^6.1.5`, dropped `patch-package` from devDependencies, deleted both patch files, and added the `react.version` override to `eslint.config.mjs`. One less maintenance burden going forward, and one less thing to break on the next dependency bump.

app/src/app/[lng]/cities/[cityId]/GHGI/onboarding/setup/page.tsx
Comment (tatimoreira) [reply app/src/app/[lng]/cities/[cityId]/GHGI/onboarding/setup/page.tsx:311]: Filed [CC-660](https://linear.app/openearth/issue/CC-660/finish-or-remove-unwired-selectedproject-state-in-ghgi-onboarding) — traced it a bit further: `selectedProject` is populated from either the `?project=` URL param or the first entry from `useGetUserProjectsQuery()` under Enterprise Mode, but never read anywhere (no selector UI renders it, and `addInventory`'s mutation type doesn't even accept a `projectId` field). So it's either a half-built project-selection feature for Enterprise Mode onboarding, or leftover scaffolding from an abandoned approach — needs product/design input to know which. Ticket has the details either way.

app/src/backend/hiap/BulkHiapPrioritizationService.ts
tatimoreira: Went ahead and made the fix. `HighImpactActionRanking.ts` now declares `inventory: Inventory` directly on the model class, mirroring the pattern `InventoryValue.ts` already uses for its `belongsTo` associations. The runtime association was already registered (`init-models.ts:1084`, `HighImpactActionRankingModel.belongsTo(Inventory, ...)`), it just wasn't reflected in the class's TS type — so this isn't new behavior, just making the existing association visible to the type checker.

That let me drop the `RankingWithInventoryAndCity` intersection-type workaround in `BulkHiapPrioritizationService.ts` entirely: `ranking.inventory.city` now type-checks directly, since `Inventory.ts` already declares its own `city: City` association the same way. Verified with `eslint` and `tsc --noEmit` on both files — clean.

app/patches/eslint-config-next++eslint-plugin-react+7.37.5.patch
tatimoreira: Done in `495e17d00` — dropped `patch-package` entirely rather than carrying the patches forward:

- Bumped `eslint-plugin-i18next` to `^6.1.5`, which fixes the `getSourceCode()` incompatibility upstream (no patch needed anymore).
- For `eslint-plugin-react` (still stock 7.37.5, bundled via `eslint-config-next`), added a `settings.react.version` override in `eslint.config.mjs` so it skips the auto-detection call that was crashing under ESLint 10's flat config.
- Removed the `postinstall` script and both `.patch` files.

Verified with a clean `eslint .` run against a fully fresh `node_modules` install. Stayed on ESLint 10 (`^10.7.0`) since we didn't end up needing the downgrade.

app/patches/eslint-plugin-i18next+6.1.3.patch
tatimoreira: Yes — bumped to eslint-plugin-i18next@^6.1.5 in 495e17d00, which fixes the getSourceCode() incompatibility upstream. Patch removed, no longer needed.

PR #2937, opened 2026-07-31
[1/2] CC-609: Persist CNB upload Markdown pointers

piotrnowakowski:
## Stack

1. **#2937 — Climate Advisor upload persistence** (this PR)
2. #2938 — CityCatalyst PDF OCR integration
## Summary

- add the follow-up Climate Advisor migration for nullable pre-conversion uploads
- replace inline Markdown persistence with a durable `markdown_s3_key`
- add authoritative upload creation, lifecycle, failure, retry, and pointer-delivery contracts
- make identical delivery idempotent and reject changed Markdown identity
- add the authenticated CityCatalyst Markdown verification client

## Scope

Climate Advisor backend only. This PR contains 11 `climate-advisor` files and no CityCatalyst or frontend changes. It is based on the unchanged CC-608 top branch.

## Validation

- `9 passed` across the focused Concept Note upload and migration pytest suites
- fatal Ruff checks passed
- `git diff --check` passed
- exact-content preservation verified against the original CC-609 head

## Recovery

The original combined CC-609 head is preserved at `codex/CC-609-pre-ca-cc-split-backup-06ba57b` (`06ba57ba541b7c1f94a9764e555e323b3603c11d`).

### DIFF
--- a/climate-advisor/service/app/routes/concept_note_markdown.py
+++ b/climate-advisor/service/app/routes/concept_note_markdown.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import hashlib
 import hmac
+import logging
 import re
 from collections.abc import AsyncIterator
 from typing import Any
@@ -33,6 +34,7 @@ from app.services.citycatalyst_client import (
 )
 
 
+logger = logging.getLogger(__name__)
 router = APIRouter()
 PAGE_MARKER = re.compile(r"<!-- page: (\d+) -->")
 JSON_REQUEST_MAX_BYTES = 16 * 1024
@@ -372,10 +374,17 @@ async def get_concept_note_delivery_context(
     ),
 ) -> JSONResponse | ConceptNoteUploadDeliveryContext:
     """Return delivery metadata to CC after reverse service authentication."""
+    # Authenticate the reverse service without recording its credential.
     expected_key = get_settings().cc_api_key or ""
     supplied_key = request.headers.get("X-CC-Service-Key", "")
     if not expected_key or not hmac.compare_digest(expected_key, supplied_key):
+        logger.warning(
+            "Rejected CC delivery-context request for upload_id=%s",
+            upload_id,
+        )
         return problem(401, "invalid_service_key", "CC service authentication failed")
+
+    # Return only the delivery metadata associated with the opaque upload ID.
     try:
         snapshot = await repository.get_delivery_context(upload_id=upload_id)
     except ConceptNoteMarkdownRepositoryError as exc:
--- a/climate-advisor/service/app/routes/concept_note_markdown.py
+++ b/climate-advisor/service/app/routes/concept_note_markdown.py
@@ -35,6 +35,7 @@ from app.services.citycatalyst_client import (
 
 router = APIRouter()
 PAGE_MARKER = re.compile(r"<!-- page: (\d+) -->")
+JSON_REQUEST_MAX_BYTES = 16 * 1024
 
 
 async def get_citycatalyst_client() -> AsyncIterator[CityCatalystClient]:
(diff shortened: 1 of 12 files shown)


climate-advisor/service/migrations/versions/20260730_120000_concept_note_upload_pointers.py
```diff
@@ -0,0 +1,90 @@
...
+            UPDATE concept_note_uploads
+            SET markdown_text = COALESCE(markdown_text, ''),
+                markdown_sha256 = COALESCE(
+                    markdown_sha256,
+                    repeat('0', 64)
+                ),
+                page_count = COALESCE(page_count, 1)
+            """
+        )
+    )
```
DavidGBbr, line 68: The downgrade backfills `markdown_text = ''` and `markdown_sha256` with 64 zeros for rows that never had inline text (pointer-only uploads). That's not restoring data, it's placeholder data to satisfy the old NOT NULL constraint. Worth a docstring note so nobody mistakes this for recovered content if the old app version runs against it.

climate-advisor/service/app/routes/concept_note_markdown.py
```diff
@@ -97,76 +252,138 @@ async def ingest_concept_note_markdown(
...
+        get_concept_note_markdown_repository
+    ),
+) -> JSONResponse | ConceptNoteUploadDeliveryContext:
+    """Return delivery metadata to CC after reverse service authentication."""
+    expected_key = get_settings().cc_api_key or ""
+    supplied_key = request.headers.get("X-CC-Service-Key", "")
+    if not expected_key or not hmac.compare_digest(expected_key, supplied_key):
+        return problem(401, "invalid_service_key", "CC service authentication failed")
+    try:
+        snapshot = await repository.get_delivery_context(upload_id=upload_id)
```
DavidGBbr, line 389: This route only takes `upload_id` and a shared service key, no per-run or per-user scoping. Given a valid `CC_API_KEY`, it returns `run_id`/`user_id`/`filename` for any upload. UUIDs make enumeration impractical, but is there logging/alerting on this route in case the key ever leaks?

climate-advisor/service/app/routes/concept_note_markdown.py
```diff
@@ -97,76 +252,138 @@ async def ingest_concept_note_markdown(
...
+    except ConceptNoteMarkdownRepositoryError as exc:
+        return problem(exc.status_code, exc.code, str(exc))
+    return ConceptNoteMarkdownResponse(
+        upload_id=snapshot.upload_id,
+        status=snapshot.status,
+    )
+
+
+@router.get(
+    "/concept-note-uploads/{upload_id}/delivery-context",
```
DavidGBbr, line 366: Naming nit: every other route here is `/concept-notes/{run_id}/uploads/...`, this one is `/concept-note-uploads/{upload_id}/...` with no `run_id`. Intentional, to visually separate the service-to-service route from the user-facing ones?

DavidGBbr (approved):
Nice work closing the gap I flagged earlier (no more unbounded markdown_text in the DB), and good test coverage on the lifecycle/conflict paths. A few non-blocking observations below.

climate-advisor/service/migrations/versions/20260730_120000_concept_note_upload_pointers.py
piotrnowakowski: Adjusted in `1ea8165c9`. The `downgrade()` docstring now states explicitly that the empty Markdown and zero hash are placeholders used only to satisfy the legacy `NOT NULL` constraints, and that inline content discarded by the upgrade cannot be recovered. The focused migration test still passes.

climate-advisor/service/app/routes/concept_note_markdown.py
piotrnowakowski: Added in `1ea8165c9`. Rejected delivery-context authentication now emits a dedicated `WARNING` audit event containing the `upload_id` but never the supplied credential. A regression test verifies the `401`, confirms the upload ID is logged, and confirms the rejected key is absent from logs. The README also documents the signal and recommends alerting on repeated warnings or `401 invalid_service_key` responses.

climate-advisor/service/app/routes/concept_note_markdown.py
piotrnowakowski: Yes, intentional. This is the reverse service-to-service route: CC starts with the opaque `upload_id` and uses the service credential to recover the associated run/user scope. The user-facing routes remain under `/concept-notes/{run_id}/uploads/...`. I documented that distinction in the README in `1ea8165c9`; the route itself is unchanged.

PR #2938, opened 2026-07-31
[2/2] CC-609: Connect CNB uploads to PDF OCR

piotrnowakowski:
## Stack

1. #2937 — Climate Advisor upload persistence
2. **#2938 — CityCatalyst PDF OCR integration** (this PR)
## Summary

- add authenticated CNB upload, status, retry, and internal Markdown-read routes
- store source PDFs under deterministic S3 keys derived from `upload_id`
- reuse the existing `PdfOcrJob` and support both inventory and Concept Note upload sources
- deliver stable Markdown object references to Climate Advisor
- separate OCR retry from delivery-only retry
- update the Concept Note Builder architecture documentation

## Scope

CityCatalyst backend, tests, and documentation only. This PR contains 13 files. It adds no frontend files, CityCatalyst models, or CityCatalyst migrations.

## Validation

- `24 passed` across four focused Jest suites
- `npx tsc --noEmit` passed
- `git diff --check` passed
- exact-content preservation verified across all 24 original CC-609 paths
- two preserved Jest files retain their existing Prettier warnings; no formatting changes were introduced during the split

## Recovery

The original combined CC-609 head is preserved at `codex/CC-609-pre-ca-cc-split-backup-06ba57b` (`06ba57ba541b7c1f94a9764e555e323b3603c11d`).

### DIFF
--- a/app/src/app/api/v1/concept-notes/[runId]/uploads/[uploadId]/route.ts
+++ b/app/src/app/api/v1/concept-notes/[runId]/uploads/[uploadId]/route.ts
@@ -1,0 +1,62 @@
+import createHttpError from "http-errors";
+import { NextResponse } from "next/server";
+import { z } from "zod";
+
+import {
+  loadConceptNoteRunCity,
+  loadConceptNoteUpload,
+} from "@/backend/ConceptNoteUploadService";
+import {
+  getConceptNotePdfOcrJob,
+  normalizeConceptNotePdfOcrStatus,
+} from "@/backend/PdfOcrService";
+import { PermissionService } from "@/backend/permissions/PermissionService";
+import { apiHandler } from "@/util/api";
+
+const paramsSchema = z.object({
+  runId: z.string().uuid(),
+  uploadId: z.string().uuid(),
+});
+
+export const GET = apiHandler(async (req, { session, params }) => {
+  if (!session?.user?.id) {
+    throw new createHttpError.Unauthorized("Authentication required");
+  }
+  const { runId, uploadId } = paramsSchema.parse(params);
+  const userId = session.user.id;
+  const currentRequestId = req.headers.get("x-request-id")?.trim() || undefined;
+  const cityId = await loadConceptNoteRunCity({
+    runId,
+    userId,
+    requestId: currentRequestId,
+  });
+  await PermissionService.canAccessCity(session, cityId, {
+    includeResource: false,
+  });
+  const upload = await loadConceptNoteUpload({
+    runId,
+    uploadId,
+    userId,
+    requestId: currentRequestId,
+  });
+  const job = await getConceptNotePdfOcrJob(uploadId);
+  const workerState = job ? normalizeConceptNotePdfOcrStatus(job) : null;
+  const status =
+    upload.status === "ready" ? "ready" : workerState?.status || upload.status;
+  const errorCode =
+    status === "failed"
+      ? workerState?.errorCode || upload.error_code || undefined
+      : undefined;
+
+  return NextResponse.json({
+    upload_id: upload.upload_id,
+    run_id: upload.run_id,
+    status,
+    filename: upload.filename,
+    source_label: upload.source_label || null,
+    page_count: upload.page_count || null,
+    ...(errorCode ? { error_code: errorCode } : {}),
+    received_at: upload.received_at,
+    completed_at: upload.completed_at || null,
+  });
+});
--- a/app/src/app/api/v1/concept-notes/[runId]/uploads/[uploadId]/route.ts
+++ b/app/src/app/api/v1/concept-notes/[runId]/uploads/[uploadId]/route.ts
@@ -28,18 +28,18 @@
  *             schema:
  *               type: object
  *               required:
- *                 - upload_id
- *                 - run_id
+ *                 - uploadId
+ *                 - runId
  *                 - status
  *                 - stage
- *                 - can_retry
+ *                 - canRetry
  *                 - filename
- *                 - received_at
+ *                 - receivedAt
  *               properties:
- *                 upload_id:
+ *                 uploadId:
  *                   type: string
  *                   format: uuid
- *                 run_id:
+ *                 runId:
  *                   type: string
  *                   format: uuid
  *                 status:
@@ -48,25 +48,25 @@
  *                 stage:
  *                   type: string
  *                   enum: [upload, ocr, delivery, complete]
- *                 can_retry:
+ *                 canRetry:
  *                   type: boolean
- *                 retry_kind:
+ *                 retryKind:
  *                   type: string
  *                   enum: [ocr, delivery]
  *                 filename:
  *                   type: string
- *                 source_label:
+ *                 sourceLabel:
  *                   type: string
  *                   nullable: true
- *                 page_count:
+ *                 pageCount:
  *                   type: integer
  *                   nullable: true
- *                 error_code:
+ *                 errorCode:
  *                   type: string
- *                 received_at:
+ *                 receivedAt:
  *                   type: string
  *                   format: date-time
- *                 completed_at:
+ *                 completedAt:
  *                   type: string
  *                   format: date-time
  *                   nullable: true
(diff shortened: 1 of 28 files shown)


app/src/app/api/v1/concept-notes/[runId]/uploads/[uploadId]/route.ts
```diff
@@ -0,0 +1,147 @@
...
+    stage,
+    can_retry: canRetry,
+    ...(retryKind ? { retry_kind: retryKind } : {}),
+    filename: upload.filename,
+    source_label: upload.source_label || null,
+    page_count: upload.page_count || null,
+    ...(errorCode ? { error_code: errorCode } : {}),
+    received_at: upload.received_at,
+    completed_at: upload.completed_at || null,
+  });
```
Comment (lemilonkh) [inline app/src/app/api/v1/concept-notes/[runId]/uploads/[uploadId]/route.ts:146]: Our API endpoints usually return data in camelCase variable name formatting instead of snake_case. Might be good to adjust for consistency before this is deployed/ used in the UI and we depend on it.
Also make sure to keep the API docs above consistent in case it's changed.

app/src/app/api/v1/concept-notes/[runId]/uploads/route.ts
```diff
@@ -0,0 +1,303 @@
...
+  const identity = JSON.stringify([
+    "citycatalyst-concept-note-upload-v1",
+    args.runId,
+    args.userId,
+    args.filename,
+    args.sourceLabel,
+    contentSha256,
+  ]);
+  return uuidv5(identity, uuidv5.URL);
+}
```
Comment (lemilonkh) [inline app/src/app/api/v1/concept-notes/[runId]/uploads/route.ts]: I think even a normal UUID (v4) might be sufficient here (generated with the `randomUUID()` function from the `node:crypto` package) as we are highly unlikely to have conflicts even then. Which would make this route quite a bit faster/ CPU efficient I think as the uploaded file (presumably a large binary file like a PDF) doesn't have to be hashed.

app/src/app/api/v1/concept-notes/[runId]/uploads/route.ts
```diff
@@ -0,0 +1,303 @@
...
+  if (fileEntry.size === 0) {
+    throw new createHttpError.UnprocessableEntity("PDF file is empty");
+  }
+  if (fileEntry.size > INVENTORY_IMPORT_MAX_FILE_SIZE_BYTES) {
+    throw new createHttpError.PayloadTooLarge(
+      "PDF exceeds the 20 MiB upload limit",
+    );
+  }
+  const fileBuffer = Buffer.from(await fileEntry.arrayBuffer());
+  if (fileBuffer.subarray(0, 5).toString("ascii") !== "%PDF-") {
```
Comment (lemilonkh) [inline app/src/app/api/v1/concept-notes/[runId]/uploads/route.ts:160]: This is quite intense as it makes a copy of the entire (potentially large) file in memory.
It might be simpler do do it like this:

```suggestion
  const fileBuffer = await fileEntry.arrayBuffer();
  if (fileBuffer.slice(0, 5).toString("ascii") !== "%PDF-") {
```

Haven't tested this but it's at least one file copy less in memory...

app/src/app/api/v1/concept-notes/[runId]/uploads/route.ts
```diff
@@ -0,0 +1,303 @@
...
+    }
+    return NextResponse.json(
+      {
+        upload_id: uploadId,
+        status: state.status,
+        stage: state.stage,
+        can_retry: state.canRetry,
+        ...(state.retryKind ? { retry_kind: state.retryKind } : {}),
+      },
+      { status: state.status === "ready" ? 200 : 202 },
```
Comment (lemilonkh) [inline app/src/app/api/v1/concept-notes/[runId]/uploads/route.ts]: Ideally API response key names are in camelCase instead of snake_case.

app/src/app/api/v1/concept-notes/[runId]/uploads/route.ts
```diff
@@ -0,0 +1,303 @@
...
+  const state = normalizeConceptNotePdfOcrStatus(job);
+
+  return NextResponse.json(
+    {
+      upload_id: uploadId,
+      status: state.status,
+      stage: state.stage,
+      can_retry: state.canRetry,
+      ...(state.retryKind ? { retry_kind: state.retryKind } : {}),
+    },
```
Comment (lemilonkh) [inline app/src/app/api/v1/concept-notes/[runId]/uploads/route.ts:240]: camelCase is preferred over snake_case.

app/src/backend/ConceptNoteUploadService.ts
```diff
@@ -0,0 +1,96 @@
...
+export const conceptNoteUploadSchema = z.object({
+  upload_id: z.string().uuid(),
+  run_id: z.string().uuid(),
+  status: z.enum(["queued", "processing", "ready", "failed"]),
+  filename: z.string(),
+  source_label: z.string().nullable().optional(),
+  page_count: z.number().int().positive().nullable().optional(),
+  error_code: z.string().nullable().optional(),
+  received_at: z.string(),
+  completed_at: z.string().nullable().optional(),
```
lemilonkh: camelCase names preferred for route bodies as a convention

app/src/backend/PdfOcrService.ts
```diff
@@ -58,6 +66,28 @@ export async function enqueueInventoryPdfOcr(
...
+      sourceId: uploadId,
+      status: "queued",
+      attemptCount: 0,
+      runAfter: new Date(),
+      deliveryTarget: "climate_advisor",
+      deliveryStatus: "pending",
+      deliveryAttemptCount: 0,
+    },
+  });
+  return job;
```
lemilonkh, line 88: If it exists already, should it be set to `status: "queued"` here in case it is e.g. retried?

lemilonkh (approved):
Awesome, thanks 👍 ✨

app/src/backend/PdfOcrService.ts
piotrnowakowski: Good question. We intentionally leave an existing job unchanged here because this function creates or reuses the initial OCR job; it is not the retry path. Explicit retries go through `retryConceptNotePdfOcr`, which resets failed OCR jobs or retries delivery separately. Setting every existing job back to `queued` here could rerun already successful OCR and would break delivery-only retries and idempotent upload replays. I can rename this function to make that distinction clearer if helpful.

lemilonkh: Function name is clear, was just wondering because it uses findOrCreate

app/src/app/api/v1/concept-notes/[runId]/uploads/route.ts
Comment (piotrnowakowski) [reply app/src/app/api/v1/concept-notes/[runId]/uploads/route.ts:160]: Tested this on Node 22.21.1. `Buffer.from(ArrayBuffer)` creates a zero-copy view that shares the same backing memory; with a 20 MiB file it added 0 ArrayBuffer bytes. The allocation comes from `fileEntry.arrayBuffer()`, which both versions still call, and `InventoryFileStorageService.putFile()` requires a `Buffer`.

The suggested version also does not type-check (`TS2554`) and `ArrayBuffer.slice(...).toString("ascii")` returns `[object ArrayBuffer]`, causing valid PDFs to be rejected. The existing upload tests pass with the current implementation, so I will keep the current implementation intact.

app/src/app/api/v1/concept-notes/[runId]/uploads/[uploadId]/route.ts
Comment (piotrnowakowski) [reply app/src/app/api/v1/concept-notes/[runId]/uploads/[uploadId]/route.ts:146]: Addressed on the current PR head (`71b45fb4`). The status endpoint now returns camelCase fields including `uploadId`, `runId`, `canRetry`, `retryKind`, `sourceLabel`, `pageCount`, `errorCode`, `receivedAt`, and `completedAt`. I updated the Swagger schema to use the same names and added focused response coverage.

app/src/app/api/v1/concept-notes/[runId]/uploads/route.ts
Comment (piotrnowakowski) [reply app/src/app/api/v1/concept-notes/[runId]/uploads/route.ts]: Addressed on the current PR head (`71b45fb4`). The route now uses `randomUUID()` from `node:crypto`; the content hash and UUID v5 generation were removed. Each accepted request receives a fresh UUID v4, and the upload tests verify that repeated identical files receive different identities.

app/src/app/api/v1/concept-notes/[runId]/uploads/route.ts
Comment (piotrnowakowski) [reply app/src/app/api/v1/concept-notes/[runId]/uploads/route.ts]: Addressed on the current PR head (`71b45fb4`). The upload response now uses `uploadId`, `canRetry`, and optional `retryKind`, with the Swagger response schema and Jest assertions updated to match.

app/src/app/api/v1/concept-notes/[runId]/uploads/route.ts
Comment (piotrnowakowski) [reply app/src/app/api/v1/concept-notes/[runId]/uploads/route.ts:240]: Addressed across the upload route on the current PR head (`71b45fb4`). Public request and response fields now use camelCase, including `sourceLabel`, `uploadId`, `canRetry`, and `retryKind`. Snake case remains only inside the private Climate Advisor wire adapter, where it matches the upstream API contract.

app/src/backend/ConceptNoteUploadService.ts
piotrnowakowski: Addressed on the current PR head (`71b45fb4`). `ConceptNoteUploadService` now exposes camelCase service objects and arguments. A dedicated wire schema converts Climate Advisor's snake_case responses into camelCase, and outbound `errorCode` is converted back to the CA `error_code` contract only at that boundary. Added adapter tests cover both directions.

PR #2942, opened 2026-08-03
feat(ui): CC-518 password settings improvements

lemilonkh:
Removed subtitle since it was redundant
Added password strength
Updated button disabled style
Changed section copy to just “Password”


cephaschapa (approved):
Looks great 🙌

PR #2943, opened 2026-08-03
feat(ui): CC-587 use drawer organization selection in settings

lemilonkh:
## Summary

Makes the account and projects tabs in settings also use the organization state from Redux, which is set by the top selector in the navigation bar drawer

Fixes translation string issues CC-600 and CC-602

Removes OrganizationSelector component, so also fixes CC-601 through that as it is no longer used

## How to test

Go to settings.
Select an organization in the top dropdown in the drawer title (opened from the hamburger menu in the navigation bar)
See each settings tab that is organization related adjust to it

## Ticket

CC-587, CC-600, CC-601, CC-602


cephaschapa (approved):
Durch den geschickten Einsatz von Haken wird die Arbeit erleichtert. ✨

PR #2950, opened 2026-08-03
CC-471: Show hazard instead of sector on adaptation action cards

tatimoreira:
## Summary

Based on the discussion with @amanda-eames adaptation actions are classified by hazard, not sector. `ClimateActionCard` was always rendering a sector label, which is often empty for adaptation actions and showed up as a missing/N-A string. The card now shows hazard for adaptation actions and sector for mitigation actions, matching the distinction `ActionDrawer` already makes.

## Changes

- `ClimateActionCard.tsx`: branch on `action.type` to read `action.hazards[0]` (adaptation) vs `action.sectors[0]` (mitigation) for both the label and value
- Added `hazard-name` translation key to `en`/`de`/`es`/`fr`/`pt` `hiap.json`
image

## How to test

1. Open the Top Actions view for a city and switch to the Adaptation tab.
2. Confirm each action card shows a "Hazard" label with a real hazard value (e.g. Droughts, Floods) instead of "Sector" / N/A.
3. Switch to the Mitigation tab and confirm cards still show "Sector" as before.

## Ticket

CC-471

### DIFF
--- a/app/src/components/ClimateActionCard.tsx
+++ b/app/src/components/ClimateActionCard.tsx
@@ -63,7 +63,9 @@ export const ClimateActionCard = ({
   };
 
   const reductionLevel = getReductionPotentialLevel();
+  const isAdaptation = action.type === "adaptation";
   const actionSector = action.sectors?.[0];
+  const actionHazard = action.hazards?.[0];
 
   return (
     <Card.Root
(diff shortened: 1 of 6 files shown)


lemilonkh (approved):
Awesome fix, thanks for reaching out and figuring this out ✨

PR #2955, opened 2026-08-04
CC-614: Fix year selectors in inventory setup screens

tatimoreira:
## Summary
- Population year selects (`cityPopulationYear`, `regionPopulationYear`, `countryPopulationYear` in `set-population-data-step.tsx`) now render "Select year" as a `disabled hidden` placeholder option instead of a normal, selectable `<option value="">` — Chakra's `placeholder` prop alone doesn't prevent reselection.
- Removed the green checkmark (`MdCheck` in `InputGroup`'s `endElement`) and its dead validation state from all four year selectors (the 3 population fields + the "Inventory year" field in `set-inventory-details-step.tsx` and `GHGIInventoryDetailsStep.tsx`). That wrapper had `pointer-events: auto` and sat on top of the dropdown's chevron, silently swallowing clicks meant for the arrow.
- Found along the way: the "Inventory year" checkmark in `GHGIInventoryDetailsStep.tsx` was gated on `let year;` — a variable declared but never assigned — so it was dead code that could never have rendered.

## Test plan
- [ ] Onboarding "Add population data" step: leave each of the 3 year fields empty → placeholder shown, not selectable from the dropdown list
- [ ] Onboarding "Add population data" step: pick a year on each of the 3 fields → year shows, no checkmark, "Select year" no longer reachable
- [ ] Onboarding "Inventory details" step (both onboarding entry points: `cities/onboarding/setup` and `cities/[cityId]/GHGI/onboarding/setup`) → click directly on the dropdown arrow (not just the text) → dropdown opens
- [ ] `npx tsc --noEmit` and `npx eslint` pass on all three changed files (verified — no new errors vs. baseline)

### DIFF
--- a/app/src/app/[lng]/cities/[cityId]/GHGI/onboarding/GHGIInventoryDetailsStep.tsx
+++ b/app/src/app/[lng]/cities/[cityId]/GHGI/onboarding/GHGIInventoryDetailsStep.tsx
@@ -16,10 +16,9 @@ import {
   Link,
   Text,
 } from "@chakra-ui/react";
-import { MdCheck, MdWarning } from "react-icons/md";
+import { MdWarning } from "react-icons/md";
 import { Trans } from "react-i18next";
 import { CustomRadio, RadioGroup } from "@/components/ui/custom-radio";
-import { InputGroup } from "@/components/ui/input-group";
 import {
   SelectContent,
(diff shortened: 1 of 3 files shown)


lemilonkh (approved):
Great 👍🏻

PR #2887, opened 2026-07-23
docs(meed): MEED+ / HIAP v3 → CityCatalyst module migration assessment

carlosgraffi:
> **How to use this PR.** This is a **planning document for review and decision, not a code review** — there is no code here, and merging is not the goal. Please read the three docs, comment inline where you disagree or want detail, and let's use the meeting to work through the checkboxes together:
> - **Open questions (§7)** and the **Week-0 checklist (§5.7)** in `MeedModuleMigration.md` are tickable — we resolve those live and turn them into work items.
> - Each is tagged with a *suggested* lead; those are starting points to confirm or reassign, not assignments.

## What

Planning docs for migrating the **MEED+ / HIAP v3 prioritizer** from its standalone repo
([`meed-mitigation-prioritizer-frontend`](https://github.com/Open-Earth-Foundation/meed-mitigation-prioritizer-frontend))
into CityCatalyst as a native module — visible in the Journey Navigator's Modules screen, behind
CityCatalyst auth, and reading emissions from the Inventories (GHGI) module.

**Docs only. No code. Everything here is a proposal for review, not an approved plan.**

| File | Contents |
|---|---|
| [`docs/MeedModuleMigration.md`](../blob/docs/meed-module-migration/docs/MeedModuleMigration.md) | Assessment, analysis of the full-stack lead's initial read, effort estimate, proposed workstreams, open questions, risks |
| [`docs/MeedModuleMigration-Inventory.md`](../blob/docs/meed-module-migration/docs/MeedModuleMigration-Inventory.md) | Mechanical audit — every external call site, `localStorage` key, bundled data file and screen, each with a proposed target |
| [`docs/MeedModuleMigration-Interop.md`](../blob/docs/meed-module-migration/docs/MeedModuleMigration-Interop.md) | Proposed contract for modules reading each other's data — MEED+ reading GHGI inventories is the first real case |

## Headline findings

1. **The backend already exists.** [`hiap-meed/`](../tree/develop/hiap-meed) is already in this
   monorepo, deployed (`hiap-meed-service-dev`), with CI and k8s manifests. What's missing is a
   CityCatalyst service + persistence + proxy layer over it — not a prioritization backend.
   This makes the backend scope meaningfully smaller than assumed, and moves the critical path to
   the **frontend**.

2. **The screen-vs-report discrepancies have a single identifiable cause.** The prototype makes
   **16 `fetch()` calls across 7 files** to the Global API directly from the browser, while
[...]

PR #2956, opened 2026-08-04
feat(meed): Actions & Plans v2 as a native CityCatalyst module

carlosgraffi:
## Actions & Plans v2 (MEED) — native CityCatalyst module

Draft. Ports the [MEED+ / HIAP v3 prioritizer](https://github.com/Open-Earth-Foundation/meed-mitigation-prioritizer-frontend) from a standalone Vite SPA into a native module at `[lng]/cities/[cityId]/MEED/[inventory]/…`, per Milan's folder structure and the plan in `docs/MeedModuleMigration.md` (PR #2887).

**Everything is behind the `MEED_MODULE` feature flag, which is off by default.** With the flag unset, the app behaves exactly as on `develop`.

Full write-up: **[`docs/MeedModuleImplementation.md`](docs/MeedModuleImplementation.md)** — this description is the summary.

### What works

Eleven screens, per city and per inventory, on real data:

- **Emissions** from CityCatalyst's own inventory endpoints — the first real cross-module read, MEED ← GHGI
- **Action catalog** (102 actions), **city attributes**, **policy scores** (7 national plans for São Paulo) and **climate-finance feasibility** (52 self-deliverable / 27 co-finance / 7 finance+support) from the Global API, all proxied **server-side** through `…/api/v1/city/[city]/modules/meed/…` rather than called from the browser as the prototype did

Navigation is hub-and-spoke, not a linear wizard: every step is reachable from the hub, the stepper, the breadcrumb or the URL, and `?from=` means editing a step from pre-flight returns you to pre-flight. Only the final generate action is gated, and it says what is missing.

### What is not wired — the prioritizer

There is **no client, no proxy route and no env var** for the four `hiap-meed` endpoints (`/v1/prioritize`, `/v1/prioritize/exclusions/preview`, `/v1/reports/output-plan`, `/v1/explanations/translate`). Their types are already defined in `app/src/util/types/meed.ts` and the frontend is built against that contract.

The service **is** deployed: `hiap-meed-service-dev`, namespace `default`, port 80 → 8000 — but `type: ClusterIP` with no ingress, so it is reachable from a CityCatalyst pod at `http://hiap-meed-service-dev` and needs `kubectl port-forward` from a laptop. §3 of the doc lists exactly what wiring it takes, including the GPC transform that is the correctness-critical piece.

Until then, `meedMockRanking.ts` builds a stand-in ranking from the **live catalog** behind a second flag, `MEED_MOCK_RANKING` — real action ids, names, sectors and co-benefits; only the three pillar scores are synthetic, derived deterministically from the action id. Results are tagged `isMock`. Delete the file when the real endpoint lands.

### One root-level change is included

`radii.full` was defined as `50%` instead of a stadium radius, so every wide, short element — progress tracks, slider tracks and thumbs, pill chips — rendered as an **ellipse**, app-wide, including inside Chakra's own slider and progress recipes. Fixed in `803b5da` (its own branch, `fix/theme-radii-full`, merged here) and easy to split back out. A square element clamps to the same circle either way; verified on the GHGI page that everything resolving `full` there is square and renders identically.

`Hero` also gained an optional `moduleLabel` prop — purely additive; without it, behaviour is byte-identical.

§8 of the doc lists six further platform issues found and deliberately **not** fixed here, each needing its own ticket. The largest: **the theme's `tag`/`card`/`tooltip`/`accordion`/`progress`/`switch`/`tabs` recipes are registered under `recipes` instead of `slotRecipes`**, so `tagRecipe`'s `success`/`warning`/`low`/`medium`/`high` variants have never rendered anywhere in CityCatalyst. That is why code reaches for `colorPalette` and lands on stock greys, and why this module carries its own `MeedStatusTag`.

### Reviewing it

`npx tsc --noEmit` and `npm run build` are both clean. (ESLint is broken repo-wide on `develop` — eslint 10 vs the `eslint-config-next` react plugin — so it isn't a usable gate.)

To see it: add `MEED_MODULE,MEED_MOCK_RANKING` to `NEXT_PUBLIC_FEATURE_FLAGS`, grant the module to your project, and run the rename seeder — §9 of the doc has the commands and explains why a new dated seeder is needed (`modules.json` is not read at runtime).

### Known gaps

Translations exist in `en` only (deliberately last, so the copy stops moving first; `fallbackLng` is `en`, so other languages show English, not raw keys). State lives in `localStorage` and should become the `MeedPreferences`/`MeedExclusion`/`MeedRanking` tables — `useMeedRanking` is the single swap point. The processing screen runs a scripted animation not gated on any real request. Report generation needs the async job + cron pattern. No keyboard-only pass or Playwright happy path yet.


app/seeders/20260804090000-rename-meed-module.cjs
```diff
@@ -0,0 +1,54 @@
+"use strict";
+
+const { parseJsonFile, bulkUpsert } = require("./util/util.cjs");
+
+const MEED_MODULE_ID = "9f622243-fba8-4f32-a000-ce6e66982bd1";
+
+/**
+ * Re-upserts the MEED module row so the display-name change in
+ * seed-data/modules/modules.json reaches environments where the original
+ * modules seeder (20250923103314) has already run. Sequelize records executed
```
lemilonkh, line 10: The seeders run each time we deploy, this migration file is not necessary I think.

app/migrations/20260812150000-register-meed-module.cjs
lemilonkh, line 1: This file can be removed, the deployed environments do in fact call the `k8s/cc-seed.yml` file (e.g. in `.github/workflows/web-develop.yml` Line 454).

app/src/backend/meed/MeedGlobalApiService.ts
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * The prototype called these six endpoints directly from the browser (16 call
+ * sites — see docs/MeedModuleMigration-Inventory.md §1); routing them through
+ * here gives the module one API boundary and keeps filtering consistent with
+ * what `hiap-meed` reads server-side.
+ *
+ * All methods return `null` on upstream errors or missing data — screens
+ * render precondition/empty states rather than failing.
+ */
+export class MeedGlobalApiService {
```
lemilonkh, line 15: We are fetching this data via the HIAP-MEED service so it can apply the same filtering criteria as are used for the ranking.
This file can be removed.

app/src/app/api/v1/city/[city]/modules/meed/actions/route.ts
Comment (lemilonkh) [inline app/src/app/api/v1/city/[city]/modules/meed/actions/route.ts:1]: We have this same API route outside of the `modules` folder (in `app/src/app/api/v1/city/[city]/meed`). File can be removed.

app/src/app/api/v1/city/[city]/modules/meed/city-attributes/route.ts
Comment (lemilonkh) [inline app/src/app/api/v1/city/[city]/modules/meed/city-attributes/route.ts:1]: We have this same API route outside of the `modules` folder (in `app/src/app/api/v1/city/[city]/meed`). File can be removed.

app/src/app/api/v1/city/[city]/modules/meed/finance/feasibility/route.ts
Comment (lemilonkh) [inline app/src/app/api/v1/city/[city]/modules/meed/finance/feasibility/route.ts:1]: We have this same API route outside of the `modules` folder (in `app/src/app/api/v1/city/[city]/meed`). File can be removed.

app/src/app/api/v1/city/[city]/modules/meed/finance/follow/route.ts
Comment (lemilonkh) [inline app/src/app/api/v1/city/[city]/modules/meed/finance/follow/route.ts:1]: We don't need this API route as discussed in Slack. File can be removed, references to this route should use the `finance/projects` and `finance/opportunities` routes instead.

app/src/app/api/v1/city/[city]/modules/meed/policy-scores/route.ts
Comment (lemilonkh) [inline app/src/app/api/v1/city/[city]/modules/meed/policy-scores/route.ts:1]: We have this same API route outside of the `modules` folder (in `app/src/app/api/v1/city/[city]/meed`). File can be removed.

app/src/services/api.ts
```diff
@@ -220,6 +232,139 @@ export const api = createApi({
...
+      }),
+      getMeedFinanceLink: builder.query<
+        unknown,
+        { cityId: string; link: string }
+      >({
+        query: ({ cityId, link }) =>
+          `city/${cityId}/modules/meed/finance/follow?link=${encodeURIComponent(link)}`,
+        transformResponse: (response: { data: unknown }) => response.data,
+        providesTags: ["Meed"],
+      }),
```
lemilonkh, line 270: All of these route accessors should be removed. All references to them should be replaced with the `getMeedReference*` versions. References to `getMeedFinanceLink` should be replaced with the respective reference finance projects and reference finance opportunities route accessors.

app/src/services/api.ts
```diff
@@ -220,6 +232,139 @@ export const api = createApi({
...
+      // ─── MEED reference data (city-scoped routes) ───
+      //
+      // These six replace the `modules/meed/*` proxies above. They are pure
+      // pass-throughs to hiap-meed, so unlike the ranking route their payloads
+      // are snake_case with a `meta`/`warnings` envelope — see the contract
+      // types. The older proxies stay until their consumers are migrated;
+      // that migration also retires `finance/follow`, whose guard only permits
+      // `/api/v1/cities/` while the real links are `/api/v1/climate-finance/`,
+      // so both of its calls 400 and the cards silently render "no data".
```
lemilonkh, line 294: Would be good to implement this transition as mentioned above. All routes are available and have the same return types as the previous implementations.

app/src/services/api.ts
```diff
@@ -220,6 +232,139 @@ export const api = createApi({
...
+      >({
+        query: ({ cityId, actionId }) =>
+          `city/${cityId}/meed/finance/projects?actionId=${encodeURIComponent(
+            actionId,
+          )}`,
+        transformResponse: (r: {
+          data: MeedReferenceFinanceProjectsResponse;
+        }) => r.data,
+        providesTags: ["Meed"],
+      }),
```
lemilonkh, line 358: Once the other route accessors are removed, all of these could be renamed to remove the `Reference` part of their names for simplicity/ clarity. Make sure to update the references to them as well of course.

app/src/app/[lng]/cities/[cityId]/MEED/meedLocalState.ts
```diff
@@ -0,0 +1,176 @@
+/**
+ * TEMPORARY client-side persistence for MEED+ wizard choices, keyed per
+ * inventory. This replaces the prototype's `hiap:{locode}:*` localStorage
+ * families 1:1 until the CityCatalyst backend persistence lands
+ * (MeedPreferences / MeedExclusion / MeedRanking models — Milan's track,
+ * see docs/MeedModuleMigration.md §6.B). Swap these helpers for RTK Query
+ * mutations then; the call sites should not need to change shape.
```
Comment (lemilonkh) [inline app/src/app/[lng]/cities/[cityId]/MEED/meedLocalState.ts:7]: Can we create a follow-up ticket for this?
Not sure how urgent it is but ideally we don't rely on local storage for too long, otherwise users will have a different experience when using a different device or browser.

Comment (lemilonkh) [reply app/src/app/[lng]/cities/[cityId]/MEED/meedLocalState.ts:7]: Also an issue here is that the API routes don't return data instantaneously, so we will need to implement a loading state for all of these features where they are referenced (once the database models and API routes are implemented).

app/src/app/[lng]/cities/[cityId]/MEED/meedMockRanking.ts
```diff
@@ -0,0 +1,156 @@
...
+ * Every action id, name, sector, timeline and co-benefit in the output is real
+ * — it comes from `/api/v1/action-pathways` via the module's own proxy. Only
+ * the three pillar scores are synthetic, and they are derived deterministically
+ * from the action id so the ordering is stable between runs rather than
+ * reshuffling on every render.
+ *
+ * This is strictly a review aid. It is only ever called behind the
+ * `MEED_MOCK_RANKING` feature flag, which is off by default, and the ranking it
+ * produces is tagged `isMock` so the UI can say so out loud. Delete this file
+ * when `POST /v1/prioritize` is wired up.
```
Comment (lemilonkh) [inline app/src/app/[lng]/cities/[cityId]/MEED/meedMockRanking.ts:20]: Can you create a follow-up ticket for removing the mock files once everything is integrated?

app/src/app/[lng]/cities/[cityId]/MEED/meedRankingAdapter.ts
```diff
@@ -0,0 +1,253 @@
...
+    ownership_category: str(raw.ownershipCategory),
+    ownership_description: str(ownership?.en),
+    ownership_description_es: str(ownership?.es),
+    restrictions_category: str(raw.restrictionsCategory),
+    restrictions_description: str(restrictions?.en),
+    restrictions_description_es: str(restrictions?.es),
+    legal_justification: str(justification?.es) ?? str(justification?.en),
+    legal_justification_en: str(justification?.en),
+    legal_references: references,
+  };
```
Comment (lemilonkh) [inline app/src/app/[lng]/cities/[cityId]/MEED/meedRankingAdapter.ts:103]: It would be better to use the language maps directly since it's easier to use for translation and adding more languages in the future without requiring lots of code changes.
Also I would prefer using the canonical camelCase spelling of our database models for clarity, but that's not as urgent.

Maybe it can be done in this PR or alternatively a follow-up PR.

app/src/app/[lng]/cities/[cityId]/MEED/meedRankingAdapter.ts
```diff
@@ -0,0 +1,253 @@
...
+  return {
+    action_id: raw.actionId as string,
+    rank: num(raw.rank),
+    final_score: num(raw.finalScore),
+    impact_score: num(raw.impactScore),
+    alignment_score: num(raw.alignmentScore),
+    feasibility_score: num(raw.feasibilityScore),
+    // Already snake_case inside — the screens read it as-is.
+    evidence_summary: raw.evidenceSummary ?? {},
+    ...(raw.explanations ? { explanations: raw.explanations } : {}),
```
Comment (lemilonkh) [inline app/src/app/[lng]/cities/[cityId]/MEED/meedRankingAdapter.ts:75]: I would prefer using the canonical camelCase spelling of our database models for clarity, but that's not as urgent.

app/src/app/[lng]/cities/[cityId]/MEED/useMeedRanking.ts
```diff
@@ -0,0 +1,112 @@
...
+    setHasReadStore(true);
+  }, [inventoryId]);
+
+  useEffect(() => {
+    refresh();
+    window.addEventListener("focus", refresh);
+    window.addEventListener(MEED_STATE_CHANGED_EVENT, refresh);
+    return () => {
+      window.removeEventListener("focus", refresh);
+      window.removeEventListener(MEED_STATE_CHANGED_EVENT, refresh);
```
Comment (lemilonkh) [inline app/src/app/[lng]/cities/[cityId]/MEED/useMeedRanking.ts:70]: This is a little bit problematic to see - we are going outside of React and Redux to send events between specific components. Ideally we would use Redux/ RTK for wiring this up instead of relying on custom window event listeners.
Nice to have/ tech debt though.

app/src/i18n/locales/es/dashboard.json
```diff
@@ -315,6 +315,7 @@
   "legend": "Leyenda",
   "breadcrumb-tools": "Herramientas",
   "breadcrumb-hiap": "Actions & Plans",
+  "breadcrumb-meed": "Acciones y Planes v2",
```
lemilonkh: Might be good to use the same syntax here (`y` vs. `&`).

app/src/i18n/locales/es/navigation.json
```diff
@@ -7,5 +7,6 @@
   "account-frozen-warning-text": "Su cuenta está congelada. Por favor, póngase en contacto con un administrador o con <bold>{{ email }}</bold> para reactivarla",
   "page-title-ghg-inventories": "Inventarios de GEI",
   "page-title-hiap": "Acciones & Planes",
+  "page-title-meed": "Acciones y Planes v2",
```
lemilonkh: Might be good to use the same syntax here (`y` vs. `&`).

app/src/i18n/locales/fr/dashboard.json
```diff
@@ -312,6 +312,7 @@
   "legend": "Légende",
   "breadcrumb-tools": "Outils",
   "breadcrumb-hiap": "Actions & Plans",
+  "breadcrumb-meed": "Actions et Plans v2",
```
lemilonkh: Might be good to use the same syntax here (`et` vs. `&`).

app/src/i18n/locales/fr/navigation.json
```diff
@@ -7,5 +7,6 @@
   "account-frozen-warning-text": "Votre compte est gelé. Veuillez contacter un administrateur ou <bold>{{ email }}</bold> pour le réactiver",
   "page-title-ghg-inventories": "Inventaires de GEI",
   "page-title-hiap": "Actions & Plans",
+  "page-title-meed": "Actions et Plans v2",
```
lemilonkh: Might be good to use the same syntax here (`et` vs. `&`).

app/src/i18n/locales/pt/dashboard.json
```diff
@@ -313,6 +313,7 @@
   "legend": "Legenda",
   "breadcrumb-tools": "Ferramentas",
   "breadcrumb-hiap": "Actions & Plans",
+  "breadcrumb-meed": "Ações e Planos v2",
```
lemilonkh: Might be good to use the same syntax here (`e` vs. `&`).

app/src/i18n/locales/pt/navigation.json
```diff
@@ -7,5 +7,6 @@
   "account-frozen-warning-text": "Sua conta está congelada. Por favor, entre em contato com um administrador ou com <bold>{{ email }}</bold> para reativá-la",
   "page-title-ghg-inventories": "Inventários de GEI",
   "page-title-hiap": "Ações & Planos",
+  "page-title-meed": "Ações e Planos v2",
```
lemilonkh: Might be good to use the same syntax here (`e` vs. `&`).

docs/MeedEndpoints.md
```diff
@@ -0,0 +1,104 @@
...
+
+### MEED-specific
+
+| Method | Path | Purpose | Guard |
+|---|---|---|---|
+| GET | `city/{cityId}/modules/meed/actions` | Action catalog (102 mitigation actions) | `findUserCity` |
+| GET | `city/{cityId}/modules/meed/city-attributes` | Socioeconomic context | `findUserCity` |
+| GET | `city/{cityId}/modules/meed/policy-scores` | Policy-alignment scores per action | `findUserCity` |
+| GET | `city/{cityId}/modules/meed/finance/feasibility` | Climate-finance feasibility | `findUserCity` |
+| GET | `city/{cityId}/modules/meed/finance/follow?link=` | Follows a Global API link | `findUserCity` |
```
lemilonkh: These should be adjusted to remove the `/modules/` route segment as those routes will be replaced with the ones from our codebase.

lemilonkh (commented):
Awesome, thanks 👍 ✨

app/migrations/20260812150000-register-meed-module.cjs
carlosgraffi: You are right and my original reasoning was wrong — removed in f73b18ef4.

I had checked `k8s/cc-migrate.yml`, seen only `db:migrate`, and concluded deploys never seed. I missed that `web-develop.yml` separately does `kubectl create -f k8s/cc-seed.yml`, which is a Job running `npm run db:seed`, and its `generateName: cc-seed-` means it runs on every deploy rather than clashing with the previous one. The module is in `seed-data/modules/modules.json` and the seeder upserts, so the migration was redundant.

app/seeders/20260804090000-rename-meed-module.cjs
carlosgraffi: Already gone — that seeder was removed in adb16ee68, along with the migration that replaced it (see the thread on `20260812150000-register-meed-module.cjs`). Your original point stands: seeders do run on every deploy.

app/src/app/[lng]/cities/[cityId]/MEED/meedRankingAdapter.ts
Comment (carlosgraffi) [reply app/src/app/[lng]/cities/[cityId]/MEED/meedRankingAdapter.ts:103]: Done in this PR (f73b18ef4).

The `{ en, es }` maps are now carried across whole rather than split into `_en`/`_es`, and `resolveLocalizedText` picks a language at the point of display — which is where it belongs, since one stored ranking is read by users in different languages. Your reasoning was the deciding one: the split cost a type change, an adapter change and a consumer change per new language, and it made the adapter do work the map already did.

app/src/app/[lng]/cities/[cityId]/MEED/meedRankingAdapter.ts
Comment (carlosgraffi) [reply app/src/app/[lng]/cities/[cityId]/MEED/meedRankingAdapter.ts:75]: Done in the same commit — the legal evidence fields now use the ranking route model spelling (`verdictCategory`, `ownershipDescription`, `legalJustification`, `legalReferences`), so they read the same as the rows they came from.

Note the two conventions are deliberate elsewhere: the ranking route is camelCase because it goes through CC models, while the six reference-data routes are pass-throughs and stay snake_case. `meedRankingAdapter.ts` is the one place that bridges the ranking side.

app/src/app/[lng]/cities/[cityId]/MEED/meedLocalState.ts
Comment (carlosgraffi) [reply app/src/app/[lng]/cities/[cityId]/MEED/meedLocalState.ts:7]: Ticket raised: #3070.

Agreed it should not last. Worth flagging the scope: it is not only preferences — confirmed exclusions and per-step progress are local too, and step progress drives the stepper, the readiness gate and the confidence score. Only the ranking got tables, so today a user on another browser keeps their ranking but loses every answer that produced it.

Comment (carlosgraffi) [reply app/src/app/[lng]/cities/[cityId]/MEED/meedLocalState.ts:7]: Captured in #3070 as part of the same work.

You are right that this is the harder half: these reads are synchronous today, so every consumer treats the value as simply present. The gate and the confidence score are the awkward ones — they compute during render from state that would become asynchronous.

app/src/app/[lng]/cities/[cityId]/MEED/meedMockRanking.ts
Comment (carlosgraffi) [reply app/src/app/[lng]/cities/[cityId]/MEED/meedMockRanking.ts:20]: Ticket raised: #3071.

Kept for now because it is still doing two jobs: it is the only way to demonstrate the module end to end with no hiap-meed running, and it is the no-regression baseline we check the API path against. The ticket lists what comes out with it — the flag, the `isMock` field and the "Sample data" tag.

app/src/app/[lng]/cities/[cityId]/MEED/useMeedRanking.ts
Comment (carlosgraffi) [reply app/src/app/[lng]/cities/[cityId]/MEED/useMeedRanking.ts:70]: Agreed, and noted as tech debt rather than fixed here.

For context on why it exists: the event is how a write in one screen tells the others to re-read local storage, which is not React state at all. Once persistence moves server-side (#3070) the listener should disappear on its own, since RTK Query cache invalidation would do the same job properly. Happy to do it sooner if you would rather not wait on that.

app/src/services/api.ts
carlosgraffi: Agreed these go — but as their own PR rather than this one, because the swap is not just a URL change.

One correction on "same return types as the previous implementations": they differ. The Global API returns camelCase (`actionId`, `actionName`, `timelineForImplementation`); the new routes are pass-throughs to hiap-meed, whose `reference_data` models have no alias generator, so they serialise snake_case (`action_id`, `action_name`, `implementation_timeline`). `buildActionIndex` only accepts the camelCase spelling and skips rows without an id, so pointed at the new payload it would return an empty index — silently, with every catalog screen rendering as if the city had no actions.

So it needs an adapter at the boundary, the way the ranking did, plus the 13 consumer files. It also makes hiap-meed a local dev dependency for those screens, which is worth landing on its own so it can be reverted cleanly if that bites.

This PR is already large; happy to start the migration PR next.

app/src/services/api.ts
carlosgraffi: Covered in the thread above — the transition is the next PR, and it needs a catalog adapter because the payload casing and field names differ from the Global API ones.

app/src/services/api.ts
carlosgraffi: Agreed — once the old accessors are gone the `Reference` prefix stops earning its place. I will drop it as part of the migration PR, in the same change that removes the ones it disambiguates from, so the rename and the deletion land together rather than leaving a window where both names mean the same thing.

app/src/app/api/v1/city/[city]/modules/meed/actions/route.ts
Comment (carlosgraffi) [reply app/src/app/api/v1/city/[city]/modules/meed/actions/route.ts:1]: Yes — removing this together with the consumer migration in the next PR, since deleting it before the screens move would take the catalog offline. Details in the `services/api.ts` thread: the payloads are not equivalent, so it needs an adapter rather than a repoint.

app/src/app/api/v1/city/[city]/modules/meed/city-attributes/route.ts
Comment (carlosgraffi) [reply app/src/app/api/v1/city/[city]/modules/meed/city-attributes/route.ts:1]: Yes — removing this together with the consumer migration in the next PR, since deleting it before the screens move would take the catalog offline. Details in the `services/api.ts` thread: the payloads are not equivalent, so it needs an adapter rather than a repoint.

app/src/app/api/v1/city/[city]/modules/meed/finance/feasibility/route.ts
Comment (carlosgraffi) [reply app/src/app/api/v1/city/[city]/modules/meed/finance/feasibility/route.ts:1]: Yes — removing this together with the consumer migration in the next PR, since deleting it before the screens move would take the catalog offline. Details in the `services/api.ts` thread: the payloads are not equivalent, so it needs an adapter rather than a repoint.

app/src/app/api/v1/city/[city]/modules/meed/policy-scores/route.ts
Comment (carlosgraffi) [reply app/src/app/api/v1/city/[city]/modules/meed/policy-scores/route.ts:1]: Yes — removing this together with the consumer migration in the next PR, since deleting it before the screens move would take the catalog offline. Details in the `services/api.ts` thread: the payloads are not equivalent, so it needs an adapter rather than a repoint.

app/src/app/api/v1/city/[city]/modules/meed/finance/follow/route.ts
Comment (carlosgraffi) [reply app/src/app/api/v1/city/[city]/modules/meed/finance/follow/route.ts:1]: Agreed, and it is worse than redundant — it is broken. Its guard only permits `/api/v1/cities/`, but the links it receives are `/api/v1/climate-finance/...`, so both calls 400 and the opportunity and project cards render "no data" without saying why.

It comes out with `useGetMeedFinanceLinkQuery`, `followLink` and the guard in the migration PR, with the consumers moving to `finance/projects` and `finance/opportunities`.

docs/MeedEndpoints.md
carlosgraffi: Updated in f73b18ef4 — the doc now leads with the canonical `city/{cityId}/meed/*` routes, including the two finance routes it was missing.

I kept a short section on the `modules/meed/*` proxies rather than deleting the mention, since they are still what the screens call until the migration lands, and it records why they cannot simply be repointed.

app/src/i18n/locales/es/dashboard.json
carlosgraffi: Fixed in f73b18ef4 — each now matches its neighbour in the same file: the ampersand in `navigation.json`, where `page-title-hiap` is translated but keeps `&`, and the untranslated product name in `dashboard.json`, where `breadcrumb-hiap` reads "Actions & Plans".

Flagging the alternative: harmonising the other way, on proper conjunctions, would read better in all three languages but means changing HIAP strings too. That felt like a product-wide call rather than ours — say the word and I will do it.

app/src/i18n/locales/es/navigation.json
carlosgraffi: Fixed in f73b18ef4 — each now matches its neighbour in the same file: the ampersand in `navigation.json`, where `page-title-hiap` is translated but keeps `&`, and the untranslated product name in `dashboard.json`, where `breadcrumb-hiap` reads "Actions & Plans".

Flagging the alternative: harmonising the other way, on proper conjunctions, would read better in all three languages but means changing HIAP strings too. That felt like a product-wide call rather than ours — say the word and I will do it.

app/src/i18n/locales/fr/dashboard.json
carlosgraffi: Fixed in f73b18ef4 — each now matches its neighbour in the same file: the ampersand in `navigation.json`, where `page-title-hiap` is translated but keeps `&`, and the untranslated product name in `dashboard.json`, where `breadcrumb-hiap` reads "Actions & Plans".

Flagging the alternative: harmonising the other way, on proper conjunctions, would read better in all three languages but means changing HIAP strings too. That felt like a product-wide call rather than ours — say the word and I will do it.

app/src/i18n/locales/fr/navigation.json
carlosgraffi: Fixed in f73b18ef4 — each now matches its neighbour in the same file: the ampersand in `navigation.json`, where `page-title-hiap` is translated but keeps `&`, and the untranslated product name in `dashboard.json`, where `breadcrumb-hiap` reads "Actions & Plans".

Flagging the alternative: harmonising the other way, on proper conjunctions, would read better in all three languages but means changing HIAP strings too. That felt like a product-wide call rather than ours — say the word and I will do it.

app/src/i18n/locales/pt/dashboard.json
carlosgraffi: Fixed in f73b18ef4 — each now matches its neighbour in the same file: the ampersand in `navigation.json`, where `page-title-hiap` is translated but keeps `&`, and the untranslated product name in `dashboard.json`, where `breadcrumb-hiap` reads "Actions & Plans".

Flagging the alternative: harmonising the other way, on proper conjunctions, would read better in all three languages but means changing HIAP strings too. That felt like a product-wide call rather than ours — say the word and I will do it.

app/src/i18n/locales/pt/navigation.json
carlosgraffi: Fixed in f73b18ef4 — each now matches its neighbour in the same file: the ampersand in `navigation.json`, where `page-title-hiap` is translated but keeps `&`, and the untranslated product name in `dashboard.json`, where `breadcrumb-hiap` reads "Actions & Plans".

Flagging the alternative: harmonising the other way, on proper conjunctions, would read better in all three languages but means changing HIAP strings too. That felt like a product-wide call rather than ours — say the word and I will do it.

app/src/backend/meed/MeedGlobalApiService.ts
carlosgraffi: Removing this with the consumer migration in the next PR rather than here — it is what the five `modules/meed/*` proxies call, so it goes when they do.

Agreed on the reasoning: routing the catalog through hiap-meed means the same filtering the ranking applies, which is the actual argument for the migration rather than just tidiness.

lemilonkh (approved):
Great work, thanks 👍 ✨

PR #2968, opened 2026-08-05
CC-557 Establish the CNB database

piotrnowakowski:
## Summary

- add an independent CNB Alembic chain and reversible initial schema for the planned workspace and funding-reference tables
- connect the reviewed-reference importer and similar-project reader to `CNB_DATABASE_URL`
- add Kubernetes Secret references, CNB migration Jobs, and deployment gates that run CA migrations before CNB migrations and application rollout
- document the two-database boundary and the externally managed credential/data rollout

## Database boundary

- `CA_DATABASE_URL` continues to own concept-note runs, context bundles, and uploads
- `CNB_DATABASE_URL` owns CNB workspace and funding-reference storage through `cnb_alembic_version`
- workspace `run_id` values remain external identifiers, with no cross-database foreign keys
- this change adds no HTTP endpoints, production seed data, RDS provisioning, or checked-in credentials

## Deployment and credentials

- deployments and migration Jobs consume only `CNB_DATABASE_URL` from environment-specific Kubernetes Secrets
- workflows fail fast when the corresponding GitHub Actions secret is missing, reconcile the Secret, and wait for both migration Jobs before rollout
- test intentionally reuses `CNB_DATABASE_URL_DEV` until a separate test database is supplied
- the credentials shared during planning must be rotated before the final URLs are stored as GitHub Secrets

## Validation

- focused CNB tests: 54 passed, 1 skipped
- Climate Advisor core suite: 284 passed, 10 skipped
- Ruff passed for the new and changed Python files
- Kubernetes/workflow YAML parsed successfully and the migration runner passed Bash syntax validation
- CNB Alembic offline SQL rendered successfully
- `git diff --check` passed

The skipped local migration test requires an available PostgreSQL instance; CI provisions PostgreSQL/pgvector and exercises upgrade, downgrade, re-upgrade, and CA/CNB chain isolation. The broader local suite's remaining failures were the pre-existing vector-insertion tests that require a populated PostgreSQL/pgvector corpus.

## Rollout notes

Apply and verify dev first, then test. Take an RDS snapshot or schema backup before production. Application images can be rolled back independently; the production schema should not be automatically downgraded after data is written.

No Linear changes are included.


amanda-eames (commented):
Wondering if it makes sense start with user/workspace tables that are clearly managed by CNB, this would be the chapter, revisions, gaps etc.

For the funders and funding_records, evidence this is data we have defined in the global-api and the current schemas don't fully align. It would be good not to duplicate effort, and whether the data team is managing the human review part.

I added some example funding data that can come from the global-api for you to review

amanda-eames (approved):
dB looks great

PR #2969, opened 2026-08-06
CC-594: Add HIAP-MEED reference-data API contracts

mircorudolph:
## Summary

Adds the HIAP-MEED backend proposal for stable MEED reference-data APIs, so public reads and existing processing workflows use the same canonical filtering and selection logic.

## Changes

- Define seven proposed HIAP-MEED reference-data contracts and backend-owned query rules
- Add complete request/response examples plus a product-facing API contract review
- Preserve existing prioritize, pre-flight, output-plan, and explanation API contracts


hiap-meed/frontend-data-endpoint-examples.md
```diff
@@ -0,0 +1,303 @@
...
+Each route has a direct Global API data-family mapping. HIAP-MEED owns technical query parameters, upstream validation, normalization, and post-fetch selection; responses are not raw passthroughs.
+
+These examples use the proposed **same rules and current data** guarantee. They do not include a snapshot identifier: a later prioritize or output-plan request applies the same canonical HIAP-MEED logic but may observe newer Global API data. An exact-snapshot guarantee would require a separately approved contract change.
+
+## 1. City attributes
+
+Maps Global API `GET /api/v0/city_attributes/{locode}`.
+
+```http
+GET /v1/city_attributes/CL%20IQQ
```
lemilonkh: It might make sense to use consistent folder structures, so perhaps this can be joined into the cities folder with
GET /v1/cities/CL%20IQQ/action-mitigation-feasibility-scores?country_code=CL

lemilonkh: The proposed action-pathways route is using a dash (-) instead of an underscore as the word separator. Let's pick one and use it consistently across the API if possible.

If we use the folder structure with the cities then that solves it as well of course.

lemilonkh (approved):
Great work 🙌 Let's align the API route paths, otherwise the returned data structures are good 👍🏻

hiap-meed/frontend-data-endpoint-examples.md
mircorudolph: Thanks, aligned now. City attributes use GET /v1/cities/{locode}/attributes, consistent with the other city-scoped routes. URL path segments use kebab-case where needed; JSON fields and query parameters remain snake_case

PR #2970, opened 2026-08-06
CC-651: Fix mangled external tool URLs in sidebar

cephaschapa:
## Summary

Fixes sidebar Tools links that prefixed external Open Earth URLs with the city path, producing mangled addresses like `/cities/{id}https://...`.

## Changes

- Resolve external module URLs as-is in the city home drawer (same rule as module cards)
- Open only external sidebar links in a new tab; keep internal tools in the same tab

## How to test

1. Open city home and expand the left drawer → Tools.
2. Click an external tool (e.g. boundary picker / Replit app).
3. Confirm it opens `https://...` in a new tab, not a CityCatalyst city URL with the external host appended.
4. Click an internal tool (e.g. GHGI) and confirm it navigates within the app on the same tab.

## Ticket

[CC-651](https://linear.app/openearth/issue/CC-651/sidebars-links-incorrect-urls)

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -2742,9 +2742,6 @@
       "cpu": [
         "arm"
       ],
-      "libc": [
-        "glibc"
-      ],
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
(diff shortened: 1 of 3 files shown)


lemilonkh (approved):
Thanks 👍🏻✨

PR #2975, opened 2026-08-07
CC-637: Register GHGI inventories and OCR artifacts

DavidGBbr:
## Summary

- Add a GHGI adapter for the shared `NativeInputCatalog` contract.
- Register uploaded source files separately from completed structured inventories.
- Register successful OCR Markdown with `PdfOcrJob.id`, digest, page count, and provenance metadata.
- Retry missed catalog writes from import status polling and the OCR cron.
- Withdraw GHGI catalog entries before inventory or city deletion.
- Document the producer mapping and add focused adapter tests.

## Validation

- 17 focused Jest tests pass across the GHGI adapter and PDF OCR queue/cron suites.
- ESLint passes for the changed backend/test files.
- Prettier check passes for the changed backend/test files.
- `git diff --check` passes.

## Notes

- Full TypeScript compilation still reports existing generated Next type, `react-email`, and `TTLCache` issues unrelated to this change.
- Full API inventory tests are currently limited by the existing missing `react-email` dependency.


app/src/backend/GHGINativeInputCatalogService.ts
```diff
@@ -0,0 +1,221 @@
...
+
+export async function syncPendingGHGIOcrArtifacts(limit: number) {
+  const jobs = await db.models.PdfOcrJob.findAll({
+    where: {
+      sourceType: "inventory_import",
+      status: "succeeded",
+      resultS3Key: { [Op.ne]: null },
+      resultSha256: { [Op.ne]: null },
+    },
+    order: [["completedAt", "ASC"]],
```
mircorudolph: The retry query always selects the oldest successful OCR jobs, including jobs that are already registered. Once limit older jobs exist, a newer artifact whose initial registration failed can never be reached; filter out existing non-withdrawn pdf_ocr_job catalog entries or use a progressing cursor.

The retry job checks only a small batch of the oldest successful OCR files—for example, the first two.
Even after those two are already registered successfully, it keeps checking the same two every time. It never moves on to later files.
So if file #10 failed to register temporarily, the retry system may never reach it. That file would remain missing from the catalog indefinitely.

app/src/app/api/v1/city/[city]/inventory/[inventory]/import/[importedFileId]/route.ts
```diff
@@ -271,6 +275,11 @@ export const GET = apiHandler(async (req: NextRequest, { session, params }) => {
     );
   }
 
+  await syncGHGIImportedInventorySource(importedFile);
```
Comment (mircorudolph) [inline app/src/app/api/v1/city/[city]/inventory/[inventory]/import/[importedFileId]/route.ts:278]: This retry omits the source digest. If upload-time registration failed, polling creates the catalog entry with contentDigest: null, and idempotent registration never updates it afterward; persist or recompute the digest and supply it on every retry.

A temporary failure can permanently remove the file’s integrity/provenance information from the catalog. The retry should save or recalculate the fingerprint and include it every time.

mircorudolph (approved):
Approved thanks!

But please have a look at the 2 comments I did as those will harden the retry logic.
Also might be good to add 2 regression tests for both of them.

1. verify that OCR catalog recovery skips already-registered jobs and reaches newer missing artifacts, 
2. and that retrying a failed source registration preserves its content digest.

app/src/backend/GHGINativeInputCatalogService.ts
DavidGBbr: Good catch, fixed! The retry now filters out OCR jobs that are already registered and paginates through the successful jobs before applying the batch limit, so newer missing artifacts can still be recovered. I also added a regression test for this case

app/src/app/api/v1/city/[city]/inventory/[inventory]/import/[importedFileId]/route.ts
Comment (DavidGBbr) [reply app/src/app/api/v1/city/[city]/inventory/[inventory]/import/[importedFileId]/route.ts:278]: The source SHA-256 digest is now persisted at upload time, recomputed for older files when missing, and supplied on every retry!
Idempotent registration also fills in the digest if an existing catalog entry is missing it.

Added regression coverage for a failed initial catalog write followed by a try

PR #2976, opened 2026-08-07
fix(CC-613): validate Create Account form on submit, not on every keystroke

tatimoreira:
## Summary

Fixes the Create Account (signup) form's validation UX: errors now fire only on submit instead of live while typing, render consistently through a single error slot with no icons, and a real bug where the password-mismatch error silently never appeared (because it was only checked after every other field already passed) is fixed. Also aligns the form's copy, field order, and heading typography with the Figma redesign.

## Changes

- Switch signup form validation to fire on submit instead of live `onChange`; the button stays enabled throughout so clicking it is what triggers validation
- Move the password-match check into a proper react-hook-form `validate` rule so it surfaces alongside every other field's errors in the same pass, instead of being skipped whenever another field (e.g. the terms checkbox) already failed
- Remove the duplicate "Minimum length should be 4" message and the warning icons on all field errors — everything now renders through the single standard error slot
- Force white "Create account" button text in both enabled and disabled states
- Update signup heading/subheading copy, field order (name → password → confirm password → email), and heading typography to match the CC-612 Figma redesign
- Update `e2e/signup.spec.ts` for the new submit-triggered validation flow
image
image

## How to test

1. Run `npm run dev` and open `/en/auth/signup`
2. Type into the Full name / Password / Confirm Password fields — confirm no error messages appear while typing
3. Leave fields empty/invalid (e.g. mismatched passwords, weak password, terms unchecked) and click **Create Account** — confirm all errors appear together in one pass, as plain text with no icons
4. Fix the errors and resubmit — confirm the account is created successfully
5. Trigger the disabled/loading state (submit with valid data) — confirm the button text stays white, not greyed out

## Ticket
[
CC-613](https://linear.app/openearth/issue/CC-613/fix-create-account-validation-timing-and-visual-feedback)


app/src/app/[lng]/auth/signup/page.tsx
```diff
@@ -203,24 +181,21 @@ export default function Signup(props: { params: Promise<{ lng: string }> }) {
...
-          mismatch={passwordsMismatch}
+          isSubmitted={isSubmitted}
+          validate={(value) => value === watchPassword || t("passwords-dont-match")}
+        />
+        <EmailInput
+          register={register}
+          error={errors.email}
+          t={t}
+          disabled={Boolean(prefilledEmail)}
+          defaultValue={prefilledEmail}
```
Comment (lemilonkh) [inline app/src/app/[lng]/auth/signup/page.tsx:200]: Why is the email input below the password inputs now?
Was that a requirement?

lemilonkh (approved):
Awesome, great to have this first experience that users have with our platform polished! ✨

app/src/app/[lng]/auth/signup/page.tsx
Comment (tatimoreira) [reply app/src/app/[lng]/auth/signup/page.tsx:200]: a per [design](https://www.figma.com/design/NjezWgcv4BpIrspfLmsX3l/CityCatalyst---Heuristic-analysis?node-id=477-19707&m=dev) provided @briansebaOEF do this needed to be change?

Comment (briansebaOEF) [reply app/src/app/[lng]/auth/signup/page.tsx:200]: Hi! I didn't realize there was a difference with the sign up in prod vs the design with the email input (I'm using some screens from previous designs). But it being a disabled input we could leave it at the bottom like in the design, so we don't have a disabled input in the middle of the editable inputs. Does that make sense?

PR #2974, opened 2026-08-07
[1/3] CC-606: Add Climate Advisor run listing

piotrnowakowski:
## Stack

This PR is layer 1 of the CC-606 native stack and now contains only the Climate Advisor run-list contract and its direct tests.

- #2974 — Climate Advisor run listing (this PR)
- #2977 — CityCatalyst wiring foundation
- #2978 — Concept Notes dashboard

The original combined-PR context is preserved below; its summary and test notes describe the complete stack.

---

## Summary

Adds the authorized Concept Note Builder run-list flow from Climate Advisor through CityCatalyst, plus the first Figma-derived dashboard screen for reviewing and resuming a city's runs.

## Changes

- add owner- and city-scoped Climate Advisor run listing with deterministic newest-first ordering and persisted lifecycle/progress fields
- add the authenticated CityCatalyst proxy, upstream contract validation, and RTK Query integration
- implement the Concept Notes dashboard with loading, error, empty, context, and run-card states
- include the existing Concept Note upload wiring harness from this branch
- add regression coverage and update Concept Note Builder runtime documentation

## How to test

1. Run `cd climate-advisor && ./.venv/Scripts/python.exe -m pytest service/tests/test_concept_note_runs.py -q`.
2. Run `cd app && npm run jest -- --runTestsByPath tests/concept-note-dashboard-utils.jest.ts tests/concept-note-runs.jest.ts --coverage=false`.
3. Run `cd app && npx tsc --noEmit`.
4. Open `/{lng}/cities/{cityId}/concept-notes` as a user authorized for that city.

## Ticket

CC-606

## Notes

- No database migration is required.
- Dashboard progress is derived only from persisted backend state; it does not infer percentages or counts.
- Targeted tests pass: 15 Climate Advisor tests and 14 CityCatalyst Jest tests. TypeScript compilation also passes.
- The repository's ESLint invocation is currently blocked before file linting by an ESLint 10 / `react/display-name` compatibility error: `contextOrFilename.getFilename is not a function`.

Recorded with sample data 

https://github.com/user-attachments/assets/9ec11779-3ed2-4c81-89e6-fbbaf7650229

PR #2977, opened 2026-08-09
[2/3] CC-606: Add concept note wiring foundation

piotrnowakowski:
## Summary

Adds the guarded CityCatalyst wiring flow for starting or resuming a Concept Note Builder run and uploading a PDF. This is layer 2 of 3 and depends on #2974.

## Changes

- add the Concept Note wiring page, controller, and upload UI
- add typed RTK Query operations for run detail, start, upload, status, and retry
- require both `CA_SERVICE_INTEGRATION` and `CONCEPT_NOTE_BUILDER` for page access
- add wiring and page-guard tests plus QA flag documentation

## How to test

1. Run `cd app && npm run jest -- --runTestsByPath tests/concept-note-wiring-utils.jest.ts tests/concept-note-page-guard.jest.ts --coverage=false`.
2. Run `cd app && npx tsc --noEmit`.
3. Enable both required flags and open `/{lng}/cities/{cityId}/concept-notes/wiring`.

## Ticket

CC-606

## Notes

- Stack order: #2974 → this PR → #2978.
- `CONCEPT_NOTE_BUILDER` remains disabled by default.
- Targeted ESLint is locally blocked inside `eslint-plugin-i18next` before reporting code findings.


app/src/backend/concept-note-page-guard.ts
```diff
@@ -0,0 +1,11 @@
+import { notFound } from "next/navigation";
+import { FeatureFlags, hasServerFeatureFlag } from "@/util/feature-flags";
+
+export function requireConceptNoteBuilderPageEnabled(): void {
+  if (
+    !hasServerFeatureFlag(FeatureFlags.CA_SERVICE_INTEGRATION) ||
+    !hasServerFeatureFlag(FeatureFlags.CONCEPT_NOTE_BUILDER)
+  ) {
+    notFound();
```
cephaschapa, line 9: We do have a custom 404 notFound page on CC not sure if we want to use it here as well.

app/src/components/ConceptNoteWiringHarness/concept-note-home-screen.tsx
```diff
@@ -0,0 +1,224 @@
...
+export function ConceptNoteHomeScreen({
+  controller,
+  lng,
+  reducedMotion,
+}: ConceptNoteHomeScreenProps) {
+  const { t } = useTranslation(lng, "concept-notes");
+  const statusLabel = t(uploadStatusTranslationKey(controller.uploadStatus));
+
+  return (
+    <VStack align="stretch" gap={8} maxW="1180px" mx="auto">
```
cephaschapa: We can also add a bg color here i.e `base.light` in cases where the user flips the theme to dark mode.

app/src/components/ConceptNoteWiringHarness/concept-note-scope-screen.tsx
```diff
@@ -0,0 +1,236 @@
...
+export function ConceptNoteScopeScreen({
+  controller,
+  lng,
+  reducedMotion,
+}: ConceptNoteScopeScreenProps) {
+  const { t } = useTranslation(lng, "concept-notes");
+  const statusLabel = t(uploadStatusTranslationKey(controller.uploadStatus));
+
+  return (
+    <VStack align="stretch" gap={6} maxW="1320px" mx="auto">
```
cephaschapa: Same here and all the other screens

cephaschapa (approved):
Great work thanks

app/src/backend/concept-note-page-guard.ts
piotrnowakowski: Done — I kept the guard’s `notFound()` call and added a route-level Concept Notes not-found boundary that renders the existing branded CityCatalyst 404 UI. I also extracted the shared UI so the existing catch-all route and Concept Notes use the same component. Covered by focused tests and the production build.

app/src/components/ConceptNoteWiringHarness/concept-note-home-screen.tsx
piotrnowakowski: Done - I added `bg="base.light"` to the home screen's root container so it retains the intended light surface when the application theme is switched to dark mode.

app/src/components/ConceptNoteWiringHarness/concept-note-scope-screen.tsx
piotrnowakowski: Done - I applied `bg="base.light"` to the scope screen as well and audited the complete wiring flow. Home and scope are the only top-level screens in this PR; the nested upload and workflow panels already define explicit semantic backgrounds.

PR #2981, opened 2026-08-10
CC-669: Add signed webhook subscriptions and delivery

cephaschapa:
## Summary

Org admins can subscribe HTTPS endpoints to CityCatalyst domain events. When an inventory is published, a plan is generated, or a datasource is connected, CC delivers an HMAC-signed JSON envelope asynchronously so partners (GCOM and others) do not need to poll.

## Changes

- Org-scoped webhook subscriptions: CRUD API, encrypted signing secrets, rotate-secret
- Admin **Manage Webhooks** tab (create/edit/delete, enable/disable, one-time secret modal)
- v1 emits: `inventory.published`, `plan.generated`, `datasource.connected`
- Postgres outbox + minute CronJob that POSTs signed deliveries with retry/backoff
- Event catalog, HMAC-SHA256 signature helpers, k8s manifests for dev/test/prod
- Architecture doc at `docs/WebhooksArchitecture.md`

## How to test

1. Run the app, open Admin → **Manage Webhooks**, pick an organization, and create a subscription with an `https://` URL and at least one v1 event.
2. Copy the signing secret from the one-time modal (it is not shown again).
3. Publish an inventory (or generate a plan / connect a datasource) and confirm a `WebhookDelivery` row is created.
4. In cluster, confirm the CronJob hits `POST /api/v1/cron/process-webhook-deliveries`. Locally, POST that route with `Authorization: Bearer $CC_CRON_JOB_API_KEY`.
5. Receiver should verify `X-CityCatalyst-Signature` as `HMAC-SHA256(secret, timestamp + "." + rawBody)`.

## Ticket

[CC-669](https://linear.app/openearth/issue/CC-669/webhooks-eventing-backbone-subscriptions-signed-delivery)

## Notes

- Partners still host their own HTTPS endpoint; this PR is the CityCatalyst side.
- Reserved catalog types (e.g. `inventory.unpublished`) are defined but not emitted yet.
- Outbound URLs are HTTPS-only; SSRF private-IP blocking is a follow-up.


app/src/app/[lng]/admin/CreateEditWebhookModal.tsx
```diff
@@ -0,0 +1,279 @@
...
+const schema = z.object({
+  name: z.string().trim().min(1).max(255),
+  url: z
+    .string()
+    .url()
+    .refine((value) => value.toLowerCase().startsWith("https://"), {
+      message: "webhook-url-https",
+    }),
+  events: z.array(z.enum(WEBHOOK_EMITTED_EVENT_TYPES)).min(1),
+});
```
Comment (lemilonkh) [inline app/src/app/[lng]/admin/CreateEditWebhookModal.tsx]: Are we enforcing the same validation in the API route?

app/src/backend/webhooks/WebhookDeliveryService.ts
```diff
@@ -0,0 +1,221 @@
...
+    consecutiveFailures,
+    enabled: shouldDisable ? false : params.subscription.enabled,
+    disabledAt: shouldDisable ? new Date() : params.subscription.disabledAt,
+  });
+
+  if (shouldDisable) {
+    logger.warn(
+      { subscriptionId: params.subscription.id, consecutiveFailures },
+      "Webhook subscription auto-disabled after consecutive failures",
+    );
```
lemilonkh, line 114: Can we add the webhook URL and the request response and HTTP code to the log output? Might make it easier to diagnose so we don't need to check the database and make a manual request when the webhook fails to figure out why it failed.

app/src/backend/webhooks/WebhookDeliveryService.ts
```diff
@@ -0,0 +1,221 @@
...
+
+  const due = await db.models.WebhookDelivery.findAll({
+    where: {
+      status: { [Op.in]: ["pending", "delivering"] },
+      [Op.or]: [{ runAfter: null }, { runAfter: { [Op.lte]: new Date() } }],
+    },
+    include: [{ model: db.models.WebhookSubscription, as: "subscription" }],
+    order: [["runAfter", "ASC"]],
+    limit: BATCH_SIZE,
+  });
```
lemilonkh, line 203: Would it make sense to delete the database entries after some time so this table doesn't fill up with lots of data over time?

app/src/backend/webhooks/WebhookService.ts
```diff
@@ -0,0 +1,287 @@
...
+  for (const event of events) {
+    if (!isWebhookEventType(event)) {
+      throw new createHttpError.BadRequest(`unknown-webhook-event:${event}`);
+    }
+  }
+}
+
+export default class WebhookService {
+  static toPublic(sub: WebhookSubscription): PublicWebhookSubscription {
+    return toPublic(sub);
```
lemilonkh: Why is this a separate local method? The implementation could be moved here inside the service I think and just called with this.toPublic() inside of the service if necessary?

app/src/models/init-models.ts
```diff
@@ -988,6 +1006,36 @@ export function initModels(sequelize: Sequelize) {
...
+    foreignKey: "organizationId",
+    onDelete: "CASCADE",
+    onUpdate: "CASCADE",
+  });
+  WebhookSubscription.belongsTo(User, {
+    as: "creator",
+    foreignKey: "createdBy",
+    onDelete: "SET NULL",
+    onUpdate: "CASCADE",
+  });
```
lemilonkh, line 1050: Do we need to add the other side of this relation? So user has many webhook subscription?

Probably less likely to be queried, mostly for completeness.

app/src/services/api.ts
```diff
@@ -2084,6 +2089,87 @@ export const api = createApi({
...
+        { success: boolean },
+        { organizationId: string; webhookId: string }
+      >({
+        query: ({ organizationId, webhookId }) => ({
+          url: `/organizations/${organizationId}/webhooks/${webhookId}`,
+          method: "DELETE",
+        }),
+        invalidatesTags: (_result, _err, { organizationId, webhookId }) => [
+          { type: "Webhook", id: webhookId },
+          { type: "Webhook", id: `ORG-${organizationId}` },
```
lemilonkh, line 2160: Why do we track these both by the ID of the webhook and the ID of the organization in the RTK cache?

app/src/util/validation.ts
```diff
@@ -406,3 +406,33 @@ export const nativeInputCatalogRegisterRequest = z.object({
...
+    message: "webhook-url-must-be-https",
+  });
+
+const webhookEventTypes = z.array(z.string().min(1)).min(1);
+
+export const createWebhookSubscriptionRequest = z.object({
+  name: z.string().trim().min(1).max(255),
+  url: httpsUrl,
+  events: webhookEventTypes,
+});
```
lemilonkh: I think we have the same or a very similar Zod validation object in the edit modal component in this PR? Can we import the validation there if it has to be done both in the frontend and backend, so both are ensured to be consistent?

app/tests/api/city-hiap-action_plan.jest.ts
```diff
@@ -577,70 +577,6 @@ describe("City HIAP Prioritization API", () => {
       await expectStatusCode(res, 401);
     });
 
-    it("deletes action plan successfully", async () => {
-      // Create ranking, ranked action, and action plan
```
lemilonkh, line 592: Why was this test case removed?

app/tests/api/webhooks.jest.ts
```diff
@@ -0,0 +1,231 @@
...
+  setupTests,
+  testUserID,
+} from "../helpers";
+import {
+  cleanupTestData,
+  createTestData,
+  type TestData,
+} from "../helpers/testDataCreationHelper";
+
+process.env.WEBHOOK_SECRET_ENCRYPTION_KEY = randomBytes(32).toString("base64");
```
lemilonkh: Can we set this in beforeAll and remove it in afterAll? So it doesn't affect the global env for all tests.

app/tests/webhooks-deployment-contract.jest.ts
```diff
@@ -0,0 +1,49 @@
+import { readFileSync } from "node:fs";
+import { resolve } from "node:path";
+
+const repoRoot = resolve(process.cwd(), "..");
+
+function readRepoFile(path: string): string {
+  return readFileSync(resolve(repoRoot, path), "utf8");
+}
+
+describe("webhook delivery deployment contract", () => {
```
lemilonkh, line 14: Are we sure that we need this test suite?
Seems a little weird to check the deployment YAML file in a Jest suite.

k8s/cc-process-webhook-deliveries.yml
```diff
@@ -0,0 +1,33 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: citycatalyst-process-webhook-deliveries
+  namespace: default
+spec:
+  schedule: "* * * * *"
```
lemilonkh: This means it runs every minute. This might be quite a lot of load if there are a bunch of webhooks to process? Should we configure it to run a bit less frequently?

Same for the cron jobs in the other environments.

lemilonkh (approved):
Amazing work, thank you! 🙌🎉✨

PR #2982, opened 2026-08-10
CC-603: Add HIAP-MEED reference-data APIs

mircorudolph:
## Summary

Adds seven HIAP-MEED reference-data GET APIs so frontend reads, prioritization, and output-plan generation use the same backend-owned queries and selection rules.

## Changes

- Add typed routes and public response models for city attributes, action pathways, policy and feasibility scores, finance opportunities, and comparable projects
- Centralize mitigation-only action selection and reuse existing finance opportunity/project logic across read and processing flows
- Expose the frontend-required Global API fields for actions, policy evidence, feasibility inputs, opportunities, and projects without relabelling source values
- Preserve normalized financial rows with missing scores represented as `null`, while keeping the neutral `0.5` fallback internal to prioritization
- Report malformed actions with missing `action_type`; preserve policy `document_type` and return `scope: null` when its scope is unknown
- Document caller-selectable inputs, backend-owned filtering, response contracts, and shared data flow
- Add integration and regression coverage for the new APIs and existing processing behavior

## Integration references

- [API routes and selectable parameters](https://github.com/Open-Earth-Foundation/CityCatalyst/blob/cc-603-implement-apis/hiap-meed/app/modules/reference_data/api.py)
- [Pydantic request/response models](https://github.com/Open-Earth-Foundation/CityCatalyst/blob/cc-603-implement-apis/hiap-meed/app/modules/reference_data/models.py)
- [Frontend request/response examples for all seven APIs](https://github.com/Open-Earth-Foundation/CityCatalyst/blob/cc-603-implement-apis/hiap-meed/frontend-data-endpoint-examples.md)
- [Compact product contract and backend-owned filtering](https://github.com/Open-Earth-Foundation/CityCatalyst/blob/cc-603-implement-apis/hiap-meed/reference-data-api-product-contract.md)

## Verification

- `261 passed` with `python -m pytest -q`
- `git diff --check`


hiap-meed/app/services/climate_finance_opportunities_api.py
```diff
@@ -145,7 +154,15 @@ def _screen_report_opportunities(
     route: str | None,
     limit: int,
 ) -> list[ClimateFinanceOpportunityRecord]:
-    """Select current candidates and recurring closed programmes to monitor."""
+    """
+    Select up to ``limit`` current and ``limit`` monitoring opportunities.
+
+    Open or otherwise active programmes are current. Closed programmes are kept
+    only when marked annual, periodic, recurring, or sporadic; cancelled and
+    expired programmes are excluded. A technical-assistance route narrows the
```
DavidGBbr, line 162: Normalize the instrument value before route filtering. The requested route is normalized, but opportunity.instrument is compared as an exact string. If the upstream API returns casing, whitespace, or space-separated variations such as Technical Assistance or technical assistance, the technical-assistance filter may not be applied and unrelated active opportunities could be returned. Could we normalize both values, or enforce the upstream enum in the response model, and add a regression test for this case?

hiap-meed/app/modules/reference_data/response_builders.py
```diff
@@ -0,0 +1,472 @@
...
+            endpoint="GET /v1/action-pathways",
+            total_records=len(actions),
+        ),
+        warnings=_warnings(result.warning),
+    )
+
+
+def _policy_scope(document_type: object) -> str:
+    """Classify a policy document as national, regional, or municipal."""
+    if document_type == "parcc":
```
DavidGBbr: Avoid silently classifying unknown document types as national. document_type is optional and the upstream models are additive-tolerant. At the moment, every value other than parcc and paccc, including missing or future values, is mapped to national. This could expose an incorrect policy scope to frontend consumers. Could we normalize the known values and add a regression test for missing, mixed-case, and unknown document types? If unknown values are expected to be possible, we may also want to avoid including them in scope-specific aggregates.

hiap-meed/app/services/action_pathways_api.py
```diff
@@ -55,6 +56,38 @@ def map_action_pathway_api_item_to_action(action: ActionPathwayApiItem) -> Actio
...
+    """
+    selected_actions: list[Action] = []
+    excluded_actions: list[Action] = []
+    missing_action_type_actions: list[Action] = []
+
+    # Keep malformed rows visible in diagnostics without treating them as eligible.
+    for action in actions:
+        normalized_action_type = (action.action_type or "").strip().lower()
+        if not normalized_action_type:
+            missing_action_type_actions.append(action)
```
DavidGBbr, line 81: Surface missing action types to API consumers. `select_prioritizable_actions()` silently drops entries when `action_type` is missing or unknown, but this endpoint still returns 200 without any signal that upstream data was excluded. If this is possible with the Global API contract, could we expose a warning/count (or otherwise make the data-quality issue observable) so consumers do not mistake a partial list for the complete set of pathways?

DavidGBbr (approved):

hiap-meed/app/services/climate_finance_opportunities_api.py
mircorudolph: Global API owns a defined and consistent instrument vocabulary, including technical_assistance. HIAP-MEED intentionally compares that canonical value directly and passes it through unchanged. We therefore won’t normalize alternative spellings here; an unexpected value should be treated as an upstream contract issue rather than silently accepted.

I agree however that it would be good to have an upstream enum list for those fields (however that would be relevant for many other fields as well and out of scope for this PR here)

hiap-meed/app/services/action_pathways_api.py
mircorudolph: The action-pathways response now includes meta.api_context.missing_action_type_count and adds a warning when upstream actions are excluded because action_type is missing. Valid non-mitigation actions remain intentionally excluded without warnings because the endpoint represents prioritizable mitigation actions.

hiap-meed/app/modules/reference_data/response_builders.py
mircorudolph: Policy evidence now preserves the original Global API document_type. We derive scope only for recognized document types using normalized matching. Missing or unknown types remain in the response with scope: null and do not contribute to regional or municipal aggregates. Regression coverage includes mixed-case, missing, and unknown document types.

lemilonkh (approved):
Amazing work, thanks 👍 ✨

PR #2984, opened 2026-08-11
CC-612: Simplify admin onboarding flow, fix invite/login handoff bugs

tatimoreira:
## Summary

Decouples inventory creation from first-time org-admin onboarding — the wizard now covers only project/city selection, population data, and inviting collaborators, ending at a "done" screen instead of creating a GHG inventory. Also fixes several bugs in the invite-acceptance and login/signup handoff that were blocking new org-admins from ever completing onboarding.

## Changes

- Onboarding wizard is now 3 steps (select city → population → invite collaborators); removed the "Inventory details" and "Third-party data" steps and the final `addInventory` call, and dropped the "Upload Existing Inventory" welcome-screen entry point.
- Fixed `POST /api/v1/city` falling back to a hardcoded, unrelated global default project when no `projectId` is supplied — it now finds or creates a default project scoped to the caller's own organization, which was causing brand-new org-admins to hit "You do not have access to this resource".
- Fixed the invite-accept flow allowing whoever is currently logged in (not necessarily the invited email) to silently claim a city invite; now requires the session email to match the invited email.
- Fixed unauthenticated visitors hitting a dead-end error screen when opening an invite link instead of being routed through login/signup and back.
- Fixed the onboarding welcome page's "Start inventory" button resolving to the wrong route (`/cities/setup` instead of `/cities/onboarding/setup`) and a typo'd onboarding image filename.
- Fixed `done/page.tsx` reading `cityId` from a URL path param that doesn't exist on that route (always `undefined`, breaking the "Check dashboard" link); now reads it from the query string.
- Fixed `callbackUrl` handling on login/signup so it isn't corrupted by an unused `from=signup` param or malformed values, both of which broke `signIn()` with a generic "unexpected error".

## How to test

1. As a site admin (or org-admin), send a new-user an org-admin invite to an organization with zero projects/cities.
2. Open the invite link while logged out — should route through login/signup and land back on the invite-accept page authenticated, instead of a dead-end error.
3. On the onboarding welcome screen, confirm only "Get started" is shown (no upload button), and the wizard has 3 steps.
4. Walk through: select/add a city, fill in population data, then either skip or invite a collaborator on the last step — should finish at `/cities/onboarding/done` with accurate copy, and "Check dashboard" should link to the real city, not `/cities/undefined`.

## Ticket

[CC-612](https://linear.app/openearth/issue/CC-612/onboarding-readjust-admin-user-flow-according-to-figma)

## Notes

- The population step's "auto-fill from historical data" previously keyed off the now-removed inventory year field; it degrades gracefully (no crash) but population fields now need manual entry.
- Found but did not fix: the org-admin invite-accept endpoint isn't idempotent — revisiting an already-accepted invite link creates a duplicate `OrganizationAdmin` row. Worth a follow-up.


app/src/app/api/v1/city/route.ts
```diff
@@ -145,8 +145,36 @@ export const POST = apiHandler(async (req, { session }) => {
...
+      const [defaultProject] = await Project.findOrCreate({
+        where: { organizationId: orgAdmin.organizationId },
+        defaults: {
+          projectId: randomUUID(),
+          name: "Default Project",
+          description: "Default project created automatically during onboarding",
+          cityCountLimit: 999999,
+          organizationId: orgAdmin.organizationId,
+        },
+      });
```
lemilonkh, line 165: A project is automatically created in the admin panel when creating an organization. Not sure this is necessary, but good to bullet proof the process I suppose.

lemilonkh (approved):
Amazing work, thanks 🎉

app/src/app/api/v1/city/route.ts
tatimoreira: Good flag — you're right that the admin panel already creates a project when an org is created there. This `findOrCreate` isn't redundant with that though, it's covering a path that skips the panel entirely:

Orgs created via the `/cities/onboarding` invite-acceptance flow (an org-admin accepting an invite) never go through the admin panel's org-creation code, so there's no guarantee a project already exists for that org by the time this endpoint runs. Before this fix, a city created with no `projectId` fell back to the hardcoded global `DEFAULT_PROJECT_ID`, which belongs to an unrelated org — the permission check then rejected brand-new org-admins with "no access to this resource."

`findOrCreate` closes that gap without assuming which code path created the org: it reuses an existing project if the org already has one, and only creates the "Default Project" fallback if not. So it's less "just in case" and more "this is the only thing keeping city creation working for invite-onboarded orgs."

PR #2985, opened 2026-08-11
fix: adds filter to prvent duplicate list of organizations

cephaschapa:
## Summary

Tightens spacing and sizing in the top navigation bar so language, organization, and user menus align more cleanly with the design system. Also deduplicates organizations in the org switcher.

## Changes

- Reduce fixed min-widths on language/org triggers (`minW="auto"`) and align padding/`gap` between menu controls
- Group language + org menus with a consistent `32px` gap; remove extra nesting and user-menu left margin
- Deduplicate orgs in the org dropdown by `organizationId` so the same org does not appear twice

## How to test

1. Open any authenticated page with the main nav.
2. Confirm language, organization (if multi-org), and user avatar menus look evenly spaced without oversized empty trigger width.
3. With an account that returns duplicate org entries, open the org menu and confirm each org appears once.
4. Switch language and organization; confirm behavior is unchanged.

## Ticket

[CC-654](https://linear.app/openearth/issue/CC-654/fix-navigation-bar-spacing-and-organization-selector-ui-issues)

### DIFF
--- a/app/src/components/navigation-bar.tsx
+++ b/app/src/components/navigation-bar.tsx
@@ -240,78 +240,75 @@ export function NavigationBar({
         {/* Menu Items */}
         <Box display="flex" gap="48px" alignItems="center">
           {children}
-          <Box display="flex">
-            <Box display="flex">
-              <MenuRoot
-                onOpenChange={(details) => {
-                  setLanguageMenuOpen(details.open);
-                }}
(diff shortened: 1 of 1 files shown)


lemilonkh (approved):
Awesome, thanks 👍🏻✨

PR #2986, opened 2026-08-11
CC-610: Fix empty State Name column on Projects page

tatimoreira:
## Summary

The Projects page "State Name" column was always empty because the backend never fetched or returned the `region` field for cities.

## Changes

- Add `region` to the City attribute allowlists in `UserService.findAllProjectForAdminAndOwner`, `findAllProjectForProjectAdmin`, and `findAllProjectForCollaborators`
- Include `region` in the two manually mapped city response objects in `findUserProjectsAndCitiesInOrganization`

## How to test

1. Open a project's detail page (`/organization/{id}/project/{project}`) for a project whose cities have a `region` value set.
2. Confirm the "State Name" column in the cities table now shows the region instead of being blank.

## Ticket

CC-610

### DIFF
--- a/app/src/backend/UserService.ts
+++ b/app/src/backend/UserService.ts
@@ -573,7 +573,7 @@ export default class UserService {
         {
           model: db.models.City,
           as: "cities",
-          attributes: ["cityId", "name", "countryLocode", "country"],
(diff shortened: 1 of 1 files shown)


lemilonkh (approved):
Great fix, thanks 👍 ✨

PR #2987, opened 2026-08-11
Register completed legacy HIAP artifacts in NativeInputCatalog

DavidGBbr:
## Summary

- Add the legacy HIAP NativeInputCatalog adapter for completed rankings, committed ranked/unranked selections, and action-plan versions.
- Wire registration and withdrawal to ranking completion, selection persistence, action-plan persistence/deletion, and inventory/city deletion boundaries.
- Preserve source-of-truth HIAP tables and keep catalog entries pointer-based without copying artifact content.
- Add focused Jest coverage and document the legacy HIAP mapping, explicitly excluding HIAP-MEED.

## Validation

- Manual local validation completed against the CityCatalyst PostgreSQL container.
- Confirmed one active `hiap_ranking` catalog entry for the Krakow inventory.
- Confirmed the source ranking is `SUCCESS` with 20 persisted ranked actions.
- `git diff --check` passed.
- The current local Jest rerun is blocked before test execution by an inconsistent dependency tree in `node_modules` (Jest 30.4.2 with extraneous/invalid Jest packages); no CC-638 assertion failure was produced.

Linear: CC-638


DavidGBbr:
### Local validation completed

Validated the CC-638 flow locally against the `citycatalyst-db` PostgreSQL container:

- `NativeInputCatalog` contains 1 active `hiap_ranking` entry.
- The entry is linked to the Krakow inventory and ranking `cdded9eb-e554-4fc5-af77-98d1c142d250`.
- The authoritative `HighImpactActionRanking` has status `SUCCESS` and 20 persisted ranked actions.
- This confirms the ranking completion boundary registered the artifact in the catalog.

`git diff --check` also passed. The current local Jest rerun is blocked before test execution by an inconsistent `node_modules` dependency tree (`jest@30.4.2` with extraneous/invalid Jest packages), rather than by a CC-638 assertion failure.

app/src/backend/hiap/HiapService.ts
```diff
@@ -463,6 +467,7 @@ async function processBulkJobResults(
 
       // Update ranking status to success
       await ranking.update({ status: HighImpactActionRankingStatus.SUCCESS });
+      await syncHIAPRanking(ranking);
```
mircorudolph, line 470: Both completion paths mark the ranking SUCCESS before calling syncHIAPRanking, while that function catches catalog failures and returns null. Because the cron only revisits pending rankings, any transient registration failure permanently leaves a completed ranking uncatalogued; add a durable retry/backfill path or make completion and registration atomic.

The fix is likely to either:
retry failed catalog registrations later, or
only mark the ranking successful when both saving the ranking and catalog registration succeed together.

app/src/app/api/v1/city/[city]/hiap/action-plan/route.ts
```diff
@@ -156,6 +156,7 @@ export const POST = apiHandler(
       actionId: validatedData.actionId,
       highImpactActionRankedId: validatedData.hiActionRankingId,
       cityLocode: validatedData.cityLocode,
+      inventoryId: validatedData.inventoryId,
```
Comment (mircorudolph) [inline app/src/app/api/v1/city/[city]/hiap/action-plan/route.ts:170]: Could we clarify whether this direct-save endpoint is still intentionally supported?
This is not the distinction between “generate rankings” and “generate an action plan.” Both routes concern action plans:
[POST /city/{city}/hiap/action-plan/generate/{rankingId}](https://github.com/Open-Earth-Foundation/CityCatalyst/blob/257c85237ab302ec663725e236aed9ec76f6ffa4/app/src/app/api/v1/city/%5Bcity%5D/hiap/action-plan/generate/%5BrankingId%5D/route.ts#L72-L88) is used by the frontend. It checks inventory access, calls the HIAP plan-creator service, and then saves the generated plan in CC.
[POST /city/{city}/hiap/action-plan](https://github.com/Open-Earth-Foundation/CityCatalyst/blob/257c85237ab302ec663725e236aed9ec76f6ffa4/app/src/app/api/v1/city/%5Bcity%5D/hiap/action-plan/route.ts#L148-L167) only accepts an already-generated plan and saves it directly.
The direct-save route does not require a signed-in user, call PermissionService, or verify that the body-supplied inventoryId belongs to the city in the URL. This PR now also uses that value to determine the plan’s inventory, city, project, and organization catalog scope.
Running the services in EKS does not by itself protect this route: it is part of the regular CC web API, so a client can call it directly if the CC application is publicly reachable. I also could not find a current frontend caller for this direct-save POST; the frontend uses the authenticated generation endpoint instead.
If the direct-save endpoint is obsolete, could we remove it? If it is still required, please apply PermissionService.canAccessInventory(...) and verify that the inventory, route city, ranking, and action plan belong together before saving or registering the catalog entry. This concern is separate from HIAP ranking generation.

Comment (mircorudolph) [reply app/src/app/api/v1/city/[city]/hiap/action-plan/route.ts:170]: It seems there are 2 routes, 1 using auth and the other does not (which was not part of your changes in this PR, but worthwhile checking it out)

mircorudolph (changes requested):
Tested the feature and works. Thanks!

But please have a look at the comment specifically in 
`app/src/app/api/v1/city/[city]/hiap/action-plan/route.ts`

app/src/backend/hiap/HiapService.ts
DavidGBbr: Implemented a durable backfill path for this case.

The HIAP cron now checks successful rankings with persisted ranked actions and retries the NativeInputCatalog registration when the active `hiap_ranking` entry is missing. If registration fails, the ranking remains `SUCCESS`, and the next cron execution retries it again.

I also added regression coverage for the failed-registration-then-retry scenario and included `catalogBackfilled` in the cron metrics.

app/src/app/api/v1/city/[city]/hiap/action-plan/route.ts
Comment (DavidGBbr) [reply app/src/app/api/v1/city/[city]/hiap/action-plan/route.ts:170]: Confirmed that the direct-save endpoint is intentionally kept for legacy and backward-compatible flows.

It now requires an authenticated session and calls `PermissionService.canAccessInventory(...)`. The HIAP relationship checks were moved into `ActionPlanService`, which validates the inventory/city relationship, city locode, ranked action/action relationship, ranking inventory, and `SUCCESS` status before upserting the plan or registering its catalog entry.

I also added regression tests for unauthenticated requests and invalid relationships. The route remains available, but the client can no longer use a body-supplied inventory or unrelated ranking to write an action plan.

app/src/backend/hiap/ActionPlanService.ts
```diff
@@ -307,7 +375,9 @@ export default class ActionPlanService {
         return null;
       }
 
-      return await this.getActionPlanById(input.id);
+      const actionPlan = await this.getActionPlanById(input.id);
+      if (actionPlan) await syncHIAPActionPlan(actionPlan);
```
mircorudolph, line 416: This is the same authorization issue that was fixed for the direct-save POST, but it remains on the PATCH and DELETE routes. A caller could still modify or delete another user’s action plan—and update or withdraw its catalog entry—so please require authentication, load the plan by ID, check access using its stored inventoryId, and verify its cityId matches the {city} URL before changing it.

app/src/backend/hiap/HiapNativeInputCatalogService.ts
```diff
@@ -0,0 +1,596 @@
...
+}) {
+  return tryCatalogSync(() => registerHIAPSelections(input), {
+    inventoryId: input.inventoryId,
+    actionType: input.actionType,
+    catalogKind: "hiap_selection",
+  });
+}
+
+export function syncHIAPActionPlan(plan: ActionPlan) {
+  return tryCatalogSync(() => registerHIAPActionPlan(plan), {
```
mircorudolph, line 631: This is the same catalog-retry issue that was fixed for completed rankings, but action plans and selections are still not covered. Their synchronization also catches failures and returns null, so a successful plan or selection write can remain missing from the catalog. Please extend the durable backfill/retry mechanism to these artifact types as well.

mircorudolph (approved):
Thanks for fixing those issues.

I think we missed similar issues for the PATCH/DELETE route and retry mechnism for the plan generation

app/src/backend/hiap/ActionPlanService.ts
DavidGBbr: Fixed!

app/src/backend/hiap/HiapNativeInputCatalogService.ts
DavidGBbr: Fixed!

PR #3002, opened 2026-08-12
fix(CC-615): simplify 404 page copy and update CTA text

cephaschapa:
## Summary

Simplifies the 404 page by removing the descriptive paragraph and rewording the primary action to point users back to tools more clearly.

## Changes

- Remove the `not-found-description` block from the 404 page UI
- Rename the CTA translation key to `back-to-tools`
- Update the 404 button text across `en`, `de`, `es`, `fr`, and `pt` locales

## How to test

1. Open a non-existent route and confirm the 404 page no longer shows the description text
2. Verify the CTA button still renders correctly and uses the updated “back to tools” wording
3. Spot-check the localized 404 page in the supported languages

### DIFF
--- a/app/src/i18n/locales/de/not-found.json
+++ b/app/src/i18n/locales/de/not-found.json
@@ -1,6 +1,5 @@
 {
-  "not-found-description": "Scheint, als wären wir in den Emissionsnebel geraten. Bitte kehren Sie zum Dashboard zurück",
-  "goto-dashboard": "Gehe zum Dashboard",
+  "back-to-tools": "Zuruck zu den Tools",
   "invite-not-valid": "Diese Einladung ist nicht mehr gültig",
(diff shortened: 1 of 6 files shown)


app/src/i18n/locales/de/not-found.json
```diff
@@ -1,6 +1,5 @@
 {
-  "not-found-description": "Scheint, als wären wir in den Emissionsnebel geraten. Bitte kehren Sie zum Dashboard zurück",
-  "goto-dashboard": "Gehe zum Dashboard",
+  "back-to-tools": "Zuruck zu den Tools",
```
lemilonkh:
```suggestion
  "back-to-tools": "Zurück zu den Tools",
```

lemilonkh (approved):
Good to make this more intuitive, thanks 👍🏻

PR #3005, opened 2026-08-12
chore(meed): set HIAP_MEED_API_URL on the web deployments

carlosgraffi:
Sets `HIAP_MEED_API_URL` on the web deployments, so the Actions & Plans v2 module has an address to reach the prioritizer once its client lands.

Four lines: one per deploy workflow, plus `env.example` for local setup.

| Workflow | Value |
|---|---|
| `web-develop.yml` | `http://hiap-meed-service-dev` |
| `web-test.yml` | `http://hiap-meed-service-test` |
| `web-tag.yml` | `http://hiap-meed-service-prod` |
| `app/env.example` | `http://localhost:8000` (run hiap-meed via its own `docker compose`) |

Service names confirmed by @Mirco and matching `hiap-meed/k8s/service-{dev,test,prod}.yml` — all `ClusterIP` on port 80 → container 8000, no ingress, so they resolve only from inside the cluster.

### Why a runtime variable and not `.env`

Worth stating, because it isn't obvious from reading the repo. The build writes a `.env` (workflow line ~88), but the runtime image copies only `.next`, `public` and `templates` — **`.env` never reaches the running container.** A value set only there would be silently `undefined` in dev, and the client would quietly fall back to its default rather than failing. So this follows `HIAP_API_URL`, `CA_BASE_URL` and `OPENCLIMATE_API_URL`: set on the deployment at deploy time.

`env.example` still gains the entry — that's documentation for local setup, not deployment.

### Safe to merge now

**No code reads this variable yet.** The module's hiap-meed client doesn't exist, so this is inert until it lands. Merging now means dev picks it up on the next redeploy instead of needing a second workflow change later. Nothing else is affected.

### One observation, deliberately not changed here

`HIAP_API_URL` is set to `http://hiap-service-dev` in **all three** workflows — including `web-test.yml` and `web-tag.yml`. That looks unintended, but it's the existing HIAP service and not this module's to change; repointing prod's HIAP endpoint as a side effect of a MEED chore would be the wrong way to fix it. Flagging it in case it's news. The new `HIAP_MEED_API_URL` lines use the correct per-environment service in each workflow.

### Not included

`MEED_MODULE` in `NEXT_PUBLIC_FEATURE_FLAGS` is left out on purpose, so the flag question doesn't block this. For the record: the module *is* gated today — `HomePage.tsx:293` filters it out of the tool accordion unless the flag is set, so without it the module is invisible on dev (reachable only by direct URL, and then only with the module granted to the project). Setting the flag or removing the gate are both reasonable; it needs a decision rather than a default. @Milan.


lemilonkh (approved):
Great, thanks 👍 ✨

PR #3006, opened 2026-08-12
CC-513 Assemble PDF-first context bundles

piotrnowakowski:
## Summary

- assemble a Concept Note context bundle automatically after at least one city PDF reaches `ready`
- keep GHGI and persisted HIAP optional, retaining partial/usable data and storing explicit typed empties otherwise
- analyze every ready PDF with page-preserving partitions, exact excerpt verification, bounded reader concurrency, and guarded build commits
- expose a run-scoped, read-only selected-document query capability plus a scoped retry route and CityCatalyst proxy
- document the PDF-first contract, model/prompt configuration, failure handling, and capability scope

## Full-app validation

A real local full-stack run exercised CityCatalyst browser login, Concept Note PDF upload, PDF OCR, Climate Advisor background bundle assembly, and the real CityCatalyst GHGI and HIAP capability routes. The persisted result reached `ready` with GHGI `partial`, HIAP `available`, no warnings, and completion event `concept_note_context_bundle_ready`.
## User and developer impact

- a successfully ingested city PDF is now the only mandatory context source
- missing or unavailable GHGI/HIAP data cannot block Concept Note readiness
- stale background builds cannot overwrite a bundle after the ready-upload set changes

## Validation

- GitHub checks on head `0ba24144d`: CodeQL, CA auth contract, CA tests, and CityCatalyst `runTests` all passed

- real full-app browser path: login -> create run -> upload PDF -> OCR -> automatic bundle build -> persisted ready bundle
- CodeQL findings fixed with exact parsed-host matching and non-reflective retry errors; regression tests added
- focused Climate Advisor review suite - 59 passed
- focused CityCatalyst Concept Note upload/OCR/pointer suite - 54 passed
- `git diff --check`

Linear: https://linear.app/openearth/issue/CC-513/combine-all-datasources-into-a-context-bundle


climate-advisor/docs/examples/cc-513-full-context-bundle.json
```diff
@@ -0,0 +1,255 @@
...
+      },
+      "availability": "available",
+      "inventory_id": "58830000-0000-4000-8000-000000000004",
+      "requested_language": "en"
+    },
+    "project": null
+  },
+  "funder_context": null,
+  "document_context": null,
+  "selected_sources": [
```
piotrnowakowski: empty PDF so that we can just see the builded bundle

climate-advisor/service/app/services/cnb/context_bundle.py
```diff
@@ -0,0 +1,381 @@
...
+    service: ContextBundleService,
+    user_id: str,
+    run_id: UUID,
+    token: str,
+    force: bool = False,
+    snapshot: ContextBundleBuildSnapshot | None = None,
+) -> None:
+    """Retain an in-process background build until terminal completion."""
+
+    task = asyncio.create_task(
```
DavidGBbr, line 356: This uses an in-process asyncio.create_task after the upload is marked as ready. If the Climate Advisor pod restarts, is redeployed, or is rescheduled while the task is running, the task is lost and the run can remain stuck in building indefinitely.

Could we enqueue this work durably, or add a startup/periodic reconciler for builds left in building before relying on automatic bundle assembly?

climate-advisor/service/app/persistence/concept_notes/runs.py
```diff
@@ -137,3 +143,24 @@ async def list_for_user_city(
...
+            select(Thread).where(
+                Thread.thread_id == thread_id,
+                Thread.user_id == user_id,
+            )
+        )
+        thread = result.scalar_one_or_none()
+        if thread is None:
+            return
+        context = dict(thread.context or {})
+        context["concept_note_run_id"] = str(run_id)
```
DavidGBbr: This adds concept_note_run_id without clearing a possible stationary_energy_draft_run_id already stored on the thread. In that case, the streaming layer prioritizes Stationary Energy context while AgentService can register both Stationary Energy and Concept Note tools.

Could we clear the previous workflow identifier when binding a new one, or enforce a single scoped workflow per thread?

climate-advisor/service/app/tools/concept_note_source_tools.py
```diff
@@ -0,0 +1,157 @@
...
+                    upload_id=str(upload_uuid),
+                    token=token,
+                )
+                pages = verify_source_artifact_fn(
+                    artifact=artifact,
+                    markdown_s3_key=upload.markdown_s3_key,
+                    sha256=upload.markdown_sha256,
+                    page_count=upload.page_count,
+                )
+                result = await query_document_fn(
```
DavidGBbr, line 98: Each invocation of this capability re-fetches and analyzes every page of the selected document through multiple LLM calls. There is no per-run cache or query/token budget, so repeated questions in the same chat can multiply latency and provider cost.

Could we cache results by (run_id, upload_id, question) and enforce a per-run or per-turn query/token limit?

DavidGBbr (commented):
The overall PDF-first context-bundle design is coherent, and the automated checks are green. I left three inline comments about background-build recovery, mutually exclusive workflow context, and repeated full-document source-query cost.

DavidGBbr (approved):
Great work! The PDF-first context-bundle flow is thoughtfully designed, and the validation, persistence safeguards, optional GHGI/HIAP handling, and test coverage are all very solid.

I’m approving the PR. A few follow-up improvements would be worth addressing:

- Consider making background bundle builds durable or adding recovery for builds interrupted by pod restarts.
- Ensure Concept Note and Stationary Energy workflow contexts cannot coexist on the same thread.
- Consider caching source queries and adding per-run or per-turn limits to control repeated full-document LLM calls.

Thanks for the thorough implementation and documentation!

climate-advisor/service/app/services/cnb/source_analysis.py
```diff
@@ -0,0 +1,693 @@
...
+    for result in results:
+        if isinstance(result, BaseException):
+            raise result
+    return cast(list[OutputModel], results)
+
+
+def resolve_model_name(model_name: str, client: AsyncOpenAI) -> str:
+    """Strip the OpenRouter provider prefix only for the OpenAI endpoint."""
+    base_url = str(client.base_url or "")
+    if "api.openai.com" in base_url and model_name.startswith("openai/"):
```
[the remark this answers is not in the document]
piotrnowakowski: Resolved. The original substring check was first replaced with exact parsed-host comparison and was subsequently removed from CNB source analysis because this path is OpenRouter-only. The current CodeQL checks pass.

climate-advisor/service/app/routes/concept_note_context_bundle.py
```diff
@@ -0,0 +1,92 @@
...
+        yield client
+    finally:
+        await client.close()
+
+
+def problem(status_code: int, code: str, message: str) -> JSONResponse:
+    """Return a stable machine-readable bundle error."""
+    return JSONResponse(
+        status_code=status_code,
+        content={"code": code, "detail": message, "status": status_code},
```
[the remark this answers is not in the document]
piotrnowakowski: Resolved. Persistence exceptions now map to fixed public messages by error code, with a stable generic fallback. Internal exception and stack-trace details are no longer returned to the caller. The current CodeQL checks pass.

climate-advisor/service/app/services/cnb/context_bundle.py
piotrnowakowski: Implemented in `434840a12` using the smaller reconciler approach. Climate Advisor now runs a five-minute reconciler that marks bundles left in `building` for more than one hour as retryable `context_bundle_build_interrupted` failures. Recovery uses database row locking with `SKIP LOCKED`, and guarded failure updates cannot overwrite a bundle that has already become ready. Tests cover stale selection, retryable state, and periodic execution.

climate-advisor/service/app/persistence/concept_notes/runs.py
piotrnowakowski: Implemented in `434840a12`. Workflow binding now removes both known workflow identifiers before setting exactly one active identifier. The rule is applied in both the Concept Note and Stationary Energy persistence paths while preserving access tokens and unrelated `threads.context` values. This reuses the existing JSONB column and includes regression coverage for both binding directions.

climate-advisor/service/app/tools/concept_note_source_tools.py
piotrnowakowski: Partially addressed in `434840a12`. For this PR, I applied the smaller operational guard by reducing configured and process-wide source-reader concurrency from 4 to 3. This limits simultaneous provider requests, but it does not cache repeated questions or cap total per-run token usage. Exact-result caching and per-turn/per-run budgets are deferred so their persistence, invalidation, and quota semantics can be designed explicitly.

PR #3007, opened 2026-08-13
fix: CC-627 inconsistent minimum password length and validation

lemilonkh:
- Set password minimum length to 8 in all places
- Make sure password is valid before enabling submit button on manage password tab in settings


cephaschapa (approved):
Great thanks 👍✨

PR #3008, opened 2026-08-13
fix: CC-704 don't require ranking ID for action plan generation

lemilonkh:
Fixes not being able to generate an action plan for unranked actions in HIAP.

Changing the path of the actiin plan generation route was required as unranked actions don't have a ranking ID, which made the request fail because of a missing URL path parameter.

### DIFF
--- a/app/src/app/api/v1/city/[city]/hiap/action-plan/generate/route.ts
+++ b/app/src/app/api/v1/city/[city]/hiap/action-plan/generate/route.ts
@@ -15,14 +15,14 @@ const generateRankingRequest = z.object({
 
 /**
  * @swagger
- * /api/v1/city/{city}/hiap/action-plan/generate/{rankingId}:
+ * /api/v1/city/{city}/hiap/action-plan/generate:
  *   post:
  *     tags:
  *       - city
  *       - hiap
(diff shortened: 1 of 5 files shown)


app/src/app/api/v1/city/[city]/hiap/action-plan/generate/route.ts
```diff
@@ -15,27 +15,21 @@ const generateRankingRequest = z.object({
...
  *         required: true
  *         schema:
  *           type: string
  *         description: City ID
- *       - in: path
- *         name: rankingId
- *         required: true
- *         schema:
- *           type: string
- *         description: Ranking ID
```
Comment (cephaschapa) [inline app/src/app/api/v1/city/[city]/hiap/action-plan/generate/route.ts:38]: Should we also update the ActionPlan model due this change?

cephaschapa (approved):
Great thanks 👍

app/src/app/api/v1/city/[city]/hiap/action-plan/generate/route.ts
Comment (lemilonkh) [reply app/src/app/api/v1/city/[city]/hiap/action-plan/generate/route.ts:38]: The action plan stays the same, this is just the parameters that the HIAP service API requires in order to generate it.

PR #3009, opened 2026-08-13
Add native input catalog reconciliation sandbox

DavidGBbr:
## Summary

- Add bounded reconciliation for the current GHGI and legacy HIAP catalog adapters.
- Add deterministic keyset pagination with aggregate reports and incomplete-scan safeguards.
- Add an authenticated internal reconciliation endpoint with dry-run/apply modes.
- Add a public GitHub Actions sandbox that runs dry-run first, then disposable apply/rollback validation.

## Safety

- Reconciliation never modifies or deletes producer source records.
- Apply repairs only deterministic missing registrations through the existing registration lifecycle.
- The sandbox uses disposable Postgres fixtures only.
- The sandbox has no secrets, no production endpoint, and no production database access.
- The isolated apply test uses a catalog-only writer role and rolls back the created fixture row.

## Validation

- 7 focused reconciliation tests passed.
- Changed-file lint passed.
- Prettier checks passed.
- Docker Compose validation passed.
- Local dry-run and apply/rollback JSON artifacts were generated successfully.
- Full TypeScript validation remains blocked only by pre-existing missing `@zag-js/popover` and `@zag-js/checkbox` dependencies.

Production apply/backfill is not included in this PR.


app/src/app/api/v1/internal/native-input-catalog/reconcile/route.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * @swagger
+ * /api/v1/internal/native-input-catalog/reconcile:
+ *   post:
```
mircorudolph, line 4: Add a responses block to this Swagger operation. The required runTests check currently fails OpenAPI validation here, causing all Jest and Playwright steps to be skipped.

See: https://github.com/Open-Earth-Foundation/CityCatalyst/actions/runs/31740988998/job/94584026877?pr=3009

app/src/backend/NativeInputCatalogReconciliationService.ts
```diff
@@ -0,0 +1,782 @@
...
+async function collectCatalogPage(
+  pageSize: number,
+  states: Record<StreamName, StreamState>,
+): Promise<PageCollection> {
+  const stream = await readPage(
+    "catalog",
+    db.models.NativeInputCatalog,
+    {
+      owningModule: { [Op.in]: PRODUCERS },
+      availability: "active",
```
mircorudolph: An inventory_source_file catalog entry is created after the uploaded file is stored, before OCR or inventory import succeeds. If PDF processing later fails with a non-retryable source error, PdfOcrService intentionally keeps the ImportedInventoryFile row with importStatus = failed but withdraws its catalog entry; this active-only scan then treats it as missing, and apply mode creates a new active entry, undoing that withdrawal. Please account for withdrawn identities or exclude failed/ineligible uploads before repairing missing registrations.

app/src/app/api/v1/internal/native-input-catalog/reconcile/route.ts
DavidGBbr: Added the missing responses block to the Swagger definition for the reconciliation endpoint. Thanks for highlighting it to me!

app/src/backend/NativeInputCatalogReconciliationService.ts
DavidGBbr: Fixed this by making the reconciliation scan both active and withdrawn catalog entries. 
Withdrawn identities are now kept separate and classified as skipped, so 'apply' will not recreate them. 
Failed GHGI imports are also excluded from repair and reported as skipped.
I also added regression tests covering both withdrawn registrations and failed imports.

mircorudolph (approved):

PR #3011, opened 2026-08-14
CC-612: Fix GHGI module not enabled by default after onboarding

tatimoreira:
## Summary

QA feedback on CC-612: after completing onboarding as an org-admin, the GHG Inventory module wasn't shown on the dashboard by default. New default projects created during onboarding had no `ProjectModules` rows, so GHGI access was never enabled.

## Changes

- Enable GHGI module access on the org-admin's default project in `POST /api/v1/city` whenever it's missing, right after the project is resolved/created.
- Check-then-enable rather than only on creation, so orgs whose default project was already created during the regression window get self-healed on next city creation too.

## How to test

1. Sign up / log in as a new org-admin and complete onboarding (create a city, no need to create an inventory).
2. Land on the dashboard (`/cities/[cityId]`) and confirm the GHG Inventory module card is visible instead of "No modules" / empty state.
3. For an org that already has a default project with zero modules (pre-existing regression case), create another city and confirm GHGI access gets backfilled.

## Ticket

https://linear.app/openearth/issue/CC-612/onboarding-readjust-admin-user-flow-according-to-figma

### DIFF
--- a/app/src/app/api/v1/city/route.ts
+++ b/app/src/app/api/v1/city/route.ts
@@ -130,10 +130,11 @@ import createHttpError from "http-errors";
 import { NextResponse } from "next/server";
 import { randomUUID } from "node:crypto";
 import { logger } from "@/services/logger";
-import { DEFAULT_PROJECT_ID } from "@/util/constants";
+import { DEFAULT_PROJECT_ID, Modules } from "@/util/constants";
 import EmailService from "@/backend/EmailService";
 import UserService from "@/backend/UserService";
 import { PermissionService } from "@/backend/permissions/PermissionService";
+import { ModuleAccessService } from "@/backend/ModuleAccessService";
 import { Project } from "@/models/Project";
 
 export const POST = apiHandler(async (req, { session }) => {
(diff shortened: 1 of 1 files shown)


lemilonkh (approved):
This is a good fix for now, but let's look into why the project was created without access to the GHGI module if possible.

tatimoreira:
> This is a good fix for now, but let's look into why the project was created without access to the GHGI module if possible.

@lemilonkh Dug into this. The org's default project started being created fresh at runtime (via `Project.findOrCreate`) in c252c163 ("scope default project to the org-admin's own org"), which replaced a hardcoded fallback to a single pre-seeded `DEFAULT_PROJECT_ID`. That old default project had GHGI access only because of a one-time backfill migration (`20250725163645-ProjectModules_backfill.cjs`) that ran when the module-access system was introduced — it was never something the app granted at runtime.

So more precisely: there's no code path in the app that grants baseline module access when a project is created — it's always been either a one-time migration backfill or a manual toggle in the admin UI. `POST /api/v1/organizations/[organization]/projects` (admin-created projects) has the same gap. c252c163 just became the first runtime path to create a genuinely new project instead of reusing a pre-seeded one, which is what surfaced this.

This fix here is the right immediate patch. If we want to close the gap properly, I think the fix belongs at the project-creation layer (e.g. a small helper/hook that grants baseline modules whenever a `Project` row is created) rather than repeating the check-then-enable pattern at each call site — happy to file a follow-up ticket for that if it sounds right.

PR #3012, opened 2026-08-15
CC-710: Fix duplicated organization names in sidebar dropdown

tatimoreira:
## Summary

Fixes organization names appearing more than once in the sidebar/org-switcher dropdown, caused by duplicate `OrganizationAdmin` rows for the same user/org pair.

## Changes

- Migration: dedupe existing `OrganizationAdmin` duplicates (keeping the oldest row) and add a unique constraint on `(organization_id, user_id)`
- `OrganizationAdmin` model: declare the matching unique index
- `GET /api/v1/user/organizations`: defensive dedupe of the response
- Extracted a shared `uniqueBy()` helper (`src/util/array.ts`) and used it in both `JNDrawer.tsx` (sidebar) and `navigation-bar.tsx` (top nav), replacing an inline filter that only existed in one of the two places

## How to test

1. Log in as a user who is an admin on the same organization more than once (or seed a duplicate `OrganizationAdmin` row before running the migration).
2. Run `npm run db:migrate` in `app/`.
3. Open the sidebar drawer and the top-nav org switcher — each organization should appear exactly once.

## Ticket

CC-710


app/migrations/20260814205324-dedupe-and-unique-organization-admin.cjs
```diff
@@ -0,0 +1,36 @@
...
+                 PARTITION BY organization_id, user_id
+                 ORDER BY created ASC, organization_admin_id ASC
+               ) AS rn
+        FROM "OrganizationAdmin"
+      ) ranked
+      WHERE oa.organization_admin_id = ranked.organization_admin_id
+        AND ranked.rn > 1;
+    `);
+
+    await queryInterface.addIndex("OrganizationAdmin", {
```
cephaschapa: Consider creating the unique index CONCURRENTLY (Postgres) to avoid long locks. Sequelize’s addIndex may not create indexes CONCURRENTLY by default; if using Postgres consider a separate migration that runs CREATE UNIQUE INDEX CONCURRENTLY ... (this cannot run inside a transaction).

app/src/components/HomePage/JNDrawer.tsx
```diff
@@ -460,9 +461,18 @@ const JNDrawer = ({
...
-  const { data: organizations } = api.useGetUserOrganizationsQuery(undefined, {
-    skip: !isOpen,
-  });
+  const { data: rawOrganizations } = api.useGetUserOrganizationsQuery(
+    undefined,
+    {
+      skip: !isOpen,
+    },
+  );
+  const organizations = useMemo(
```
cephaschapa, line 470: The frontend dedupe is defensive/harmless, but if backend is fixed you could remove frontend dedupe. Keeping it is still okay and safe.

cephaschapa (approved):
Great work 👍👍

app/migrations/20260814205324-dedupe-and-unique-organization-admin.cjs
tatimoreira: great catch!

PR #3017, opened 2026-08-17
fix(test): reference-count shared DB connection to stop Jest worker race (CC-727)

tatimoreira:
## Summary

- `tests/api/city-hiap-action_plan.jest.ts` (and others) have been failing intermittently in CI with 404s / `"Ranked HIAP action not found"`, unrelated to the code under test — confirmed failing on `develop` before this branch existed, and again on an unrelated PR (#3012).
- Root cause: Jest's ESM support (`--experimental-vm-modules`) doesn't reset the module registry between test files in the same worker, so ~38 test suites' `beforeAll`/`afterAll` all share the same `db.sequelize` singleton from `src/models/index.ts`. One suite's `afterAll` closing the connection while another suite in the same worker (or a leftover un-awaited background job) was still using it threw `ConnectionManager.getConnection was called after the connection manager was closed!`, surfacing as unrelated 404s downstream.
- Fix: `db.initialize()` now reuses an already-live connection instead of creating a new pool, and `sequelize.close()` is reference-counted so it only actually tears the pool down once every caller that initialized it has also closed it. Single-file change, no test files touched; safe for production since `close()` is never called there.

## Verification

Reproduced the exact race locally (isolated Postgres container, not the shared dev DB) by running `tests/api/` with parallel workers on the pre-fix code — hit the identical `ConnectionManager...closed` error/stack trace. Re-ran the same batch with the fix applied: the race is gone, `city-hiap-action_plan.jest.ts` / `bulk-hiap-prioritization.jest.ts` / `import-routes.jest.ts` all pass. (4 remaining failures in that local run are unrelated — missing seed/catalog data from skipping `db:seed` locally, not this bug.)

Linear: CC-727

## Test plan
- [x] Reproduced the race on unfixed code locally
- [x] Confirmed the race is gone with the fix, same test batch
- [ ] CI green on this PR


cephaschapa (approved):
Great, thanks 👍

PR #3018, opened 2026-08-17
CC-708: Fix unexpected input validation in Create Organization flow

tatimoreira:
## Summary

Fixes inconsistent/unexpected validation errors on the description and city-limit fields in the admin Create Organization flow (step 2) and the Add/Edit Project modal.

## Changes

- The city-limit error box was gated on `errors.description` instead of `errors.cityCountLimit` (copy/paste bug), so it showed/hid based on the wrong field's validity. Fixed in both `CreateOrganizationModal.tsx` and `CreateEditProjectModal.tsx`.
- `FormattedNumberInput` pushes a string value on every keystroke while the Zod schema expected `number`, causing an untranslated "Expected number, received string" error to flash while typing. Fixed by changing `cityCountLimit` to `z.coerce.number()` in both schemas.
- Removed dead code in `FormattedNumberInput`: a `Controller.rules` block that RHF's resolver already superseded (using untranslated i18n keys not present in the `admin` namespace), and the unused `setError`/`clearErrors` props.

## How to test

1. Run `npm run dev`, log in as admin, and open **Admin Panel → Add an organization**.
2. On step 2 ("Project"), type a value into **City limit** while leaving **Description** blank — confirm no error box appears (previously it would, incorrectly).
3. Fill in a description that's too short (e.g. `"ab"`), then type a valid city-limit value — confirm only the description shows a "This field is required" error, and the city-limit box stays clean.
4. Repeat the same checks in **Admin Panel → \<org\> → Projects → Add a project**.

## Ticket

CC-708

## Notes

The `max={99999}` city-limit cap was previously enforced only by the now-removed dead `rules` block, so it was already non-functional (Zod schema has no `.max()`). Not fixed here since it's a separate missing-validation gap, not part of this ticket's "unexpected validation" symptom — flagging as a possible fast-follow.

### DIFF
--- a/app/src/app/[lng]/admin/CreateOrganizationModal.tsx
+++ b/app/src/app/[lng]/admin/CreateOrganizationModal.tsx
@@ -43,7 +43,7 @@ const schema = z.object({
   planType: z.nativeEnum(OrganizationPlanType),
   projectName: z.string().min(3, "required"),
   description: z.string().min(3, "required"),
-  cityCountLimit: z.number().min(1, "required"),
+  cityCountLimit: z.coerce.number().min(1, "required"),
 });
 
 type Schema = z.infer<typeof schema>;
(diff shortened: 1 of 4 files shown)


cephaschapa (approved):
Great spot. Thanks ✨👍

PR #3023, opened 2026-08-18
feat(api): CC-598 create MEED ranking route and data models

lemilonkh:
## Summary

Implements API route `/api/v1/city/CITY_ID/meed/rank`, along with the `MeedActionRanked` and `MeedActionRemoved` data models.

## Changes

- Created MeedApiService for interacting with the MEED microservice
- Implements API route `/api/v1/city/CITY_ID/meed/rank`
- Creates data models MeedActionRanked and MeedActionRemoved along with the respective migrations

## How to test

Make an API request to `/api/v1/city/CITY_ID/meed/rank` with the following body:
```json
{
  "inventoryId": "27aaf288-a90b-4600-8e1f-7f8f3cc1c78a",
  "requestedLanguages": ["en"],
  "topN": 5,
  "createExplanations": true,
  "cityDataList": [
    {
      "excludedActionIds": [],
      "weightsOverride": {},
      "cityStrategicPreferenceSectors": ["stationary_energy"],
      "cityStrategicPreferenceTimeframes": ["short"],
      "cityStrategicPreferenceCoBenefitKeys": []
    }
  ]
}
```

Verify data is returned after ~20 seconds (the model is run during the request in the MEED service).

## Ticket

CC-598

### DIFF
--- a/app/migrations/20260817120341-create-meed-action-ranked.cjs
+++ b/app/migrations/20260817120341-create-meed-action-ranked.cjs
@@ -51,6 +51,11 @@ module.exports = {
         allowNull: false,
         defaultValue: {},
       },
+      weights: {
+        type: Sequelize.JSONB,
+        allowNull: false,
+        defaultValue: {},
+      },
       created: {
         type: Sequelize.DATE,
         allowNull: false,
--- a/app/migrations/20260817120341-create-meed-action-ranked.cjs
+++ b/app/migrations/20260817120341-create-meed-action-ranked.cjs
@@ -44,7 +44,11 @@ module.exports = {
         type: Sequelize.FLOAT,
         allowNull: false,
       },
-      // TODO do we need evidence summary?
+      evidenceSummary: {
+        type: Sequelize.JSONB,
+        allowNull: false,
+        defaultValue: {},
+      },
       // this contains keys for each language
       explanations: {
         type: Sequelize.JSONB,
--- a/app/migrations/20260817120341-create-meed-action-ranked.cjs
+++ b/app/migrations/20260817120341-create-meed-action-ranked.cjs
@@ -24,6 +24,14 @@ module.exports = {
         type: Sequelize.TEXT,
         allowNull: false,
       },
+      rank: {
+        type: Sequelize.INTEGER,
+        allowNull: false,
+      },
+      final_score: {
+        type: Sequelize.FLOAT,
+        allowNull: false,
+      },
       lang: {
         type: Sequelize.TEXT,
         allowNull: false,
@@ -46,7 +54,7 @@ module.exports = {
     });
   },
 
-  async down(queryInterface, Sequelize) {
+  async down(queryInterface) {
     await queryInterface.dropTable("MeedActionRanked");
   },
 };
--- a/app/migrations/20260817120341-create-meed-action-ranked.cjs
+++ b/app/migrations/20260817120341-create-meed-action-ranked.cjs
@@ -51,11 +51,6 @@ module.exports = {
         allowNull: false,
         defaultValue: {},
       },
-      is_selected: {
-        type: Sequelize.BOOLEAN,
-        allowNull: false,
-        defaultValue: true,
-      },
       created: {
         type: Sequelize.DATE,
         allowNull: false,
--- a/app/migrations/20260817120341-create-meed-action-ranked.cjs
+++ b/app/migrations/20260817120341-create-meed-action-ranked.cjs
@@ -32,9 +32,24 @@ module.exports = {
         type: Sequelize.FLOAT,
         allowNull: false,
       },
-      lang: {
-        type: Sequelize.TEXT,
+      impact_score: {
+        type: Sequelize.FLOAT,
         allowNull: false,
+      },
+      alignment_score: {
+        type: Sequelize.FLOAT,
+        allowNull: false,
+      },
+      feasibilityScore: {
+        type: Sequelize.FLOAT,
+        allowNull: false,
+      },
+      // TODO do we need evidence summary?
+      // this contains keys for each language
+      explanations: {
+        type: Sequelize.JSONB,
+        allowNull: false,
+        defaultValue: {},
       },
       is_selected: {
         type: Sequelize.BOOLEAN,
--- a/app/migrations/20260817120341-create-meed-action-ranked.cjs
+++ b/app/migrations/20260817120341-create-meed-action-ranked.cjs
@@ -1,0 +1,52 @@
+"use strict";
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.createTable("MeedActionRanked", {
+      id: {
+        type: Sequelize.UUID,
+        allowNull: false,
+        primaryKey: true,
+        defaultValue: Sequelize.UUIDV4,
+      },
+      inventory_id: {
+        type: Sequelize.UUID,
+        allowNull: false,
+        references: {
+          model: "Inventory",
+          key: "inventory_id",
+        },
+        onUpdate: "CASCADE",
+        onDelete: "CASCADE",
+      },
+      action_id: {
+        type: Sequelize.TEXT,
+        allowNull: false,
+      },
+      lang: {
+        type: Sequelize.TEXT,
+        allowNull: false,
+      },
+      is_selected: {
+        type: Sequelize.BOOLEAN,
+        allowNull: false,
+        defaultValue: true,
+      },
+      created: {
+        type: Sequelize.DATE,
+        allowNull: false,
+        defaultValue: Sequelize.NOW,
+      },
+      last_updated: {
+        type: Sequelize.DATE,
+        allowNull: false,
+        defaultValue: Sequelize.NOW,
+      },
+    });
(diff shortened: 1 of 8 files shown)


app/src/backend/InventoryService.ts
```diff
@@ -78,4 +88,70 @@ export class InventoryService {
...
+        : undefined;
+    let activityUnit = (
+      data[`${titleKey}-unit`] ?? data["activity-unit"]
+    )?.toString();
+    if (activityUnit?.startsWith("units-")) {
+      activityUnit = activityUnit.slice("units-".length);
+    }
+
+    return {
+      //methodology?.id,
```
cephaschapa: We can remove this commented value prop.

app/src/models/init-models.ts
```diff
@@ -1108,6 +1126,24 @@ export function initModels(sequelize: Sequelize) {
...
+  Inventory.hasMany(MeedActionRanked, {
+    as: "meedActionRankeds",
+    foreignKey: "inventoryId",
+  });
+  MeedActionRemoved.belongsTo(Inventory, {
+    as: "inventory",
+    foreignKey: "inventoryId",
+  });
+  Inventory.hasMany(MeedActionRemoved, {
+    as: "meedActionRemoveds",
```
cephaschapa, line 1143: Typo here: meedActionRemoved

app/migrations/20260817120341-create-meed-action-ranked.cjs
```diff
@@ -0,0 +1,75 @@
...
+      // this contains keys for each language
+      explanations: {
+        type: Sequelize.JSONB,
+        allowNull: false,
+        defaultValue: {},
+      },
+      is_selected: {
+        type: Sequelize.BOOLEAN,
+        allowNull: false,
+        defaultValue: true,
```
cephaschapa: Should the default value be true in this case?

cephaschapa (approved):
Great work thanks 👍👍👍

app/src/models/init-models.ts
lemilonkh: This is auto generated and probably won't come up. Gramatically correct is meedActionsRemoved, but I think the plural suffix at the end makes it clearer that it's an array.

app/migrations/20260817120341-create-meed-action-ranked.cjs
lemilonkh: Column has been removed now

carlosgraffi:
Reviewed from the frontend side, against what the Actions & Plans v2 module needs and what the MEED+ prototype currently does at https://meed-frontend.openearth.dev. Thanks for turning this around quickly — the shape is right and most of it slots straight into what we have.

Three things you got right that are easy to get wrong, worth saying before the rest:

- **`totalEmissionsUnit: "kg"`** is honest. The prototype hardcoded `"t CO2e"` while sending kilograms; scoring uses a ratio so it never affected rankings, but it was wrong in report text. Good that this one matches reality.
- **The legal i18n mapping is correct.** `legalJustification: { en: legal_justification_en, es: legal_justification }` — the service puts *native Spanish* in `legal_justification` and the translation in `_en`, while `ownership_description` is the other way round. Easy to invert; you didn't.
- **`removedActions` is richer than the raw service response**, and flattening the descriptions into `{en, es}` objects means we can render them directly.

---

## Blocking

### `locode` and `countryCode` are hardcoded to Iquique

`app/src/backend/MeedApiService.ts:126-128`

```js
// TODO restore
locode: "CL IQQ",   // inventory.city.locode ?? "",
countryCode: "CL",  // inventory.city.countryLocode ?? "",
```

Every ranking is computed as Iquique, Chile, regardless of the inventory. A São Paulo ranking would be scored against Chilean legal screening, Chilean policy plans and Chilean climate finance — and the output would look entirely plausible, which is what makes it dangerous. Looks like debug scaffolding that outlived its purpose.

### No GET — a persisted ranking can't be read back

`app/src/app/api/v1/city/[city]/meed/rank/route.ts` exports `POST` only.

The ranking *is* persisted (two tables, in a transaction — that part is done), but there's no way to retrieve it. The frontend loads a ranking when the results screen mounts, so as it stands every visit would either re-run the full prioritization — LLM call plus upstream fetches, tens of seconds — or show nothing.

A `GET /api/v1/city/{city}/meed/rank?inventoryId=…` returning the stored `rankedActions` / `removedActions` would close this. Our side is a one-line swap once it exists.

### Re-running duplicates every row

`MeedApiService.ts:194,209` — `bulkCreate` with no delete-first, and neither migration adds a unique constraint on `(inventoryId, actionId)`.

"Re-run ranking" is a button we ship, so this will happen in normal use. Either delete the previous rows for that inventory inside the transaction, or add the unique constraint and upsert.

---

## Data we need that's currently dropped

### `evidence_summary` — specifically, not wholesale

You asked which fields matter. Rather than the whole blob, here's what each drives:

| Field | What it powers |
|---|---|
| `feasibility.legal.assessment_missing` | **The "Flagged — evidence missing" card.** These actions are *kept* in the ranking, so they never appear in `removedActions` — this is the only signal that identifies them |
| `impact.emissions_reduction_component_score`, `impact.timeline_component_score` | Impact score-bar explanation |
| `alignment.policy_component_score`, `sector_component_score`, `co_benefit_component_score`, `timeframe_component_score` | Alignment breakdown |
| `feasibility.legal.component_score`, `mitigation_feasibility.component_score`, `financial_feasibility.component_score` | Feasibility breakdown |
| `feasibility.financial_feasibility.route`, `.reason` | Why an action scored as it did on finance |

The prototype's detail drawer renders all of these as per-component explanations beneath each score bar, so this is existing behaviour rather than a new ask.

Simplest option is a single JSONB column holding `evidence_summary` as-is — lossless, no schema debate, and it can be flattened later if that's preferred. The one field we can't do without is `assessment_missing`.

### The resolved weights

To be clear, the route doesn't *stop* a user ranking with custom weights — `weightsOverride` is accepted and passed through, and the ranking will genuinely be scored with them. What's missing is any record of which weights produced a given ranking: `MeedActionRanked` has no weights column, and `metadata.weights` from the service is discarded.

That matters because our score-breakdown drawer prints the arithmetic — `impact × w + alignment × w + feasibility × w = finalScore` — next to the final score itself. With no stored weights we print the 55/22/23 defaults, so for anyone who moved the pre-flight sliders the printed sum won't equal the number beside it. Once a GET exists, a loaded ranking is also anonymous about its own inputs.

Storing the resolved `metadata.weights` alongside the ranking would cover it.

**Related, and cheaper to do now than later:** `POST /v1/reports/output-plan` requires a `prioritizationSnapshot` containing both the request *and* the response. Currently a flattened response is stored and no request. If report generation is on the roadmap, keeping the request envelope now avoids reconstructing it.

---

## Worth verifying before merge

These may be fine — I couldn't confirm either without running it.

**`totalEmissions` is a `bigint`.** `InventoryService.extractActivityFields` returns `totalEmissions: activity.co2eq`, and `ActivityValue.co2eq` is typed `bigint`. `JSON.stringify` throws on a BigInt; if Sequelize hands back a string instead, the service expects `float | None` and will 422. One live request settles it.

**Emissions are read only from `ActivityValue` rows.** `MeedApiService.ts:144` builds `activities` from `inventoryValue.activityValues`, so an inventory value carrying `co2eq` directly with no activity rows produces `activities: []` and those emissions never reach the service. Direct-measure entries and bulk imports are exactly that shape — worth checking against a real inventory, since it would quietly shrink the emissions denominator rather than fail.

---

## Smaller things

- **`MeedApiService.ts:142`** — `[inventoryValue.gpcReferenceNumber ?? ""]` creates a `""` key when the reference is null. The service's subsector mapping requires at least `X.Y` and raises on anything shorter, so this would 422 rather than being ignored.
- **`MeedApiService.ts:139-152`** — spreading into the accumulator under `gpcReferenceNumber` means **duplicate GPC references overwrite each other** rather than accumulating. An inventory with two values on the same reference silently loses one. (The prototype had this same bug, for what it's worth.)
- **`MeedApiService.ts:167`** — `generatedAtUtc: new Date().toUTCString()` produces `"Mon, 18 Aug 2026 12:02:14 GMT"`. The field name and the rest of the contract imply ISO 8601; `toISOString()` is what the prototype sent.
- **`MeedApiService.ts:182`** — `console.log("MEED result", resultString)` logs the entire response body on every call. Probably debug leftover.
- **`route.ts`** — `cityStrategicPreferenceSectors` is `z.array(z.string())`, so an invalid tag passes our boundary and fails at the service as a 422 with our payload in its logs. Constraining it to the five canonical tags would fail earlier with a readable message. Same for the co-benefit keys.

---

## Frontend status

For context on what's waiting on this: the module reads `removedActions` for the Regulations screen, reads the pillar scores and explanations for results, and has a single point where the ranking is loaded — so swapping from local storage to the API is small once the GET exists. Nothing on our side hardcodes a city; we send `inventoryId` and let CityCatalyst resolve the rest, which is exactly the split this PR sets up.

Happy to pair on any of these, or to take the frontend half in parallel.

lemilonkh:
Thanks! The blockers have been fixed. I'll add the evidence_summary fields in an upcoming push, as well as storing the returned weights if possible.

carlosgraffi (approved):
Thanks Milan!

PR #3035, opened 2026-08-19
CC-716: Manage concept notes (rename, duplicate, delete)

piotrnowakowski:
## Summary

- add the design-aligned card actions: pencil rename, duplicate, shared export dialog, and permanent delete
- add small authenticated CityCatalyst proxy routes and RTK Query mutations for rename, duplicate, and delete
- implement lifecycle behavior directly in Climate Advisor: rename the run/chat, copy current context and workspace records into a fresh run/chat, or delete the workspace/run/chat
- reuse immutable Markdown artifacts by key; do not add archive state, lifecycle tables, lifecycle migrations, or OCR-binding APIs

## Scope

This is a stacked PR based on #3013 (`codex/CC-604-cnb-implement-concept-note-builder-frontend`) so the review shows only CC-716 changes. Once #3013 merges, this PR can be retargeted to `develop`.

## Validation

- CityCatalyst focused Jest: 13 passed
- CityCatalyst TypeScript: passed
- scoped CityCatalyst ESLint: passed
- Climate Advisor focused pytest: 25 passed, 1 skipped
- Python compile check: passed
- `git diff --check`: passed

Linear: [CC-716](https://linear.app/openearth/issue/CC-716/cnb-manage-concept-notes-rename-duplicate-delete)

https://github.com/user-attachments/assets/eea615ca-553d-4661-9dbf-5a3142312590


app/src/app/api/v1/concept-notes/[runId]/route.ts
```diff
@@ -54,19 +54,41 @@ export const GET = apiHandler(async (req, { session, params }) => {
...
-        "Climate Advisor returned an invalid concept-note run",
-      );
-    }
-    await PermissionService.canAccessCity(session, run.data.city_id, {
-      includeResource: false,
-    });
+export const PATCH = apiHandler(async (req, { session, params }) => {
+  if (!session?.user?.id) {
+    throw new createHttpError.Unauthorized("Authentication required");
   }
```
Comment (lemilonkh) [inline app/src/app/api/v1/concept-notes/[runId]/route.ts]: Should we be doing further access control here (e.g. check if the user owns/ has access to the concept note)? Or is this handled in the Python API?

Same for the other routes.

app/src/backend/concept-notes.ts
```diff
@@ -32,6 +39,7 @@ export async function callConceptNoteApi(
     body: request.body,
     searchParams: request.searchParams,
     headers: {
+      ...Object.fromEntries(new Headers(request.headers).entries()),
```
lemilonkh: Why do you need to copy the headers object here?
```suggestion
      ...request.headers,
```
This should be fine I think?

app/src/backend/concept-notes.ts
```diff
@@ -52,3 +60,23 @@ export async function readConceptNoteApiPayload(
...
+  response: Response,
+  session: AppSession,
+): Promise<NextResponse> {
+  const payload = await readConceptNoteApiPayload(response);
+  if (response.ok) {
+    const run = upstreamRunSchema.safeParse(payload);
+    if (!run.success) {
+      throw new createHttpError.BadGateway(
+        "Climate Advisor returned an invalid concept-note run",
+      );
```
lemilonkh, line 99: If you either log or return the original validation error here it's easier to debug what went wrong when it happens. You can use `Logger.error({error: run}, "Climate Advisor returned and invalid concept-note run")` to log it.

app/src/backend/concept-notes.ts
```diff
@@ -52,3 +60,23 @@ export async function readConceptNoteApiPayload(
...
+    }
+    await PermissionService.canAccessCity(session, run.data.city_id, {
+      includeResource: false,
+    });
```
lemilonkh: Can we perhaps run this access control call before calling out to the climate assistant API? Would be preferable in case the route makes internal database/ file changes or calls to an LLM to avoid the costs if the user doesn't have access.
This would require knowing the city ID in the original request, not in this function that's parsing the response.

app/src/components/ConceptNoteDashboard/index.tsx
```diff
@@ -316,6 +373,20 @@ export function ConceptNoteDashboard({
                     progress={progress}
                     resumeHref={conceptNoteResumeHref(lng, cityId, run.run_id)}
                     resumeLabel={t("resume")}
+                    renameLabel={t("rename")}
+                    duplicateLabel={t("duplicate")}
+                    exportLabel={t("export")}
+                    deleteLabel={t("delete")}
```
lemilonkh: I think these labels could be translated in the RunCard component itself when you pass the `t` function itself, that way it's easier to find the translation keys right in the markup that uses them and you don't need to create tons of properties for the component.

Only if the labels don't change for different instances of the RunCard component of course.

app/src/backend/concept-notes.ts
```diff
@@ -1,17 +1,24 @@
...
 type ConceptNoteApiRequest = {
   path: string;
   userId: string;
-  method?: "GET" | "POST";
+  method?: "GET" | "POST" | "PATCH" | "DELETE";
   body?: Record<string, unknown>;
+  headers?: HeadersInit;
   requestId?: string;
   searchParams?: Record<string, string>;
 };
```
lemilonkh, line 25: Can we pass the cityId in here for doing the access control before the API call?

app/src/backend/concept-notes.ts
lemilonkh: Let's do access control based on the cityId in this function if possible rather than in the function that parses the API result.

lemilonkh (approved):
Awesome, thanks 👍🏻✨

app/src/app/api/v1/concept-notes/[runId]/route.ts
Comment (piotrnowakowski) [reply app/src/app/api/v1/concept-notes/[runId]/route.ts]: Good point. Climate Advisor remains authoritative for checking run ownership and live access to the run's stored city. I've also added a CityCatalyst-side check for the client-supplied `city_id` before issuing the CA token or calling upstream. This now applies to GET, PATCH, duplicate, and DELETE. Successful run responses are also checked against the requested city.

app/src/backend/concept-notes.ts
piotrnowakowski: You're right. The callers only provide plain header records, so I narrowed the type to `Record<string, string>` and now spread `...request.headers` directly. I also applied the same simplification to the only equivalent helper elsewhere in the repository.

app/src/backend/concept-notes.ts
piotrnowakowski: Done. The failed Zod result is now logged with `logger.error({ error: run.error }, ...)` before returning the 502 response. This keeps the validation details available server-side without exposing them to the client.

app/src/backend/concept-notes.ts
piotrnowakowski: Done. The lifecycle requests now include `city_id`, which is passed to `callAuthorizedConceptNoteApi`. That wrapper runs `PermissionService.canAccessCity` before issuing the Climate Advisor token or making the upstream request, avoiding unnecessary database, file, or LLM work when access is denied.

app/src/components/ConceptNoteDashboard/index.tsx
piotrnowakowski: Agreed. These labels are the same for every `RunCard`, so the component now receives `t` and resolves the action labels directly in the markup and ARIA labels. The five label props were removed from the parent.

app/src/backend/concept-notes.ts
piotrnowakowski: Done through a dedicated authorized request wrapper. `cityId` is now part of `AuthorizedConceptNoteApiRequest` and is checked before token issuance and the upstream call. The lower-level `callConceptNoteApi` remains focused on transport concerns.

app/src/backend/concept-notes.ts
piotrnowakowski: Done. The access check now happens in the authorized wrapper before calling Climate Advisor. `conceptNoteRunResponse` no longer performs a late permission lookup; it validates the successful response schema and confirms that the returned `city_id` matches the requested city. Climate Advisor remains authoritative for actual run ownership and its stored city binding.

PR #3038, opened 2026-08-20
Fix HIAP action plan save and error/email handling

cephaschapa:
## Summary

Fixes HIAP action plan generation failing to persist ranked plans (`Ranked HIAP action not found`) by using the ranked-action ID correctly. Also surfaces generation/save failures to the UI with an error toast, and only sends the ready email after a successful first-time DB save.

## Changes

- Pass `action.id` as `highImpactActionRankedId` for ranked actions; skip ranked-context validation for unranked plans
- Propagate DB save failures instead of returning success without a saved plan
- Await plan generation in `GeneratePlanDrawer` and show an error toast on failure
- Update generation API tests and failure copy

## How to test

1. Open HIAP for a city with ranked actions and generate an action plan.
2. Confirm the plan is saved and appears under “View Generated Plan”.
3. Confirm a ready email is sent only on first successful create.
4. Force or simulate a generation/save failure and confirm an error toast appears (no email).

## Ticket

CC-748

### DIFF
--- a/app/src/backend/hiap/ActionPlanService.ts
+++ b/app/src/backend/hiap/ActionPlanService.ts
@@ -462,7 +462,9 @@ export default class ActionPlanService {
         input.highImpactActionRankedId = undefined;
       }
 
-      if (input.inventoryId || input.highImpactActionRankedId) {
+      // Ranked plans must resolve against HighImpactActionRanked. Unranked plans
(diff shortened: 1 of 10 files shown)


lemilonkh (approved):
Nice work, thanks for adding the extra test as well 🙌

PR #3042, opened 2026-08-21
 Fix HIAP action plan generate timeouts with 202 Accepted

cephaschapa:
## Summary

Action plan generation no longer holds the HTTP request open while polling HIAP (which caused ingress 504s in prod). The API returns 202 immediately and finishes poll/save/email in the background; users are notified by email when the plan is ready.

## Changes

- `POST .../action-plan/generate` returns 202 and fire-and-forgets `startActionPlanJob` (same pattern as inventory import approve)
- Frontend treats acceptance as success; relies on the existing started toast + ready email
- RTK mutation typed for the 202 payload; tests updated for 202

## How to test

1. Generate an action plan from HIAP and confirm the request returns quickly with 202 (no long hang / 504).
2. Confirm the info toast still appears on start.
3. Confirm the plan is saved and the ready email is sent after HIAP finishes.
4. Confirm validation/auth failures still return 4xx and show the error toast.

## Notes

No client-side polling in this PR — completion is email-only by design.


lemilonkh (approved):
Awesome, thanks! 👍🏻✨

PR #3044, opened 2026-08-21
CC-752: Fix HIAP cron heap exhaustion by isolating catalog backfill

DavidGBbr:
## Summary

- Remove NativeInputCatalog backfills from the lightweight HIAP polling endpoint.
- Move ranking and action-plan reconciliation to a bounded, keyset-paginated command and Kubernetes Job.
- Add idempotency, partial-failure handling, advisory-lock protection, and safe dry-run behavior.
- Add a Docker PostgreSQL fixture and k6 load-test evidence for repeatable validation.

## Evidence

A controlled local Docker comparison used 1,000 successful rankings, 3,000 ranked actions, and 1,000 action plans with 32 KiB JSONB payloads. At 2 requests/second for 60 seconds:

| Metric | Baseline (origin/develop) | Fixed |
| --- | ---: | ---: |
| Completed requests | 12 | 120 |
| Dropped iterations | 101 | 0 |
| HTTP failures | 100% timeouts | 0% |
| p95 latency | 59,995.7 ms | 11.1 ms |
| Node private-memory peak | 1,058.28 MiB | 157.21 MiB |

The baseline cold request synchronously backfilled 1,000 rankings and 1,000 plans. The fixed endpoint returned the polling-only response contract in all 120 requests.

Full card-ready evidence: load-test/cc-752-local-evidence.md.

## Verification

- 13 focused Jest tests passed.
- Targeted ESLint, Prettier, and TypeScript checks passed.
- Next production build passed.
- Local Docker migrations and deterministic fixture seeding passed.
- k6 fixed-build run passed with 120/120 requests and zero unexpected backfill fields.

## Follow-up

This PR contains local evidence only. Approved staging validation is still required for Kubernetes pod restarts and kubectl top memory telemetry before re-enabling the production CronJob or running the first production backfill.

Linear: CC-752


app/src/backend/hiap/HiapCatalogBackfillRunner.ts
```diff
@@ -0,0 +1,185 @@
...
+    ),
+    maxBatches,
+    dryRun: env.HIAP_CATALOG_BACKFILL_DRY_RUN?.toLowerCase() === "true",
+  };
+}
+
+async function acquireHIAPCatalogBackfillLock(): Promise<boolean> {
+  if (!db.sequelize) throw new Error("Database not initialized");
+
+  const rows = await db.sequelize.query<{ locked: boolean }>(
```
mircorudolph: pg_try_advisory_lock is session-scoped, but this query returns its connection to Sequelize’s pool immediately. The later unlock query may receive a different PostgreSQL session, which cannot release the lock; the current CLI eventually closes the pool, but explicit lock ownership and concurrent-run protection are still not guaranteed. Please acquire one dedicated pool connection, hold it for the entire acquire → backfill → unlock lifecycle, execute both lock operations on that connection, and release it in finally; ideally add an integration test with two connections proving that a concurrent runner is skipped

app/tests/hiap-catalog-backfill-runner.jest.ts
```diff
@@ -0,0 +1,89 @@
...
+        HIAP_CATALOG_BACKFILL_DRY_RUN: "true",
+      }),
+    ).toEqual({ batchSize: 40, maxBatches: 3, dryRun: true });
+  });
+
+  it("processes ranking pages before action-plan pages and carries the cursor", async () => {
+    const rankingPages: HIAPCatalogBackfillPageOptions[] = [];
+    const actionPlanPages: HIAPCatalogBackfillPageOptions[] = [];
+    const deps: HIAPCatalogBackfillRunnerDeps = {
+      acquireLock: jest.fn().mockResolvedValue(true),
```
mircorudolph: This test uses jest.fn() without importing jest, which is not automatically available in this repository’s ESM Jest configuration. The current PR workflow consequently fails with ReferenceError: jest is not defined in two tests, preventing the build and deployment jobs from running. Import jest from @jest/globals, rerun the complete workflow, and require a green result before merging.

app/src/backend/hiap/HiapCatalogBackfillRunner.ts
```diff
@@ -0,0 +1,185 @@
...
+
+  const state = { pages: 0 };
+  try {
+    const rankings = await processPages(
+      deps.processRankingsPage,
+      config,
+      state,
+    );
+    const actionPlans =
+      config.maxBatches !== undefined && state.pages >= config.maxBatches
```
mircorudolph: maxBatches limits work without preserving either cursor, so every invocation restarts from the first ranking and action-plan pages. Separate per-type budgets would prevent rankings from starving action plans within one run, but records beyond either budget would still never be reached; please persist and resume both cursors or query only missing/outdated catalog records so completed pages disappear from subsequent runs.

Do you think it would be better to always check all entries and persist the cursor or to only check missing/ out dated?

mircorudolph:
I added a dedicated test-environment HIAP CronJob and updated the test deployment workflow. 
Previously, web-test.yml applied the generic CronJob, which calls the dev cc-web service, while the existing test-specific resource still called the obsolete /api/cron/check-hiap-jobs path. 
The new k8s/test/cc-test-check-hiap-jobs.yml updates citycatalyst-test-check-hiap-jobs in place and calls http://cc-test-web:3000/api/v1/cron/check-hiap-jobs with the cron bearer key every five minutes.
web-test.yml now applies this test-specific manifest before setting CC_CRON_JOB_API_KEY and no longer applies the generic dev CronJob. The generic CronJob remains unchanged for the dev and production workflows.
@lemilonkh please have a look if this particular change was correct.

mircorudolph (approved):
Thank you!

app/src/backend/hiap/HiapCatalogBackfillRunner.ts
DavidGBbr: Addressed in commit bd052bcf0. The runner now acquires one dedicated Sequelize pool connection, keeps it for the full acquire → backfill → unlock lifecycle, executes both advisory-lock operations on that same connection, and releases it in finally. I also added a regression test with two connections proving that a concurrent acquisition is skipped and that both connections are released correctly. The focused tests pass locally; CI/staging validation is still pending.

app/tests/hiap-catalog-backfill-runner.jest.ts
DavidGBbr: Addressed in commit bd052bcf0 by importing jest explicitly from @jest/globals. The runner test and the related HIAP tests pass locally. GitHub checks are still pending.

app/src/backend/hiap/HiapCatalogBackfillRunner.ts
DavidGBbr: Implemented the durable per-type checkpoint approach in commit bd052bcf0. The new HiapCatalogBackfillCheckpoint migration stores independent rankings and action-plans cursors/completion flags after each page; failed pages keep their current cursor so they can be retried. HIAP_CATALOG_BACKFILL_MAX_BATCHES is now an independent budget per type, and completed types are skipped on later runs. I chose this approach because the backfill currently needs to compute source identities/content digests before deciding whether an entry is missing or outdated; an always-check reconciliation query would be a larger redesign. Dry runs intentionally do not read or write checkpoints. Focused tests pass locally; CI/staging validation is still pending.

PR #3045, opened 2026-08-21
CC-745: Allow negative emissions values in manual input and uploads

tatimoreira:
## Summary

Cities with emissions sinks (e.g. Chile inventories) couldn't record negative emissions values — several non-negative validation checks silently dropped or rejected them in both the Manual Input and Upload Inventories features. This removes those checks so negative values (sinks/removals) can be entered and are correctly persisted and aggregated.

## Changes

- `util/validation.ts`: removed `.gte(0)`/`.gte(0n)` constraints on `co2eq`, `gasAmount`, and `emissionsPerActivity` in the manual-input API schemas.
- `backend/formulas.ts`: removed the hard `BadRequest` thrown on negative gas amounts in `handleDirectMeasureFormula`.
- `components/Modals/activity-modal/activity-modal-body.tsx`: removed the client-side validation error blocking negative custom emission factors.
- `backend/InventoryImportService.ts`: fixed the eCRF import, which used a `totalCO2e > 0` check that silently skipped rows with negative `totalCO2e` (misclassifying them as "no emissions data") instead of importing them.

## How to test

1. **Manual Input**: open a Direct Measure activity entry and enter a negative CO2/CH4/N2O amount — it should save instead of erroring with "Gas amount cannot be negative".
2. **Upload Inventories**: upload an eCRF file with a negative value in `GHGs (metric tonnes CO2e) - Total CO2e` for a row — it should import instead of being silently skipped, and the sector/inventory total should reflect the subtraction.

Verified end-to-end locally: uploaded a file with `Total CO2e = -3.14` for GPC ref `I.2.1`; confirmed via direct DB inspection that `ActivityValue.co2eq = -3140` (kg) persisted and is correctly summed into `InventoryValue.co2eq` (reducing the aggregate total accordingly).

## Ticket

CC-745

## Notes

While testing, found a pre-existing, unrelated bug in `ECRFDownloadService.transformDataForTemplate3`: it does `BigInt(gasValue?.gasAmount)` without a null check, which throws when an `ActivityValue` has `co2eq` but no per-gas `GasValue` rows (older/legacy data). Not fixed here — happens regardless of value sign — but worth its own ticket since it currently breaks eCRF download for affected inventories.


lemilonkh (approved):
Nice, thanks 👍

PR #3049, opened 2026-08-24
CC-756: Pre-select project so onboarding city shows in invite step

tatimoreira:
## Summary

Fixes a bug where a city created during project onboarding never appeared as a selectable checkbox on the "invite collaborators" step, because that step tracked its own independent project selection with nothing to seed it.

## Changes

- `setup/page.tsx`: capture the `projectId` returned by the `addCity` mutation and pass it to `InviteCollaboratorsStep` as `createdProjectId`.
- `invite-collaborators-step.tsx`: accept `createdProjectId` and auto-select it once it's present in the fetched project list (only if the user hasn't already picked a project manually), so the new city's checkbox renders without requiring a manual dropdown guess.

## How to test

1. Run through `/cities/onboarding/setup`: create a new city (step 0), fill population data (step 1).
2. On the invite-collaborators step, confirm the project dropdown is pre-selected and the newly created city appears in the checkbox grid.
3. Manually switch the project dropdown to confirm selection still updates the checkbox list correctly.

## Ticket

CC-756


lemilonkh (approved):
Great, thanks 👍 ✨

PR #3050, opened 2026-08-24
CC-758: Animate Clima typing indicator

piotrnowakowski:
## Summary

Quick fix to replace Clima's visible thinking text/static ellipsis with a standard three-dot typing indicator that moves vertically while a Concept Note response is being generated.

## Changes

- Render three compact dots with staggered up-and-down animation.
- Remove visible “Clima is thinking…” text from the assistant bubble.
- Add an accessible status label and disable animation for reduced-motion users.
- Hide the indicator as soon as streamed response content appears.

https://github.com/user-attachments/assets/304d1f78-be6e-4263-ab70-915e8ec1fa80


app/src/components/ConceptNoteWorkspace/chat-panel.tsx
```diff
@@ -301,21 +336,23 @@ export function ConceptNoteChatPanel({
...
+            {message.role === "assistant" && message.text ? (
               <ReactMarkdown
                 remarkPlugins={[remarkGfm]}
                 components={assistantMarkdownComponents}
               >
-                {message.text || t("chat-thinking")}
+                {message.text}
               </ReactMarkdown>
+            ) : message.role === "assistant" ? (
+              <TypingIndicator label={t("chat-generating")} />
```
lemilonkh, line 347: Should we add the same component to the Clima assistant itself perhaps? Doesn't have to be in this PR of course.

lemilonkh (approved):
Nice touch! Thanks ✨

app/src/components/ConceptNoteWorkspace/chat-panel.tsx
piotrnowakowski: I think we use pule icon there i will ask Carlos how that should be aligned thankt for surfacing

PR #3052, opened 2026-08-25
 CC-486: Standardize emissions unit labels in the UI

cephaschapa:
## Summary

Standardizes in-app GHG emissions unit labels to the product set (`mgCO₂e` … `TtCO₂e`) via shared helpers, and fixes SourceDrawer so the “total included” figure is not corrupted by a `formatNumber` → `parseFloat` round-trip.

## Changes

- Unify `formatEmissions` / `convertKgToTonnes` on full Unicode unit labels (input remains kg)
- Update hero, settings, public, versions, draft, import preview, and chart legend callers
- Leave CSV/eCRF/CIRIS export unit strings unchanged (`t CO2e`)
- Fix SourceDrawer total to format `co2eq_100yr` (× scale) directly
- Add `emissions-units.jest.ts`; update related draft/e2e expectations

## How to test

1. `npm run jest:windows -- --testPathPatterns=emissions-units --no-coverage`
2. GHGI home / results / activity totals show labels like `mtCO₂e` / `ktCO₂e` (not `t` + appended `CO2e`)
3. Open a data-step SourceDrawer: header total and disaggregated activity total should match scale
4. CSV download unit column still `t CO2e`

## Ticket

[CC-486](https://linear.app/openearth/issue/CC-486/ensure-consistency-in-emissions-units-format)

## Notes

- Display-only; DB `co2eq` stays in kg.
- SourceDrawer fix is a latent scaling bug made more obvious by the new unit prefixes.


app/src/components/StationaryEnergyDraft/utils.ts
```diff
@@ -272,21 +272,33 @@ function normalizeEmissionsUnitLabel(unit: unknown): string | null {
...
   }
 
-  const normalized = text.replace(/\s+/g, "").toLowerCase();
-  if (normalized === "kgco2e") {
-    return "kg CO2e";
+  const compact = text.replace(/\s+/g, "");
+  const normalized = compact.toLowerCase().replace("₂", "2");
+
+  // Megatonnes use capital M (MtCO₂e); metric tonnes use lowercase mtCO₂e.
+  if (compact.startsWith("Mt") || compact.startsWith("MT")) {
```
tatimoreira, line 279: `normalizeEmissionsUnitLabel` distinguishes `MtCO₂e` (megatonnes, 1e9 kg) from
`mtCO₂e` (metric tonnes, 1e3 kg) purely by whether the source string's `Mt`/`MT`
prefix is capitalized

I couldn't find any real data path in this repo that actually produces a
capitalized `Mt`/`MT` unit — every existing fixture
(`climate-advisor/service/tests/fixtures/stationary_energy_load_context_mock.json`)
uses lowercase `tCO2e`. The only place `MtCO2e` appears is the new
`SEEG`-labeled fixture added in `stationary-energy-draft-flow.jest.ts`, and
"SEEG" isn't wired to any ingestion adapter here — so the megatonne branch is
currently validated only by a fixture this PR itself authored.

Given this reaches `currentValueLabel`, i.e. the draft-review UI a reviewer
uses to approve/reject proposed inventory values, a mislabel here would be a
silent, high-impact data-integrity issue rather than a cosmetic one.

Could we either:
- confirm (with a real payload example) that upstream sources reliably
  capitalize `Mt`/`MT` for megatonnes and lowercase `mt` for metric tonnes, or
- make the megatonne/metric-tonne distinction explicit rather than
  case-derived (e.g. require an unambiguous token from the source, or fail
  loudly/fall back to a safe default instead of guessing on case)?

app/src/components/GHGI/data-step/SourceDrawer.tsx
```diff
@@ -75,32 +75,33 @@ export function SourceDrawer({
...
+
     const emissionsData = sourceData?.totals?.emissions?.co2eq_100yr;
-    let totalEmissions = emissionsData
-      ? formatNumber(
-          (Number(emissionsData) * (sourceData?.scaleFactor ?? 1.0)) / 1000,
-          numberFormat,
-          2,
-        )
-      : "?";
     if (sourceData?.issue) {
```
tatimoreira: **Question: is hiding the total on `sourceData.issue` intentional?**

https://github.com/Open-Earth-Foundation/CityCatalyst/blob/30918dc0dce59b13846f678cfbe8966524b9587a/app/src/components/GHGI/data-step/SourceDrawer.tsx#L88-L90

Before this change, if `sourceData.issue` was set but a valid
`totalEmissionsData` override was passed in, the drawer would still show that
number. Now it always shows `"?"` as soon as `sourceData.issue` is set, even
if a good number is available.

This is reachable in `activity-tab.tsx`, where `totalEmissionsData` and
`sourceData` come from the same object — so a flagged source that still has a
real total would now show `?` instead of the number it used to show.

Was this intentional, or a side effect of the refactor?

tatimoreira (changes requested):

tatimoreira (approved):

PR #3053, opened 2026-08-25
perf(hiap): add indexes for catalog backfill

DavidGBbr:
## Summary

Add the PostgreSQL indexes identified during the production-like synthetic test of the HIAP catalog backfill.

- Add a concurrent index on `HighImpactActionRanked.hia_ranking_id` for the ranking join.
- Add a partial concurrent index on successful rankings ordered by `created, id`.
- Add migration coverage for creation and rollback.

## Validation

With 11,747 synthetic rankings and 469,880 ranked-action rows, the equivalent page query changed from a parallel sequential scan/hash join to index-only scans/nested loop:

- Before: ~103 ms locally
- After: ~0.065 ms locally

The migration was also executed against a temporary PostgreSQL instance and rolled back successfully. Focused tests pass 18/18, plus TypeScript, ESLint, Prettier, and diff checks.

Related to #3044.

### DIFF
--- a/app/migrations/20260825120000-add-hiap-catalog-backfill-indexes.cjs
+++ b/app/migrations/20260825120000-add-hiap-catalog-backfill-indexes.cjs
@@ -1,0 +1,30 @@
+"use strict";
+
+const RANKED_ACTIONS_INDEX = "idx_hiap_ranked_hia_ranking_id";
+const SUCCESS_RANKINGS_INDEX = "idx_hiap_ranking_success_created_id";
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    // These indexes are created concurrently because the tables are populated
+    // in production and the backfill is not allowed to block live requests.
+    await queryInterface.sequelize.query(`
+      CREATE INDEX CONCURRENTLY IF NOT EXISTS ${RANKED_ACTIONS_INDEX}
+      ON "HighImpactActionRanked" (hia_ranking_id);
+    `);
+    await queryInterface.sequelize.query(`
+      CREATE INDEX CONCURRENTLY IF NOT EXISTS ${SUCCESS_RANKINGS_INDEX}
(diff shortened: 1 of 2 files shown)


app/migrations/20260825120000-add-hiap-catalog-backfill-indexes.cjs
```diff
@@ -0,0 +1,30 @@
...
+
+const RANKED_ACTIONS_INDEX = "idx_hiap_ranked_hia_ranking_id";
+const SUCCESS_RANKINGS_INDEX = "idx_hiap_ranking_success_created_id";
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    // These indexes are created concurrently because the tables are populated
+    // in production and the backfill is not allowed to block live requests.
+    await queryInterface.sequelize.query(`
```
mircorudolph, line 11: CREATE INDEX CONCURRENTLY can leave an invalid same-named index after interruption or failure, and IF NOT EXISTS will silently skip it on the migration retry. The migration could then be recorded as successful while the backfill still lacks a usable index; detect and drop/reindex invalid indexes before creation, or remove IF NOT EXISTS so recovery cannot silently succeed. [PostgreSQL documentation](https://www.postgresql.org/docs/current/sql-createindex.html)

mircorudolph (approved):

app/migrations/20260825120000-add-hiap-catalog-backfill-indexes.cjs
DavidGBbr: Addressed in 08027cc97. Removed IF NOT EXISTS from both CREATE INDEX CONCURRENTLY statements, so an invalid same-named index cannot be silently accepted on retry; the migration now fails and requires explicit cleanup/rebuild. Added test assertions for this behavior. The 4 related Jest suites (18 tests), ESLint, Prettier, and diff checks pass.

PR #3054, opened 2026-08-25
CC-736: Register completed HIAP-MEED rankings

DavidGBbr:
## Summary

Implements the first CC-736 slice for completed persisted HIAP-MEED rankings by storing a durable, versioned `MeedRanking` source and registering a pointer-only `hiap_meed_ranking` entry in NativeInputCatalog.

## Changes

- Adds versioned `MeedRanking` parent persistence with linked ranked/removed child rows; preserves legacy child rows.
- Registers pointer-only artifacts with server-resolved inventory/city/project/organization/user scope.
- Makes retries idempotent, supersedes changed versions, excludes failed/incomplete/empty rankings, supports backfill retry, and withdraws entries on inventory/city deletion.
- Adds focused adapter/API lifecycle tests and NativeInputCatalog documentation.

## How to test

1. `cd app`
2. `npm run jest -- --runInBand --coverage=false tests/meed-native-input-catalog-service.jest.ts tests/meed-api-service.jest.ts tests/native-input-catalog-service.jest.ts tests/hiap-native-input-catalog-service.jest.ts tests/ghgi-native-input-catalog-service.jest.ts`
3. `npx tsc --noEmit --pretty false`
4. `node --check migrations/20260824120000-create-meed-ranking.cjs`

## Ticket

[CC-736](https://linear.app/openearth/issue/CC-736/register-completed-hiap-meed-artifacts-in-nativeinputcatalog)

## Notes

- Reports/output plans and caller-provided city data/context/preferences are intentionally deferred until durable reusable source records exist.
- Full Jest is environment-limited here because the sandbox cannot connect to `127.0.0.1:5432`; focused verification passes (5 suites, 30 tests).


app/migrations/20260824120000-create-meed-ranking.cjs
```diff
@@ -0,0 +1,111 @@
...
+      last_updated: {
+        type: Sequelize.DATE,
+        allowNull: false,
+        defaultValue: Sequelize.NOW,
+      },
+    });
+
+    await queryInterface.addIndex(
+      "MeedRanking",
+      ["inventory_id", "input_digest", "content_digest"],
```
mircorudolph: The catalog source identity and supersession stream include userId, but this unique key—and the lookup in MeedApiService—do not. Two users producing identical input/output for the same inventory will reuse the first user’s MeedRanking, preventing creation of a correctly user-scoped catalog entry; include user_id consistently in the durable identity/query, or explicitly make both records inventory-scoped and remove the user dimension.

app/src/backend/meed/MeedNativeInputCatalogService.ts
```diff
@@ -0,0 +1,241 @@
...
+  const input = await buildMEEDRankingInput(ranking);
+  const existing = await models().NativeInputCatalog.findOne({
+    where: {
+      owningModule: MEED_MODULE,
+      sourceType: MEED_RANKING_SOURCE_TYPE,
+      sourceId: input.sourceId,
+      availability: "active",
+    },
+  });
+  if (existing) return { catalog: existing, created: false };
```
mircorudolph: If the replacement catalog entry is created but superseding an older entry fails, the next retry finds the replacement here and returns without repairing the still-active old version. Always run the idempotent supersedePreviousVersions reconciliation for the resolved registration, including when the replacement already exists.

app/src/backend/meed/MeedNativeInputCatalogService.ts
```diff
@@ -0,0 +1,241 @@
...
+
+  const registration = await registerNativeInput(input);
+  if (registration.created) {
+    await supersedePreviousVersions(input, registration.catalog.id);
+  }
+  return registration;
+}
+
+export async function backfillMissingMEEDRankings(): Promise<number> {
```
mircorudolph: This backfill is not connected to any production runner, reconciliation route, CronJob, or command, while runRanking catches catalog failures and still returns success. A transient registration failure therefore remains undiscoverable unless the user happens to rerun the same ranking; wire a bounded/paginated invocation into an operational retry path.

app/migrations/20260824120000-create-meed-ranking.cjs
DavidGBbr: Good catch. I updated the durable lookup and unique identity to include user_id consistently. The ranking GET path is also user-scoped now. For anonymous rankings, the database index uses a COALESCE sentinel so NULL users remain idempotent as well.

app/src/backend/meed/MeedNativeInputCatalogService.ts
DavidGBbr: Fixed. Completed MEED rankings are now wired into the existing hiap-catalog-backfill command/manual Kubernetes Job. The MEED backfill uses bounded cursor pagination and persisted checkpoint state. When a page fails, its cursor isn't advanced, so the next operational run retries those records.

app/src/backend/meed/MeedNativeInputCatalogService.ts
DavidGBbr: Fixed. Registration now always runs supersedePreviousVersions after resolving the current catalog entry, including when that entry already existed. This means a retry can repair an earlier partial failure where the replacement was created but the previous active version was not superseded.

app/src/backend/meed/MeedNativeInputCatalogService.ts
```diff
@@ -0,0 +1,321 @@
...
+
+  return { created: record.created.toISOString(), id: record.id };
+}
+
+export async function registerMEEDRanking(
+  rankingId: string,
+): Promise<NativeInputCatalogRegistration> {
+  const ranking = await loadRanking(rankingId);
+  const input = await buildMEEDRankingInput(ranking);
+  const existing = await models().NativeInputCatalog.findOne({
```
mircorudolph: The current deduplication can leave no active catalog entry. After A is superseded by B, rerunning the same inputs and receiving the same output reuses the original A row rather than creating C; B is then superseded in favour of A, but A itself remains marked superseded.
Should every user-initiated ranking create a new run row instead? That would give us a clear history—A, then B, then C—with the newest completed run active, even when C has the same input/content digests as A. The digests could remain useful for comparison or content deduplication, while an explicit run ID or idempotency key could prevent duplicate rows caused by automatic request retries.

Why are we using inventoryId + userId + inputDigest + contentDigest as the identifier when e.g. a the actual runID of the hiap meed ranking run would be the more natural choice?
Maybe worth discussing the approach

app/migrations/20260824120000-create-meed-ranking.cjs
```diff
@@ -0,0 +1,114 @@
...
+        references: { model: "Inventory", key: "inventory_id" },
+        onUpdate: "CASCADE",
+        onDelete: "CASCADE",
+      },
+      user_id: {
+        type: Sequelize.UUID,
+        allowNull: true,
+        references: { model: "User", key: "user_id" },
+        onUpdate: "CASCADE",
+        onDelete: "SET NULL",
```
mircorudolph, line 25: The current per-user identity conflicts with ON DELETE SET NULL: two users can have otherwise identical ranking keys, but deleting them converts both user IDs to NULL. Because the unique index coalesces NULL to one sentinel value, the second deletion can violate the unique constraint and prevent account deletion. 

This raises a broader question about including userId in the ranking identity (see also my comment on linear: https://linear.app/openearth/issue/CC-736/register-completed-hiap-meed-rankings-in-nativeinputcatalog). Before this PR, CC stored and retrieved the latest MEED ranking by inventory only; this implementation changes that behaviour by using inventoryId + userId for persistence, lookup, catalog identity, and supersession. That allows multiple users to have separate active rankings for the same inventory. Should rankings instead belong to the inventory, with userId retained only as createdBy audit metadata? If per-user rankings are intentional, we should confirm that with product. If we just implemented user scope here because it was mentioned in the card, it was likely an error on my side.

app/src/backend/MeedApiService.ts
```diff
@@ -195,27 +220,69 @@ export default class MeedApiService {
...
+    });
+
+    if (!db.sequelize) {
+      throw new createHttpError.InternalServerError(
+        "Database is not initialized",
+      );
+    }
+
+    const data = await db.sequelize.transaction(async (transaction) => {
+      const existingRanking = await db.models.MeedRanking.findOne({
```
mircorudolph: This findOne() followed by create() has a concurrency race: two identical requests can both find no row, after which one creates successfully and the other fails against the unique index. The second caller then receives an error even though its ranking was persisted by the first request.
The fix depends on the intended run model. If identical content should reuse one ranking, this needs an atomic findOrCreate or unique-conflict reload path. If every user-initiated ranking should create a new run row with its own ranking.id, retries should instead be deduplicated using an explicit run ID or idempotency key. That would distinguish a genuinely new run from an automatic retry of the same request (see my other comment)

mircorudolph (changes requested):
Thanks! 

Please see my comments for change requests

app/src/backend/meed/MeedNativeInputCatalogService.ts
DavidGBbr: Done!
Implemented the agreed run model. 

- Each successful HIAP-MEED invocation now creates a new completed MeedRanking row, and that row UUID is used as the NativeInputCatalog sourceId. 
- Digests remain metadata only. 
- Supersession is inventory-scoped, so A -> B -> C (with C matching A) leaves C active and A/B historical. 
- Request-level idempotency is intentionally out of scope because HIAP-MEED currently has no durable runId.

app/migrations/20260824120000-create-meed-ranking.cjs
DavidGBbr: Updated. 
- Rankings are now inventory-scoped: userId is not used in persistence uniqueness, catalog scope, lookup, or supersession. 
- It remains nullable on MeedRanking only as audit/createdBy metadata. 
- The composite user/digest unique index was removed and replaced with an inventory/created index, so ON DELETE SET NULL no longer creates a uniqueness conflict.

app/src/backend/MeedApiService.ts
DavidGBbr: Updated according to the agreed model. Ranking persistence no longer does content-based findOne/create deduplication: every successful invocation creates a new UUID-backed ranking row and persists its child actions in the same transaction. Registration retries for the same persisted ranking remain idempotent through its UUID source identity and the shared catalog uniqueness handling; automatic request-retry idempotency remains a separate follow-up.

app/src/backend/hiap/HiapCatalogBackfillRunner.ts
```diff
@@ -287,8 +329,10 @@ function addPage(totals: BackfillTotals, page: HIAPCatalogBackfillPage): void {
 }
 
 async function processPages(
-  kind: "rankings" | "actionPlans",
-  processPage: HIAPCatalogBackfillRunnerDeps["processRankingsPage"],
+  kind: "rankings" | "actionPlans" | "meedRankings",
```
mircorudolph, line 332: Reusing the terminal completed behavior for MEED leaves a recovery gap. After one successful sweep, the runner persists meedRankings.completed = true. If a new ranking is created later and its immediate catalog registration fails—which runRanking() logs but otherwise treats as successful—every subsequent backfill run returns early because MEED is already marked completed. The missing catalog entry is therefore never repaired.
MEED reconciliation needs to remain ongoing. Preserve the last successfully processed (created, id) cursor and check for newer rankings on every command run instead of permanently setting completed: true. Only advance the cursor after a page succeeds; if registration fails, retain the previous cursor so the next run retries that page. Please add a regression test that completes an initial sweep, creates a new ranking with failed catalog registration, and verifies that a second backfill run repairs it.

Run 1: process A and B → bookmark = B
New ranking C is created
Run 2: process C → bookmark = C
New ranking D fails registration
Run 3: retry D → bookmark moves only after success

app/src/backend/meed/MeedNativeInputCatalogService.ts
```diff
@@ -0,0 +1,333 @@
...
+      availability: { [Op.ne]: "withdrawn" },
+    },
+  });
+  if (existing?.availability === "superseded") {
+    return { catalog: existing, created: false };
+  }
+  const registration = existing
+    ? { catalog: existing, created: false }
+    : await registerNativeInput(input);
+  await supersedePreviousVersions(input, String(registration.catalog.id));
```
mircorudolph: Registration and supersession are separate operations and are not serialized per inventory. If ranking A creates its active catalog row and pauses, ranking B can create its row and supersede A; when A resumes, it can then supersede B because every invocation assumes its own ranking is the newest. This can leave A and B superseding each other with no active entry, or allow an older delayed run to replace a newer one.
Please make registration and supersession atomic per inventory—for example, use one transaction with an inventory-row or transaction-level advisory lock—and select the active version by MeedRanking.created plus id. Add an overlapping-registration test verifying that exactly one entry remains active and that it points to the newest completed ranking.

This is likely an edge case but good to capture this.

app/src/backend/hiap/HiapCatalogBackfillRunner.ts
DavidGBbr: Fixed in 79a71242e. MEED checkpoints are now continuous rather than terminal: a successful page retains its last (created, id) cursor, including the final non-empty page, and MEED is not marked completed in a way that suppresses future scans. If registration fails, the prior cursor is retained, so the next run retries the missed page. Added regression coverage for the A/B initial sweep, failed new ranking C, and successful retry of C.

app/src/backend/meed/MeedNativeInputCatalogService.ts
DavidGBbr: Fixed in f4548e944. MEED registration and supersession now run in one transaction under a transaction-scoped advisory lock per inventory. The transaction loads the persisted sources, chooses the active entry by (MeedRanking.created, MeedRanking.id), and supersedes every other active MEED entry. Added delayed-old and overlapping-registration coverage, verifying that exactly one entry remains active and that it points to the newest ranking.

mircorudolph (approved):

PR #3069, opened 2026-08-26
CC-751: Add stable CNB MLflow interaction names

piotrnowakowski:
## Summary

- add a visible workflow=CNB tag and stable MLflow run names for CNB start, chat, missing-information, and future chat-edit interactions
- instrument the current CNB start and chat boundaries with CNB spans, interaction tags, and concept-note run correlation
- document the four-name contract and reserve cnb_missing_information and cnb_chat_edit for the CC-730 and CC-732 mutation boundaries

## Scope note

This is the interaction naming and correlation slice of CC-751. The missing-information and chat-edit endpoints are not present on develop, so this PR reserves and documents those names without adding their future endpoint implementations.

## Earlier local MLflow output

Captured from real local CC and CA interactions in the live Clima experiment. These historical examples are preserved unchanged. Identifier fields and start/end timestamps have been removed from this example; the remaining statuses, tags, parameters, and metrics are unchanged.

```json
{
  "tracking_uri": "https://mlflow-dev.openearth.dev",
  "experiment_name": "Clima",
  "runs": [
    {
      "status": "FINISHED",
      "tags": {
        "mlflow.runName": "cnb_start",
        "service": "climate-advisor",
        "environment": "local_pn",
        "endpoint": "/v1/concept-notes/start",
        "workflow": "CNB",
        "workflow_name": "concept_note_run_lifecycle",
        "interaction": "start",
        "result": "created"
      },
      "params": {},
      "metrics": {}
    },
    {
      "status": "FINISHED",
      "tags": {
        "mlflow.runName": "cnb_chat",
        "service": "climate-advisor",
        "environment": "local_pn",
        "endpoint": "/v1/messages",
        "workflow": "CNB",
        "workflow_name": "concept_note_context_chat",
        "interaction": "chat",
        "request_kind": "message_stream",
        "trace_category": "ca_agentic_context_chat",
        "context_mode": "concept_note_run",
        "prompt_name": "chat",
        "ca_agentic_flow": "True"
      },
      "params": {
        "content_length": "92",
        "has_context": "False",
        "has_options": "False",
        "has_inventory_id": "False",
        "context_keys": "[]",
        "option_keys": "[]"
      },
      "metrics": {
        "duration_ms": 7732.680752000306,
        "assistant_characters": 408.0,
        "assistant_chunks": 87.0,
        "tool_invocations": 0.0,
        "history_saved": 1.0,
        "ok": 1.0
      }
    },
    {
      "status": "FINISHED",
      "tags": {
        "mlflow.runName": "cnb_missing_information",
        "service": "climate-advisor",
        "environment": "local_pn",
        "endpoint": "/v1/concept-notes/e7cca88d-de54-4050-88ed-d6b39790853b/gaps/9492e67d-3a10-43c8-81a5-8eb388883d25/resolve",
        "workflow": "CNB",
        "workflow_name": "concept_note_missing_information",
        "interaction": "missing_information",
        "result": "regeneration_scheduled"
      },
      "params": {},
      "metrics": {}
    }
  ]
}
```

## Original interaction-naming verification

- 50 focused Climate Advisor tests passed
- Ruff I, F, and SIM117 checks passed for the changed Python files
- changed Python modules compile successfully
- git diff --check passed
- manually verified local CC and CA interactions in the Clima experiment as cnb_start, cnb_chat, and cnb_missing_information

## Linear

https://linear.app/openearth/issue/CC-751/cnb-implement-end-to-end-mlflow-logging-for-agentic-workflows

## Live CNB prompt verification (1 September 2026)

Ran one real OpenRouter model call for every CNB prompt changed by this PR against representative Kraków/EIB prompt bodies. All eight MLflow runs finished, all eight traces are `OK`, and every trace contains a successful root span plus the expected model span:

- [`cnb_chat`](https://mlflow-dev.openearth.dev/#/experiments/4/runs/306ea220bde544c8b9c33fd2be4ef99a) — `openai/gpt-5.6-sol`, reasoning `none`; run `FINISHED`, trace `OK`.
- [`cnb_source_document_mapping`](https://mlflow-dev.openearth.dev/#/experiments/4/runs/92b388df80ae4878b74e50c383daa465) — `openai/gpt-5.6-luna`, reasoning `low`; run `FINISHED`, trace `OK`.
- [`cnb_source_summary_synthesis`](https://mlflow-dev.openearth.dev/#/experiments/4/runs/1cb1f40530844890970cc5a889e43995) — `openai/gpt-5.6-sol`, reasoning `medium`; run `FINISHED`, trace `OK`.
- [`cnb_source_question_reading`](https://mlflow-dev.openearth.dev/#/experiments/4/runs/8fab3a27f16144dfb4c433316d36efa6) — `openai/gpt-5.6-luna`, reasoning `low`; run `FINISHED`, trace `OK`.
- [`cnb_chapter_drafting`](https://mlflow-dev.openearth.dev/#/experiments/4/runs/446190eede2045379959819af4f97c6c) — `openai/gpt-5.6-terra`, reasoning `medium`; run `FINISHED`, trace `OK`.
- [`cnb_funding_opportunity_research`](https://mlflow-dev.openearth.dev/#/experiments/4/runs/31af394c49614181834f966784f2300c) — `openai/gpt-5.6-sol`, reasoning `medium`; run `FINISHED`, trace `OK`.
- [`cnb_funder_identity_matching`](https://mlflow-dev.openearth.dev/#/experiments/4/runs/df412bc10de041d3984da41528c2c56e) — `openai/gpt-5.6-luna`, reasoning `low`; run `FINISHED`, trace `OK`.
- [`cnb_similar_project_matching`](https://mlflow-dev.openearth.dev/#/experiments/4/runs/45131592fbf2439dbc1fa4c2b069269b) — `openai/gpt-5.6-sol`, reasoning `medium`; run `FINISHED`, trace `OK`.

MLflow batch: `cnb-all-prompts-20260831T232103Z` in the `Clima` experiment.

Each run includes `prompt-smoke/input.json` and `prompt-smoke/output.json`. Readback also confirmed non-empty inputs and outputs on every root prompt span and the expected model on every child inference span.

Related regression verification: 221 focused CNB, chat, prompt, model-routing, and MLflow tests passed; scoped Ruff I/F/SIM117 checks, Python compilation, and `git diff --check` passed.

Scope: these are live backend prompt/model smoke tests using saved representative Kraków/EIB data. They do not write application databases and are not browser, authentication, deployment, or production end-to-end tests. MLflow 3.2.0 still emits the known OpenTelemetry `_metrics` compatibility warning after span export; all eight traces and artifacts were nevertheless verified present.


climate-advisor/service/app/routes/concept_note_runs.py
```diff
@@ -72,12 +84,47 @@ async def start_concept_note_run(
...
 ) -> JSONResponse:
     """Create a concept-note run or replay an identical idempotent request."""
-    service = ConceptNoteRunService(session)
-    response = await service.start_run_and_schedule_context(
-        payload,
-        authorization=authorization,
-        context_bundle_service=context_bundle_service,
-    )
+    interaction = CNBInteraction.START
+    with start_mlflow_run(
```
mircorudolph, line 88: This adds another async use of the shared fluent start_run() helper, which already affects chat and Stationary Energy handlers: MLflow 3.2 keeps active runs in thread-local rather than task-local state, while these contexts remain open across await. Under overlapping requests, the helper swallows the second start failure and later log_tags() calls can write its CNB correlation fields onto another request’s active run; please fix the shared helper to use explicit per-request run IDs/client logging and cover concurrent requests.

This seems to be an issue of the initial mlflow implementation done by me that now surfaced here. Wuould be a good timing to fix those.

mircorudolph (commented):
When looking at the mlflow traces for Clima in CNB, it seems we do not have a dedicated prompt for CNB or at least the normal chat interface. Instead we are using the standard Clima prompt: https://mlflow-dev.openearth.dev/#/experiments/4/runs/0dcfeb54cdff4534962324e984f8bde7

> <additional_instructions> <role> You are Clima handling general CityCatalyst climate and inventory chat. </role>
<task> Answer general climate, emissions, inventory, and CityCatalyst workflow questions. Use tools when....

We should likely add that CNB specific part similar to stationary_energy etc.

Also e.g. the 2nd system message about the document seems to contain a lot of fields not relevant to the model. Can we do some more context engineering here? (Unless really all of these fields are required).

And also why are we adding this very long document summary to the context? What happens if we upload 5 documents? Would we fill the context with all 5 long summaries? Wouldn't it be better to just have a very short index/ summary and use a tool for on demand discovery?

Also having this as a 2nd system message is likely not ideal for this type of data. System messages are typically reserved/ used for model guidance, guardrails and instructions and not for providing runtime data.

climate-advisor/service/app/routes/concept_note_runs.py
piotrnowakowski: Thanks for flagging this. I replaced the shared fluent active-run handling with explicit per-request run IDs and `MlflowClient` logging, with the active target held in task-local context.

I added coverage for overlapping requests, failed starts, nested runs and cancellation. I also ran concurrent CNB source-query requests against MLflow; each kept its own run, trace, thread and CNB correlation fields, with no cross-request logging.

piotrnowakowski:
@mircorudolph Thanks, these were good points. I added a dedicated CNB prompt composed with the shared Clima instructions, following the Stationary Energy pattern.

I also reduced the runtime context: internal IDs, UUIDs, hashes, generated references and page/block counts no longer go to the model. The context bundle and retained tool history are now passed as runtime data rather than system messages, so the system role contains only the behavioral instructions.

For detailed document questions, the chat now uses `concept_note_sources_query`. It selects one document by its readable label and filename, reads the complete document in bounded partitions, and returns verified excerpts with readable citations.

The one part still open is the always-loaded document summary. It is bounded, but five documents would still add five summaries. I agree that this should become a much shorter document index with an overall context budget, leaving detailed discovery to the tool. I wouldn’t consider that part resolved until the shorter index is in place.

I also ran all eight edited CNB prompts with live models and added the successful MLflow runs to the PR description.

climate-advisor/service/app/persistence/concept_notes/context_bundle.py
```diff
@@ -406,21 +408,25 @@ async def load_query_source(
...
+                if item.source_label == source_label and item.filename == filename
+            ]
+            if not matches:
                 raise ContextBundlePersistenceError(
                     "concept_note_source_not_selected",
                     404,
                     "Selected Concept Note source was not found",
                 )
-            upload = await session.get(ConceptNoteUpload, upload_id)
+            if len(matches) != 1:
```
mircorudolph: This makes both documents unqueryable when two accepted uploads share the same source_label and filename. The model only receives those fields—not the unique upload_id—so every tool call matches both documents and returns concept_note_source_ambiguous, with no way for the model to recover. Please expose a model-safe unique selector, such as a per-context source index mapped back to upload_id in application code, or enforce uniqueness before the bundle becomes ready.

This regression is introduced by replacing the unique upload_id selector with source_label plus filename. Two accepted uploads may share that pair, after which every source-query call matches both and returns concept_note_source_ambiguous; the model cannot recover because it no longer receives a unique selector. Please expose a model-safe unique index mapped back to upload_id, or enforce uniqueness before the bundle becomes ready.

This can happen when a user uploads twice:

source_label: Climate Action Plan
filename: climate-action-plan.pdf

which seems like it could happen.

mircorudolph (approved):

climate-advisor/service/app/persistence/concept_notes/context_bundle.py
piotrnowakowski: Thanks, good catch. Fixed in `8ee9c8320` by adding a one-based, per-context `source_index` to each model-facing source. `concept_note_sources_query` now accepts that index, and the server maps it back to the persisted `upload_id` within the authorized run.

This keeps internal IDs and fingerprints out of the model context while allowing documents with identical `source_label` and `filename` values to be queried independently.

I added regression coverage with two ready uploads sharing the same name: the model receives indexes `1` and `2`, each resolves to its distinct upload, and an out-of-range index is rejected. The prompt and architecture documentation were updated as well. The broader 160-test suite and all PR checks are green.

PR #3073, opened 2026-08-27
CC-582: Re-enable login e2e test suite

cephaschapa:
## Summary

Re-enables the skipped Playwright login happy-path test and stabilizes form submit so client handlers run before the button is clickable.

## Changes

- Unskipped login redirect test; assert navigation away from `/auth/login`
- Gate login submit with `data-ready` until after mount (avoids native GET with credentials in the query string)
- Post `/api/v1/auth/register/` with trailing slash in the e2e signup helper
- Added missing `log-in` string to `en/auth.json`

## How to test

1. From `app/`, run `npx playwright test e2e/login.spec.ts --project=chromium` (CityCatalyst on the Playwright base URL).
2. Confirm both Login page tests pass.

## Ticket

CC-582

### DIFF
--- a/app/e2e/helpers.ts
+++ b/app/e2e/helpers.ts
@@ -106,7 +106,7 @@ export async function signup(
   name: string = "Test Account",
   acceptTerms: boolean = true,
 ) {
-  const result = await request.post("/api/v1/auth/register", {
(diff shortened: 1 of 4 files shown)


lemilonkh (approved):
Great fix for the ready state, thanks 👍🏻✨

PR #3075, opened 2026-08-27
CC-742: Preserve module when switching cities via sidebar

tatimoreira:
## Summary

Switching cities (or organizations) from the sidebar always redirected to the Journey Navigator home, dropping the user out of whatever module (GHGI, HIAP, Dashboard) they were working in. This preserves the current module for the newly selected city/organization when it's available, and falls back gracefully with an informational toast when it isn't.

## Changes

- Add `resolveCitySwitchPath` / `getActiveModuleSegment` in `app/src/util/module-navigation.ts` as the single source of truth for "what module is the user in" and "where should a city switch land".
- `JNDrawer.tsx`: city dropdown, city search results, and organization switcher now preserve the active module for the new city/org (checked against that project's enabled modules), instead of hardcoding `/cities/{cityId}`.
- `navigation-bar.tsx`: module-name display logic now reuses the shared helper (removing duplicated pathname parsing); organization switcher gets the same fix.
- Add `module-unavailable-for-city` i18n key (`en/dashboard.json`, `en/navigation.json`) for the fallback toast.
- Add unit tests for the new util (`app/tests/module-navigation.jest.ts`).

## How to test

1. Open a city's GHGI dashboard (`/en/cities/{cityA}/GHGI/{inventoryId}`), open the sidebar, and switch to another city that also has GHGI enabled — confirm you land on that city's GHGI (latest inventory or onboarding), not the Journey Navigator.
2. Repeat for HIAP and for `dashboard`.
3. Switch to a city whose project does not have GHGI/HIAP enabled while on that module — confirm redirect to the Journey Navigator with an info toast naming the module.
4. Switch cities from the search-result list, and switch organizations from a module page — confirm the same preserve/fallback behavior.
5. `npm run jest -- module-navigation` — unit tests pass.

## Ticket

CC-742


app/src/components/HomePage/JNDrawer.tsx
```diff
@@ -499,12 +535,43 @@ const JNDrawer = ({
...
+      );
+      toaster.create({
+        type: "info",
+        title: t("module-unavailable-for-city", {
+          module: blockedModule?.name?.[lng] || blockedModule?.name?.en || "",
+        }),
+      });
+    }
+
+    router.push(path);
```
lemilonkh: We have this same logic repeated in 3 places in the code base. Could it be extracted into a React hook so it can be reused more easily perhaps?

app/src/components/navigation-bar.tsx
```diff
@@ -119,11 +125,17 @@ export function NavigationBar({
...
+      case "GHGI":
+        return t("page-title-ghg-inventories");
+      case "HIAP":
+        return t("page-title-hiap");
+      case "dashboard":
+        return t("page-title-dashboard");
+      default:
+        return null;
+    }
   }, [pathname, t]);
```
lemilonkh: This could be tricky to maintain going forward when we're adding more and more modules. Can the respective name come from the Modules table (where we have different translations of it as well)?
Might require adding the route segment to the table and then adding it as data to the module seeder.

app/src/util/module-navigation.ts
```diff
@@ -0,0 +1,57 @@
...
+export type ModuleRouteSegment = {
+  segment: "GHGI" | "HIAP" | "concept-notes" | "dashboard";
+  moduleId?: string; // undefined => ungated route
+};
+
+const MODULE_ROUTES: ModuleRouteSegment[] = [
+  { segment: "GHGI", moduleId: Modules.GHGI.id },
+  { segment: "HIAP", moduleId: Modules.HIAP.id },
+  { segment: "concept-notes", moduleId: Modules.CONCEPT_NOTE_BUILDER.id },
+  { segment: "dashboard" },
```
lemilonkh: This is also additional maintenance effort required when adding new modules. Can this data come from the Modules table?

lemilonkh: E.g. we just added the native MEED module. Ideally it would just work with this system directly since it's in the Modules seeder.

lemilonkh (approved):
Great work, thanks! ✨
Would just be good to make sure to not add a bunch of new places where added modules need to be configured before they work correctly with this feature.

app/src/components/HomePage/JNDrawer.tsx
tatimoreira: I'll extract it into a hook

app/src/util/module-navigation.ts
tatimoreira: Good call — fixed in fbd648bce. `getActiveModuleSegment`/`resolveCitySwitchPath` now derive `{segment, moduleId, name}` from the live Modules table (`useGetModulesQuery`) using each module's `url` field, instead of a hardcoded map. No migration/seeder change was needed — `url` (e.g. `/GHGI`, `/MEED`) already served as the route segment and `name` was already localized. New modules should work here automatically as long as their `url` is an internal app route.

app/src/components/navigation-bar.tsx
tatimoreira: fixed

PR #3077, opened 2026-08-28
CC-655: Redirect multi-project users to All Projects view on login

tatimoreira:
## Summary

Users with access to multiple projects previously always landed on the Journey Navigator (`/cities`) after login, requiring extra navigation to see their other projects. They now land on the existing "All Projects" view instead. Single/zero-project users keep the existing behavior.

## Changes

- `app/[lng]/page.tsx`: fetch the user's projects alongside user info; redirect to `/organization/{orgId}/project` when the user has more than one project, otherwise `/cities/` as before. Also stopped rendering the deprecated `GHGIHomePage` fallback except on error, since it has its own onboarding-redirect effects that raced the new redirect.
- `proxy.ts` (Next 16's `middleware.ts` equivalent): removed an unconditional server-side redirect that sent every request to `/{lng}` or `/{lng}/` straight to `/{lng}/cities/`, which made the landing-page redirect logic dead code before this change.

## How to test

1. Log in as a user who is an OrganizationAdmin (or otherwise has access to) more than one project — confirm you land on `/{lng}/organization/{orgId}/project` ("All Projects").
2. Log in as a user with a single project (or none) — confirm you still land on `/{lng}/cities/` as before.

## Ticket

CC-655

### DIFF
--- a/app/src/app/[lng]/page.tsx
+++ b/app/src/app/[lng]/page.tsx
@@ -8,7 +8,6 @@ import { useRouter } from "next/navigation";
 import { api } from "@/services/api";
 import ProgressLoader from "@/components/ProgressLoader";
 
-/** @deprecated has been replaced with GHGI module home page */
 export default function PrivateHome(props: {
   params: Promise<{ lng: string }>;
 }) {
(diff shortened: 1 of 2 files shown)


lemilonkh (approved):
Great, thanks 👍 ✨

PR #3078, opened 2026-08-31
CC-755: Remove duplicate seeded module rows

cephaschapa:
## Summary

Fixes duplicated module toggles in Admin → Projects → Modules by removing extra `Module` rows that match the seeded catalog (`modules.json`) by English name or URL but use a different UUID. User-created modules that are not in the seed catalog are left unchanged.

## Changes

- Add migration `20260831120000-remove-duplicate-seeded-modules.cjs`
- Repoint `ProjectModules` to canonical seeded ids, then delete duplicate `Module` rows

## How to test

1. Insert a duplicate module row (same `name.en` or `url` as a seeded module, different `id`).
2. Run `npm run db:migrate` in `app/`.
3. Confirm only the seeded id remains and Admin → Modules shows one row per seeded module.
4. Confirm custom modules (e.g. GHG Inventory Comparison) that are not in `modules.json` are untouched.

## Ticket

CC-755

## Notes

Root cause: admin module creation inserts new UUIDs while the seeder upserts fixed ids, producing duplicate catalog entries in dev/prod.


lemilonkh (approved):
Great fix, thanks 🙏🏻✨

PR #3081, opened 2026-08-31
[1/2] CC-605: Add chapter validation in Climate Advisor

piotrnowakowski:
## Layer scope

This existing PR is now the bottom Climate Advisor layer of a two-PR stack. It contains only `climate-advisor/**`: validation models, persistence, migration, prompts, service orchestration, route, configuration, documentation, and their tests. The CityCatalyst proxy and UI moved to #3090; the original combined-change context below is retained for review history.

## Problem

Concept Note Builder needed a reliable review-before-export workflow. Real source-backed runs also exposed connected resilience and usability problems: uploaded PDF metadata disappeared after navigation, polling could exhaust the shared API rate limit, transient refresh failures looked like missing runs, missing-information prompts were repeated across the workspace, and export showed the same missing facts as both prompts and validation blockers.

## What this PR changes

- Adds persisted, two-pass chapter validation for completeness, evidence, internal logic, and cross-chapter consistency, with bounded concurrency, latest-result persistence, stale-result detection, retry behavior, and a dedicated CNB migration.
- Adds the guided **Review & export** journey: missing information, conflicts and logic, export decision, explicit export-anyway confirmation, and DOCX/PDF selection.
- Preserves uploaded-source metadata in run responses, supports the real 19.4 MiB Richfield PDF through the Next.js proxy, and restores upload details after refresh or resume.
- Reduces workspace polling pressure, stops polling terminal uploads, and keeps the last valid run visible through transient API failures.
- Simplifies the draft workspace with inline missing-information popovers, a collapsible chapter panel that fully returns document width, and removal of duplicated chapter summaries and framing.
- Unifies unanswered prompts and validation blockers into one **Missing information** count and one fix action while keeping warnings separate.
- Routes active Climate Advisor generation through GPT-5.6 Terra or Sol and updates model-contract tests and documentation.
- Keeps the CodeQL exception-exposure fix while preserving stable machine-readable validation error codes.

## Resulting behavior

Users can upload evidence, leave and reopen a run, draft every template chapter, inspect missing facts directly in the text, collapse the chapter navigation without leaving an empty rail, run or reuse saved validation, and make one clear export decision. The export screen presents one missing-information count instead of reporting the same items twice.

## Migration

Run the CNB migration chain through revision `20260828_120000` before using chapter validation.

https://github.com/user-attachments/assets/472b9836-40cd-4dac-8999-0d9c91c04557

## Stack

1. **#3081 - Climate Advisor chapter validation** (this PR)
2. #3090 - CityCatalyst guided review and export

### DIFF
--- a/climate-advisor/llm_config.yaml
+++ b/climate-advisor/llm_config.yaml
@@ -26,6 +26,9 @@ models:
   cnb_chapter_drafter:
     name: "openai/gpt-5.6-terra"
     reasoning_effort: "medium"
+  cnb_chapter_validator:
+    name: "openai/gpt-5.6-terra"
+    reasoning_effort: "medium"
 
 # Generation Parameters
 generation:
@@ -44,6 +47,8 @@ generation:
       max_key_excerpts: 8
       max_topics: 12
       max_question_chars: 2000
+    cnb_validation:
+      max_prompt_tokens: 50000
 
 # System Prompts
 prompts:
@@ -72,6 +77,8 @@ prompts:
   cnb_source_summary_synthesis: "prompts/cnb/source_summary_synthesis.md"
   cnb_source_question_reading: "prompts/cnb/source_question_reading.md"
   cnb_chapter_drafting: "prompts/cnb/chapter_drafting.md"
+  cnb_chapter_validation_completeness: "prompts/cnb/chapter_validation_completeness.md"
+  cnb_chapter_validation_consistency: "prompts/cnb/chapter_validation_consistency.md"
 
 # API Configuration
 api:
--- a/climate-advisor/llm_config.yaml
+++ b/climate-advisor/llm_config.yaml
@@ -5,12 +5,12 @@
 models:
   orchestrator:
     name: "openai/gpt-5.6-luna"
-    reasoning_effort: "none"
+    reasoning_effort: "medium"
     description: "Default chat model for Climate Advisor conversations."
   agentic_flow:
     name: "openai/gpt-5.6-sol"
     description: "Agentic flow model for high-context review and source selection workflows."
-    reasoning_effort: "none"
+    reasoning_effort: "medium"
   funding_research:
     name: "openai/gpt-5.6-sol"
     reasoning_effort: "medium"
--- a/climate-advisor/prompts/cnb/chapter_validation_completeness.md
+++ b/climate-advisor/prompts/cnb/chapter_validation_completeness.md
@@ -1,0 +1,80 @@
+<role>
+You are the completeness validator for one CityCatalyst Concept Note Builder
+chapter. You perform the first of two validation passes. You are not a chat
+assistant and you do not rewrite the chapter.
+</role>
+
+<task>
+Review only the supplied target chapter for missing required information,
+application-template violations, unresolved content gaps, and evidence quality.
+
+Rules:
+- evaluate the chapter against its `required` flag, matching
+  `template.chapter_schema` entry, `template.required_fields`, and output format
+- apply only template requirements that are relevant to the target chapter;
+  do not flag a required field that belongs to another chapter
+- when `template` is null, do not invent template constraints
+- identify information that is explicitly required or necessary for the
+  chapter's own claims to be complete; do not invent unstated requirements
+- treat `[Information needed: ...]` markers and relevant `open_gaps` as
+  unresolved, but do not duplicate an open gap unless the chapter text itself
+  demonstrates the same omission; when it does, repeat the exact gap `reason`
+  in the finding message or suggested action so the service can deduplicate it
+- assess whether material factual claims have usable support in
+  `evidence_links`; a link label alone is not proof of claim support, but do
+  not warn merely because the array is empty when the chapter has no material
+  factual claim that requires evidence
+- treat named-project facts, route or location descriptions, quantities,
+  dates, costs, progress, financing, attributed statements, and predicted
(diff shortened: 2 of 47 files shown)


mircorudolph:
image

Not part of this PR specifically but we should not expect a user to add such a specific prompt:
- a user does not know about `run context` 
- a user would not know to say "use only this"
- walk through the template chapters in order... -> For the user it is 1 workflow that the user should not instruct on how to walk through this.

we must assume 0 knowledge of the user how to talk best to that chat. 

Let's review the prompt and make sure we are not testing only the happy path where the user makes the perfect prompt.
Instead even bad prompts should receive guidance and lead to the target

mircorudolph:
image

Refers to ticket: CC-769 source provenance. Those are likely good examples to tell the user where this wrong date appeared etc. (same as for other extracted data and figures)

climate-advisor/prompts/cnb/chapter_validation_completeness.md
```diff
@@ -0,0 +1,80 @@
...
+</role>
+
+<task>
+Review only the supplied target chapter for missing required information,
+application-template violations, unresolved content gaps, and evidence quality.
+
+Rules:
+- evaluate the chapter against its `required` flag, matching
+  `template.chapter_schema` entry, `template.required_fields`, and output format
+- apply only template requirements that are relevant to the target chapter;
```
mircorudolph: This works, but we should deterministically select the matching chapter_schema entry and its associated required fields before calling the LLM. Filtering everything we can in code reduces context size, simplifies the prompt, and lowers the risk of inconsistent results

climate-advisor/service/app/services/cnb/chapter_validation.py
```diff
@@ -0,0 +1,824 @@
...
+)
+_INTERNAL_SCOPE_CONFLICT_MESSAGE = (
+    "The target defines the proposed measure as delivery of current works while "
+    "also stating that construction or commissioning is already at a late stage."
+)
+_INTERNAL_SCOPE_CONFLICT_ACTION = (
+    "Replace current works with a distinct future residual or follow-on scope, "
+    "schedule, and cost."
+)
+_SCOPE_PATTERNS = {
```
mircorudolph: Why does the generic chapter validator contain a hard-coded pattern for EUCF support and this specific construction/commissioning scenario? EUCF otherwise appears only in the tracked reference bundle and these tests, while this check runs for every funding template. If this is an authoritative EUCF rule, could it be scoped through the relevant template or structured eligibility data? Otherwise, it seems like an example-specific rule that could cause false blocking findings for unrelated programmes

climate-advisor/llm_config.yaml
mircorudolph: What is the reason that we set this to `none`? Wouldn't at least `low` be somewhat better since this agent is also doing tool calling?

mircorudolph (approved):

mircorudolph:
Thanks Piotr! Good work. Just left some comments, but none of them blocking

piotrnowakowski:
Agreed. I updated the CNB chat prompt so users no longer need to understand run context, internal workflow stages, or chapter ordering. Short requests such as “Help me” or “What should I do next?” now use the active run context automatically and guide the user to the next useful step. I also added behavioral tests covering short and vague requests.

piotrnowakowski:
Good idea. I added trusted source provenance to validation findings, including the source label, location, claim reference, and supporting excerpt. I also added a way to surface this information to users in the stacked frontend PR #3090. I’ll add this to CC-769 as well.

climate-advisor/service/app/routes/concept_note_chapter_validation.py
```diff
@@ -0,0 +1,84 @@
...
+        chapter_id=chapter_id,
+        **validation.model_dump(),
+    )
+
+
+def _problem(status_code: int, code: str, message: str) -> JSONResponse:
+    """Return one stable machine-readable validation failure."""
+    return JSONResponse(
+        status_code=status_code,
+        content={"code": code, "detail": message, "status": status_code},
```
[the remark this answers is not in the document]
piotrnowakowski: Fixed. The route now returns a stable public error message instead of exposing the underlying exception text.

climate-advisor/service/app/services/cnb/chapter_validation.py
piotrnowakowski: Agreed. I removed the programme-specific construction and commissioning patterns from the generic validator. The consistency pass now compares the generated output with the supplied document and reports only explicit contradictions found in that input. It no longer infers EUCF or other programme rules from names or familiar phrases. I also added a regression test proving that this scenario does not create a finding without an explicit rule in the supplied document.

climate-advisor/llm_config.yaml
piotrnowakowski: Good point. I updated both `agentic_flow` and `orchestrator` to use `reasoning_effort: medium`, since both can participate in tool-calling chat flows. I also updated the request-payload tests and documentation, and CI is green.

climate-advisor/prompts/cnb/chapter_validation_completeness.md
piotrnowakowski: Addressed in f0b9a963a. Chapter validation now normalizes and deterministically selects the matching chapter schema before the LLM call, sends only that chapter and its chapter-specific required fields in document.validation_profile, and rejects ambiguous or unassigned template mappings before invoking the model. The offline research contract, focused tests, and architecture documentation now preserve that per-chapter ownership.

PR #3083, opened 2026-08-31
fix(meed): register Actions & Plans v2 via migration so it reaches deployed environments

carlosgraffi:
Actions & Plans v2 does not appear in the admin panel on dev, even though #2956 merged and deployed successfully on 31 Aug.

`GET /api/v1/modules/` on dev returns 14 modules and none is ours, and the admin route is an unfiltered `Module.findAll()` — so the row simply is not there.

## Why

The module relied on the seeder, after I removed its registration migration during review of #2956. That change was based on establishing that deploys do run `db:seed`, which they do. The gap is that running and succeeding are not the same thing:

```
run_migrations ...                              <- output checked
"Stop deployment if migration failed" exit 1    <- deploy halts
...
kubectl create -f k8s/cc-seed.yml               <- Job created, never inspected
kubectl apply -f k8s/cc-web-deploy.yml          <- proceeds regardless
```

Migrations gate the deploy. Seeds are fire-and-forget: `kubectl create` returns as soon as the Job object exists, and `db:seed:all` runs seeders in order, so a failure in any of the four preceding the modules seeder ends the run before it is reached. Neither case surfaces anywhere.

## What this does

Mirrors `20260824070000-add-concept-note-builder-module.cjs` — the module added a week earlier, which *is* present on dev and arrived exactly this way. Upserts the `Module` row and grants it to existing projects.

That second part matters independently: the seeder never touches `ProjectModules`, so even with the row present the module could not be enabled for a project.

The seed entry stays as the source of truth for fresh databases; this is the guaranteed path for existing ones.

## Verified

Locally, against a real database: deleted the row and its grants, ran `db:migrate`, and both returned (row plus 2 project grants). `db:migrate:undo` removes them; re-applying restores them.

## Note for @lemilonkh

This re-adds a migration you asked me to remove in #2956, and your reasoning was right as far as it went — seeders do run on every deploy. Dev is the counter-example for the stronger claim that they are therefore guaranteed.

Worth considering separately: gating the deploy on the seed Job would fix this class of problem for every module rather than one at a time. I did not do it here because if a seeder is currently failing, adding that gate would start failing everyone's deploys — which is correct behaviour but not something to spring on other teams inside a hotfix.

### DIFF
--- a/app/migrations/20260831180000-register-meed-module.cjs
+++ b/app/migrations/20260831180000-register-meed-module.cjs
@@ -1,0 +1,118 @@
+"use strict";
+
+const MODULE_ID = "9f622243-fba8-4f32-a000-ce6e66982bd1";
+
+/**
+ * Registers the Actions & Plans v2 module.
+ *
+ * This duplicates `seed-data/modules/modules.json`, which is not ideal — but
+ * seeding does not guarantee the row reaches a deployed environment. The deploy
+ * runs migrations behind a check that halts on failure, then creates the seed
+ * Job with `kubectl create` and moves on without ever inspecting its outcome.
+ * A seeder that fails, or that never runs because an earlier seeder in the
+ * chain threw, is silent.
+ *
+ * That is not hypothetical: after #2956 merged and deployed successfully, this
+ * module was absent from dev's Module table while every seeded module around it
+ * was present. Concept Note Builder, added a week earlier, is present — and it
+ * arrived through exactly this pattern (20260824070000).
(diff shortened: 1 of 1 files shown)


lemilonkh (changes requested):
This data is already present in the module seeder. No need for a migration.

lemilonkh:
Fixed by disabling the formula inputs seeder, which was handled by a sync script anyways and was holding up the seeder pods on the dev cluster.

PR #3085, opened 2026-08-31
CC-737: Connect NativeInputCatalog to Climate Advisor

DavidGBbr:
## Summary

Connect CityCatalyst Core NativeInputCatalog capabilities to Climate Advisor for authorized discovery and bounded, source-specific reads through the existing Core boundary.

## Changes

- Add authorized Core catalog discovery and selected-read capability routes with lifecycle checks, bounded execution, and stable non-disclosing errors.
- Add Climate Advisor request-time catalog discovery and selected-only capability/tool loading without storage credentials or raw storage access.
- Add cross-service contract, authorization, non-disclosure, boundedness, fault-isolation, and compatibility coverage.
- Record the AI-DLC inception, construction, verification, and approval evidence for CC-737.

## How to test

1. From `app/`, run the focused Core NativeInputCatalog Jest suites.
2. From `climate-advisor/`, run the focused Climate Advisor regression suite.
3. Run the configured live CC–CA contract suite against the local Core at `http://localhost:3000`.

## Ticket

[CC-737 — Connect NativeInputCatalog to Climate Advisor capabilities](https://linear.app/openearth/issue/CC-737/connect-nativeinputcatalog-to-climate-advisor-capabilities)

## Notes

- Focused verification: Core 26/26, Climate Advisor 109/109, and live CC–CA contracts 4/4 passed.
- Full Core Jest was limited by local PostgreSQL connectivity failures and one unrelated OpenAPI version mismatch; Playwright’s configured web server could not start.
- Full Climate Advisor collection was limited by repository import-path/integration environment issues. These are recorded as environment or baseline limitations, not CC-737 implementation failures.
- Core authorization, bounded reads, selected-only loading, non-disclosure, storage isolation, and the absence of Climate Advisor storage credentials remain enforced.


aidlc-docs/construction/plans/uow-01-core-catalog-capability-boundary-code-generation-plan.md
lemilonkh, line 1: Can we remove the aidlc-docs folder from the pull request?
Seems like it's mostly documents that were used in the implementation process.
Any important documentation files can be moved into our `docs/` folder or a sub-folder of it.

app/src/app/api/v1/internal/ca/capabilities/native-inputs/discover/route.ts
```diff
@@ -0,0 +1,83 @@
...
+const discoveryRequestSchema = z
+  .object({
+    user_id: z.string().uuid().optional(),
+    organization_id: z.string().uuid().optional(),
+    project_id: z.string().uuid().optional(),
+    city_id: z.string().uuid().optional(),
+    inventory_id: z.string().uuid().optional(),
+    kind: z.string().trim().min(1).max(64).optional(),
+    owning_module: z.string().trim().min(1).max(64).optional(),
+    capability_id: z.string().trim().min(1).max(128).optional(),
```
lemilonkh: Our API routes usually accept data in camelCase capitalization (for consistency).

DavidGBbr: Applied!

aidlc-docs/construction/plans/uow-01-core-catalog-capability-boundary-code-generation-plan.md
DavidGBbr: fixed!

climate-advisor/service/app/services/agent_service.py
```diff
@@ -232,6 +261,43 @@ def _build_stationary_energy_start_draft_tools(self) -> Sequence[object]:
             token_ref=self._token_ref,
         )
 
+    async def _build_native_input_catalog_tools(self) -> Sequence[object]:
+        """Discover and compose only the caller's current selected capability."""
+        service = self.native_input_catalog_service
+        context = self.native_input_catalog_context
+        requested_selection = self.native_input_selection
```
mircorudolph: Could we change this from request-preselected tool registration to runtime catalog discovery?
Our intention is that the Climate Advisor agent can discover the NativeInputCatalog entries currently available to the authenticated user. The available entries may change during an existing chat—for example, when a CNB document is uploaded and processed after the thread was created.
Currently, _build_native_input_catalog_tools() requires native_input_selection before discovery runs. The discovery result is then used only to validate that preselected pair; it is never made available to the model. This means something outside the agent must already know and select the catalog entry, and the agent cannot discover newly available entries itself.
The desired workflow is:

Agent is created with stable tools:
  native_input_discover
  native_input_read

User asks about available information
        ↓
native_input_discover queries Core at that moment
        ↓
Core returns the user’s currently authorized catalog entries
        ↓
Agent chooses an entry
        ↓
native_input_read revalidates and reads that entry through Core

“Stable tools” means that the tool definitions remain fixed while their results are dynamic. native_input_discover should query Core when invoked instead of relying on a catalog snapshot captured when the agent was created. native_input_read should accept a pair returned by discovery and have Core revalidate authorization, lifecycle, scope, and readiness on every read.
With this design, a document added during a chat becomes visible on the next discovery call without requiring the client to construct and resend native_input_selection or requiring the agent’s tool set to be rebuilt. For CNB documents specifically, they would also need to be registered in NativeInputCatalog with an appropriate bounded document capability.

mircorudolph (commented):
Left a comment regarding the workflow of NativeInput discovery procedure.

climate-advisor/service/app/services/agent_service.py
DavidGBbr: Good Idea!
Implemented this runtime-discovery direction in the follow-up commits.
Climate Advisor now registers two stable tools for an authenticated catalog context:
- native_input_discover
- native_input_read

native_input_discover queries Core when the agent invokes it, rather than using a catalog snapshot or a client-provided native_input_selection during agent construction. This allows newly authorized catalog entries to become visible during an existing chat without rebuilding the agent tool set.

native_input_read accepts the catalogId and capabilityId returned by discovery. Every read goes back through Core, which independently revalidates identity, scope, catalog lifecycle, capability membership, readiness, and the bounded execution contract before returning data.
The model-facing read schema stays finite, and Climate Advisor does not receive storage credentials, signed URLs, storage pointers, or arbitrary source input.
CNB registration and a bounded CNB document capability remain separate follow-up work, as you noted.

app/src/backend/NativeInputCatalogCapabilityService.ts
```diff
@@ -0,0 +1,422 @@
...
+        "projectId",
+        "organizationId",
+        "availability",
+        "labels",
+      ],
+      order: [
+        ["created", "ASC"],
+        ["id", "ASC"],
+      ],
+      limit: DISCOVERY_RESULT_LIMIT,
```
mircorudolph, line 97: Discovery applies the 100-row limit before authorization and compatibility filtering, so authorized entries can disappear behind older inaccessible rows. Even if the limit is moved after filtering, users with more than 100 authorized entries would still lose access to undiscovered results because native_input_discover has no pagination. Please implement deterministic cursor-based pagination through all active candidates, returning up to 100 authorized compatible entries per page plus a continuation cursor, and add regression coverage proving that entries beyond both inaccessible rows and the first authorized page remain discoverable and readable

app/src/backend/NativeInputCatalogCapabilityService.ts
```diff
@@ -0,0 +1,619 @@
...
+    ...(request.capabilityId ? { capabilityId: request.capabilityId } : {}),
+  };
+}
+
+function discoveryCursorSecret(): Buffer {
+  const secret =
+    process.env.NEXTAUTH_SECRET ||
+    process.env.CC_SERVICE_API_KEY ||
+    "cc-native-input-discovery-cursor";
+  return createHash("sha256").update(secret).digest();
```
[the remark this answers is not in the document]
DavidGBbr: This is key derivation for the opaque AES-GCM discovery cursor, not password storage.

`26ffc71b0` removes `CC_SERVICE_API_KEY` from the IKM chain and derives the 32-byte key with HKDF-SHA-256 from `NATIVE_INPUT_DISCOVERY_CURSOR_SECRET` or `NEXTAUTH_SECRET`.

app/src/backend/NativeInputCatalogCapabilityService.ts
DavidGBbr: Implemented deterministic cursor pagination after authorization and compatibility filtering.

Core now scans active candidates in created ASC, id ASC keyset order and only counts authorized, ready, compatible entries toward the page of 100. Each request is bounded to 200 raw candidates. If more remain, native_input_discover returns an opaque continuationCursor (AES-GCM, bound to the caller and filters) even when that page is empty. Climate Advisor relays the cursor; native_input_read is unchanged and still revalidated by Core.

Regressions cover a later authorized row behind inaccessible prefix rows, a second authorized page, empty-page continuation, and filter-mismatched/malformed cursor rejection.

mircorudolph (approved):

PR #3090, opened 2026-09-01
[2/2] CC-605: Add guided review and export in CityCatalyst

piotrnowakowski:
## Problem

The CityCatalyst web app needs to expose Climate Advisor's chapter validation as a clear review-before-export workflow without keeping the entire implementation in a few oversized workspace components.

## What this layer changes

- Adds the CityCatalyst proxy route and RTK Query contract for per-chapter validation.
- Adds the guided **Review & export** flow for missing information, conflicts and logic, export decisions, export-anyway confirmation, and DOCX/PDF selection.
- Splits the workspace into focused document, setup, navigation, review, summary, state, and data/focus hooks so `export-dialog.tsx`, `draft-tab.tsx`, and the workspace root stay smaller.
- Preserves uploaded-source metadata after refresh/resume and supports the configured larger PDF proxy limit.
- Reduces polling pressure, stops terminal upload polling, and retains the last valid run through transient refresh failures.
- Consolidates missing-information prompts and validation blockers into one count and one fix path while keeping warnings separate.
- Updates dashboard status helpers, API contracts, tests, and the end-to-end architecture document.

## Dependency and review order

This is the CityCatalyst layer and depends on the Climate Advisor validation API in #3081. Review and merge #3081 first, then this PR.

## Validation

- 63 focused Jest tests passed across 9 suites.
- TypeScript `tsc --noEmit` passed.
- ESLint passed for every changed TypeScript/TSX file.
- The stack preservation check confirms the combined top reproduces all 82 paths from the original PR exactly.

## Stack

1. #3081 - Climate Advisor chapter validation
2. **#3090 - CityCatalyst guided review and export** (this PR)


app/src/components/ConceptNoteDashboard/run-card.tsx
```diff
@@ -52,10 +51,11 @@ export function RunCard({
   resumeHref,
   run,
   scopeLabel,
-  statusLabel,
-  statusTone,
   t,
 }: RunCardProps) {
+  const { currentData: draft } = api.useGetConceptNoteDraftQuery(run.run_id);
```
cephaschapa: Each run card always calls `useGetConceptNoteDraftQuery(run.run_id)`, so the dashboard fetches every full draft (all chapter bodies) and CityCatalyst issues a Climate Advisor user token per card. Review badges only apply when `draft.status === "complete"`, so skip the query unless the run can actually be reviewed, or return a lightweight review summary on the run list instead.

app/src/services/api.ts
```diff
@@ -2529,20 +2532,32 @@ export const api = createApi({
...
         invalidatesTags: (_result, _error, runId) => [
           { type: "ConceptNoteDraft", id: runId },
         ],
       }),
+      validateConceptNoteChapter: builder.mutation<
+        ConceptNoteChapterValidationResponse,
+        ValidateConceptNoteChapterRequest
+      >({
+        query: ({ chapterId, runId }) => ({
+          url: `concept-notes/${runId}/chapters/${chapterId}/validation/`,
```
cephaschapa, line 2554: `validateConceptNoteChapter` invalidates `ConceptNoteDraft` after every chapter POST. Guided review runs up to 3 validations in parallel, so a 10-chapter review refetches the full draft ~10 times on top of the workspace's 15s draft polling. `onReviewComplete` already refetches; drop per-chapter invalidation (or invalidate once after the batch) so review does not stampede Climate Advisor.

app/src/components/ConceptNoteDashboard/index.tsx
```diff
@@ -460,16 +471,31 @@ export function ConceptNoteDashboard({
...
+          draftError={exportDraftFailed && exportDraft === undefined}
+          hasApplicationTemplate={Boolean(exportApplicationContext?.template)}
           hasUploadedEvidence={
             exportBundle?.availableContext.uploadedDocuments ?? false
           }
           lng={lng}
           noteName={exportRun.name}
           open
+          runId={exportRun.run_id}
+          onAddInformation={() => {
```
cephaschapa: Dashboard `onAddInformation` ignores `chapterId` / `findingKey` and only `router.push`es the resume URL. Workspace already focuses a chapter/finding from those args, so Export from a run card drops the guided "one fix path". Pass the ids as resume query params (or equivalent) and read them in the workspace the same way `onAddInformation` already works there.

cephaschapa (approved):
Great work, thanks 👍👍👍

app/src/components/ConceptNoteDashboard/run-card.tsx
piotrnowakowski: Addressed. Run cards now use the lightweight `progress_summary.draft_document.status` signal and skip `useGetConceptNoteDraftQuery` unless the draft is complete and the run is non-terminal. This avoids unnecessary full-draft and CA token requests while preserving review badges for eligible runs.

app/src/services/api.ts
piotrnowakowski: Addressed. I removed `ConceptNoteDraft` invalidation from each chapter-validation mutation. Guided review now relies on the existing single `onReviewComplete` refetch after the parallel batch finishes. I also added coverage confirming the batch triggers one completion refresh.

app/src/components/ConceptNoteDashboard/index.tsx
piotrnowakowski: Addressed. Dashboard `onAddInformation` now carries `chapterId` and `findingKey` as encoded resume query parameters. The workspace page reads them and initializes the draft focus, so the user lands on the exact chapter and finding. Navigation encoding is covered by tests.

PR #3103, opened 2026-09-02
fix(CC-657): update csv-download e2e test for checkbox download modal

tatimoreira:
## Summary
- The GHGI download modal was redesigned to use checkbox format selection with a single confirm button (`redesign(CC-657): rework GHGI download modal with checkbox format selection`), but `e2e/csv-download.spec.ts` still targeted the removed per-format buttons (`download-csv-button`, `download-ecrf-button`), causing CI to fail: https://github.com/Open-Earth-Foundation/CityCatalyst/actions/runs/33532927691/job/99940231632
- Updated the spec to check the relevant format checkbox(es) and click the shared `download-confirm-button`, following the existing Chakra/Ark checkbox click convention from `signup.spec.ts`
- Reworked the "Multiple format downloads" test to select both formats and confirm once (since confirming now triggers all selected downloads together and closes the modal), matching each `Download` by filename instead of assuming event order

## Test plan
- [x] `npx tsc --noEmit` passes for the changed file
- [x] `npx eslint e2e/csv-download.spec.ts` passes
- [x] CI e2e run passes on this PR

### DIFF
--- a/app/e2e/csv-download.spec.ts
+++ b/app/e2e/csv-download.spec.ts
@@ -41,12 +41,24 @@ async function openDownloadModal(page: Page) {
   });
 }
 
-async function downloadCsv(page: Page): Promise<Download> {
+/** Chakra v3 / Ark checkbox: click the control part rather than the root. */
+async function selectDownloadFormat(page: Page, format: string) {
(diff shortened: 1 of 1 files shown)


lemilonkh (approved):
Awesome, thanks 👍 ✨

PR #3106, opened 2026-09-02
fix: unify input height and font-size across settings tabs

briansebaOEF:
## Summary

All form inputs and selects on the Settings page (Account, Team, Projects, API Tokens, and their sub-tabs) now render at a consistent 44px height and `fontSize=md` (`var(--chakra-font-sizes-md)`), instead of a mix of default sizes, `size=lg`, and hardcoded 48px heights.

## Changes

- `form-input.tsx`, `email-input.tsx`, `password-input.tsx`: add an optional `inputProps` passthrough so Settings call sites can override sizing without changing these components' appearance on the login/signup pages.
- `ProfessionSelect`, `LanguageSelector`, `PreferencesTab`, `CreateTokenForm`, `ProjectSearchInput`, `deleteCityModal`: apply `h=44px fontSize=md` directly (settings-only usages).
- `BrandSettingsTab`: override the Chakra `Select`'s `--select-trigger-height` CSS var and set `fontSize` on `SelectValueText`, since its recipe sets font-size directly on the trigger element rather than inheriting it.
- `data-table-alt.tsx`: align the shared search input and role-filter select (used only by the Team/Projects tabs) to the same sizing.

## How to test

1. Go to Settings and check each tab (Account Details, Brand Settings, Password, Preferences, Team, Projects, API Tokens).
2. Confirm every text input, password input, and dropdown/select is the same height (44px) and font size.
3. Confirm login, signup, forgot-password, and admin modals (which share some of these input components) are unaffected.

## Notes

Verified visually and via computed-style checks (`getComputedStyle`) in a running dev server; `tsc --noEmit` passes clean.

ðŸ¤– Generated with [Claude Code](https://claude.com/claude-code)


app/src/app/[lng]/settings/account/BrandSettingsTab.tsx
```diff
@@ -209,7 +209,8 @@ const BrandSettingsTab = ({
             shadow="1dp"
             borderRadius="4px"
             border="inputBox"
-            fontSize="body.lg"
+            fontSize="md"
+            css={{ "--select-trigger-height": "44px" }}
```
Comment (lemilonkh) [inline app/src/app/[lng]/settings/account/BrandSettingsTab.tsx]: We could add this to the SelectRoot component itself?

app/src/app/[lng]/settings/account/ManagePasswordTab.tsx
```diff
@@ -110,6 +111,7 @@ const ManagePasswordTab: FC<ManagePasswordProps> = ({ t }) => {
             id="newPassword"
             shouldValidate
             watchPassword={watchPassword}
+            inputProps={{ h: "44px", fontSize: "md" }}
```
Comment (lemilonkh) [inline app/src/app/[lng]/settings/account/ManagePasswordTab.tsx]: Should we define this in our PasswordInput component so it's consitent across the application (including signup/ login)?

lemilonkh (approved):
Great, thanks 👍 ✨

app/src/app/[lng]/settings/account/BrandSettingsTab.tsx
Comment (briansebaOEF) [reply app/src/app/[lng]/settings/account/BrandSettingsTab.tsx]: Good call — added a `triggerHeight` prop to `SelectRoot` that sets the CSS var internally, so callers don't need to know about `--select-trigger-height`. Existing usages are unaffected since it's optional.

app/src/app/[lng]/settings/account/ManagePasswordTab.tsx
Comment (briansebaOEF) [reply app/src/app/[lng]/settings/account/ManagePasswordTab.tsx]: Yes, moved it into `PasswordInput` itself so login/signup/update-password get the same sizing — removed the redundant `inputProps` overrides in `ManagePasswordTab.tsx`. Verified the login page's password field now renders at 44px too, with the visibility-toggle icon still centered correctly.

app/src/app/[lng]/settings/account/BrandSettingsTab.tsx
Comment (briansebaOEF) [reply app/src/app/[lng]/settings/account/BrandSettingsTab.tsx]: Confirmed fixed — `SelectRoot` now exposes `triggerHeight`, and `BrandSettingsTab` uses it instead of the raw `css` prop.

app/src/app/[lng]/settings/account/ManagePasswordTab.tsx
Comment (briansebaOEF) [reply app/src/app/[lng]/settings/account/ManagePasswordTab.tsx]: Confirmed fixed — `PasswordInput` defaults to `size=lg` now (44px/16px), applied app-wide (login/signup/update-password/settings), and the redundant per-instance overrides are gone. Also swapped the other hardcoded `h=44px fontSize=md` spots across this PR to `size=lg` for the same reason — it's Chakra's exact 44px/16px preset rather than a magic-number stand-in.

PR #3107, opened 2026-09-03
fix: plan card styling, layout, and profession hint cleanup

briansebaOEF:
## Summary

Follow-up changes to the plan card and general Settings > Account layout, on top of #3106.

## Changes

**Plan card** (Settings > Account)
- 8px border radius, `shadow-lg` box shadow
- Removed `greta@openearth.org` from the mailto/contact text
- Replaced the folder icon with the provided design, rendered at 24x24
- Removed the card's top margin; content vertically centered
- Org name now uses `title.lg` font size
- Plan badge: 165x165, 6px rounded corners, watermark re-centered (top 48px), label text at `label.lg` font size

**Settings layout**
- 24px gap between the account-details card and the plan card
- Top-level settings tabs `margin-bottom` reduced to 24px
- Removed the "Indicate your title or profession." hint text under the Profession select
- Account tab email input now uses `shadow-1dp`

## How to test

1. Go to Settings > Account and check the plan card: rounded corners/shadow, centered content, badge size/rounding, mailto text, folder icon.
2. Check demo/trial-plan orgs to see the badge variants if available.
3. Confirm the Profession select no longer shows the hint text below it.
4. Confirm spacing between the account-details card and the plan card, and the top tab bar spacing.


app/src/app/[lng]/settings/account/index.tsx
```diff
@@ -88,7 +88,7 @@ const AccountSettingsTab = ({ t }: { t: TFunction }) => {
                 )}
               </TabContent>
             )}
-            <TabContent value="account-details" p={0}>
+            <TabContent value="account-details" p={0} gap="24px">
```
Comment (lemilonkh) [inline app/src/app/[lng]/settings/account/index.tsx]: There is only one component as the direct child of TabContent in all of the cases, so I don't think this does much?

app/src/app/[lng]/settings/account/ProfessionSelect.tsx
```diff
@@ -45,41 +42,31 @@ export function ProfessionSelect({
...
-              </option>
-            ))}
-          </NativeSelectField>
-        </NativeSelectRoot>
-      </Field>
-      <HStack>
-        <BodyMedium color="content.link">
-          <MdInfoOutline />
-        </BodyMedium>
-        <BodyMedium>{t("position-description")}</BodyMedium>
```
Comment (lemilonkh) [inline app/src/app/[lng]/settings/account/ProfessionSelect.tsx:80]: If we are removing this text, let's also remove the translation strings in the respective languages (in `app/src/i18n/locales/*/settings.json`).

app/src/components/PlanDetailsBox.tsx
```diff
@@ -21,6 +20,12 @@ const PlanDetailsBox: React.FC<PlanDetailsBoxProps> = ({ organization }) => {
 
   if (!organization) return null;
 
+  const supportEmails = (env("NEXT_PUBLIC_SUPPORT_EMAILS") ?? "")
+    .split(",")
+    .map((email) => email.trim())
+    .filter((email) => email && email !== "greta@openearth.org")
```
lemilonkh: Let's not put this into the codebase but rather edit the environment variable in the deployment scripts.

lemilonkh: In general I don't think this code block is necessary as we shouldn't filter out specific email addresses in the application code.

app/src/app/[lng]/settings/account/index.tsx
Comment (briansebaOEF) [reply app/src/app/[lng]/settings/account/index.tsx]: You're right. I'll remove it

app/src/app/[lng]/settings/account/ProfessionSelect.tsx
Comment (briansebaOEF) [reply app/src/app/[lng]/settings/account/ProfessionSelect.tsx:80]: Will do

app/src/components/PlanDetailsBox.tsx
briansebaOEF: Will do

briansebaOEF:
@lemilonkh All three review points are addressed now, ready for another look:

- **`TabContent` gap**: removed the unnecessary `gap="24px"` prop from the `account-details` `TabContent` in `account/index.tsx`.
- **Profession hint text**: removed the "Indicate your title or profession." text and its now-unused `position-description` translation key from all locales (`en`, `pt`, `fr`, `de`, `es`).
- **Support emails**: removed the code-level filter in `PlanDetailsBox.tsx` — it now reads `NEXT_PUBLIC_SUPPORT_EMAILS` directly like every other consumer, no app-side address filtering. Also dropped `greta@openearth.org` from `app/env.example`. Note: I couldn't find any k8s manifest/workflow in this repo that sets that env var for staging/prod, so it looks like it lives in an external ConfigMap/Secret that'll need updating separately for the change to take effect there.

app/env.example
```diff
@@ -18,7 +18,7 @@ SMTP_USER="user"
 SMTP_PASSWORD="password"
 SMTP_FROM_EMAIL="ethereal.user@ethereal.email"
 RESET_TOKEN_SECRET="WwjFGhcP8VikgsmtcPv+ufPhTGS8t7e+/aN5k2qY4ms="
-NEXT_PUBLIC_SUPPORT_EMAILS="info@openearth.org,greta@openearth.org"
```
tatimoreira, line 21: I think this can be removed

tatimoreira (approved):

app/env.example
briansebaOEF: Just to confirm the scope here: only `greta@openearth.org` was dropped — the `NEXT_PUBLIC_SUPPORT_EMAILS` var itself is still needed. It's read server-side at request time in `getPublicRuntimeEnv()` (`app/src/lib/runtime-env/keys.ts`) and injected into the client via `window.__ENV`, so k8s/runtime overrides aren't baked in at build. If it's missing from `.env`, every support-email mailto link across the app (navbar, error pages, chat disclaimer, plan card, etc.) breaks locally. So the old line with greta's address is what's removed, and that's it.

cephaschapa (approved):
Thank @briansebaOEF 🙌

PR #3109, opened 2026-09-03
feat(api): CC-762 implement remaining service API endpoints for MEED

lemilonkh:
Implements the 4 remaining API routes needed to communicate with the HIAP MEED service:

- [x] POST /api/v1/city/[city]/meed/actions/translate
- [x] POST /api/v1/city/[city]/meed/exclusions-preview
- [x] GET /api/v1/city/[city]/meed/feasibility-scores
- [x] POST /api/v1/city/[city]/meed/generate-plan

It also adds this route to re-fetch created plan (action report) from the database, similar to how the ranking works
- [x] GET /api/v1/city/[city]/meed/generate-plan

For the generate plan route there is the new table MeedActionReport, which stores the results of the call so it can be retrieved at a later date without re-running the plan generation.
It also stores the full request and response context of the ranking run in the new table MeedRankSnapshot.

TODO:
- [x] Test all routes extensively with the service

### DIFF
--- a/app/migrations/20260903120451-create-meed-rank-snapshot.cjs
+++ b/app/migrations/20260903120451-create-meed-rank-snapshot.cjs
@@ -1,0 +1,49 @@
+"use strict";
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.createTable("MeedRankSnapshot", {
+      id: {
+        type: Sequelize.UUID,
+        allowNull: false,
+        primaryKey: true,
+        defaultValue: Sequelize.UUIDV4,
+      },
+      inventory_id: {
+        type: Sequelize.UUID,
+        allowNull: false,
+        references: {
+          model: "Inventory",
+          key: "inventory_id",
+        },
+        onUpdate: "CASCADE",
+        onDelete: "CASCADE",
+      },
+      request: {
+        type: Sequelize.JSONB,
+        defaultValue: {},
+        allowNull: false,
+      },
+      response: {
+        type: Sequelize.JSONB,
+        defaultValue: {},
+        allowNull: false,
+      },
+      created: {
+        type: Sequelize.DATE,
+        allowNull: false,
+        defaultValue: Sequelize.NOW,
+      },
+      last_updated: {
+        type: Sequelize.DATE,
+        allowNull: false,
+        defaultValue: Sequelize.NOW,
+      },
+    });
+  },
+
+  async down(queryInterface) {
+    await queryInterface.dropTable("MeedActionReport");
+  },
+};
--- a/app/migrations/20260903120451-create-meed-rank-snapshot.cjs
+++ b/app/migrations/20260903120451-create-meed-rank-snapshot.cjs
@@ -44,6 +44,6 @@ module.exports = {
   },
 
   async down(queryInterface) {
-    await queryInterface.dropTable("MeedActionReport");
+    await queryInterface.dropTable("MeedRankSnapshot");
   },
 };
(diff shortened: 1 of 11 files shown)


app/migrations/20260903120451-create-meed-rank-snapshot.cjs
```diff
@@ -0,0 +1,49 @@
...
+      last_updated: {
+        type: Sequelize.DATE,
+        allowNull: false,
+        defaultValue: Sequelize.NOW,
+      },
+    });
+  },
+
+  async down(queryInterface) {
+    await queryInterface.dropTable("MeedActionReport");
```
tatimoreira: this file creates `MeedRankSnapshot` in up(), but its down() drops MeedActionReport instead (looks copy-pasted from the sibling 20260903120517-create-meed-action-report.cjs migration and not updated).

Effect: rolling back both migrations in sequence (db:migrate:undo x2, or :undo:all) will error on the second undo because MeedActionReport was already dropped by the other migration's down(), and MeedRankSnapshot itself never actually gets dropped — it leaks and the schema drifts from what the migration history claims.
```
async down(queryInterface) {
  await queryInterface.dropTable("MeedRankSnapshot");
},
```

app/src/backend/MeedApiService.ts
```diff
@@ -343,15 +376,188 @@ export default class MeedApiService {
...
+      locode: inventory.city.locode,
+      actionId,
+      language: languages,
+      prioritizationSnapshot: {
+        request: rankSnapshot.request,
+        response: rankSnapshot.response,
+      },
+      debugContextOnly,
+    };
+    console.log("DATA", JSON.stringify(data, null, 2));
```
tatimoreira: remove comment

app/src/app/api/v1/city/[city]/meed/actions/translate/route.ts
```diff
@@ -0,0 +1,67 @@
+import MeedApiService from "@/backend/MeedApiService";
+import { PermissionService } from "@/backend/permissions/PermissionService";
```
Comment (tatimoreira) [inline app/src/app/api/v1/city/[city]/meed/actions/translate/route.ts]: should we import this `PermissionService` from `@/backend/permissions` as `exclusions-preview/route.ts` and `feasibility-scores/route.ts` for consistency?

app/migrations/20260903120451-create-meed-rank-snapshot.cjs
cephaschapa: down() drops MeedActionReport instead of MeedRankSnapshot. Undo will leave MeedRankSnapshot in place (and can fail or no-op after the later migration already dropped the report table). Point down() at "MeedRankSnapshot".

app/src/backend/MeedApiService.ts
```diff
@@ -343,15 +376,188 @@ export default class MeedApiService {
...
+    const result = await this.makeRequest("reports/output-plan", data);
+    console.log(result);
+
+    // save result to database
+    const report = await db.models.MeedActionReport.create({
```
cephaschapa: generatePlan always creates a MeedActionReport, and getPlan uses findOne with no order. Re-running plan generation leaves duplicates; GET can return an old row. Delete-or-upsert by (inventoryId, actionId) (same pattern as ranking) and add a unique index in the report migration.

app/src/backend/MeedApiService.ts
cephaschapa: we can use logger here and below

cephaschapa (approved):
Awesome, great work ✨✨✨

app/migrations/20260903120451-create-meed-rank-snapshot.cjs
lemilonkh: Thanks, fixed! :raised_hands:

app/src/backend/MeedApiService.ts
lemilonkh: Great spot, thanks :raised_hands:

PR #3113, opened 2026-09-03
CC-815: Cache repeated Climate Advisor authorization checks

piotrnowakowski:
## Summary

Reduces repeated Concept Note authorization traffic by combining process-local token reuse in CityCatalyst with a 30-second successful authorization cache in Climate Advisor. Concurrent requests are coalesced at both boundaries; failures, denied access, and expired credentials are never retained.

## Changes

- reuse one successfully issued CNB token per user until one minute before expiry
- coalesce concurrent token issuance and isolate entries by user
- cache successful CA identity and city-access checks for at most 30 seconds
- key CA entries by a one-way token fingerprint and scope
- bound both caches and retain no failed results
- document process-local and expiry behavior

## Ticket

[CC-815](https://linear.app/openearth/issue/CC-815/cnb-cache-repeated-climate-advisor-authorization-checks)

## Notes

- No API, database, environment, or dependency changes.
- Both caches are intentionally process-local and cold after restart.
- Local Kraków smoke test over 61 seconds: 1 CA token issued, 3 identity checks, 3 city checks, and no 429/5xx after CA became healthy.
- This implements the narrow CNB token-reuse prerequisite; CC-757 still owns broader token-issuance retries and error propagation.


tatimoreira (approved):

climate-advisor/service/app/utils/single_flight_cache.py
```diff
@@ -0,0 +1,103 @@
...
+                self._values[key] = _CacheEntry(
+                    value=value,
+                    expires_at=expires_at,
+                )
+                self._values.move_to_end(key)
+                while len(self._values) > self._max_entries:
+                    self._values.popitem(last=False)
+            return value
+        finally:
+            if self._inflight.get(key) is asyncio.current_task():
```
tatimoreira: Suggestion (non-blocking): this check appears to always be true given how the task is registered in `_inflight` before it starts running (nothing else can overwrite the entry for this key while it's in flight) — could likely be simplified to an unconditional `self._inflight.pop(key, None)`.

climate-advisor/service/app/services/citycatalyst_client.py
```diff
@@ -898,6 +936,34 @@ async def get_city(
...
+
+    async def _request_city(
+        self,
+        *,
+        city_id: str,
+        token: str,
+        user_id: str,
+    ) -> Dict[str, Any]:
+        """Fetch one city directly from CityCatalyst."""
+        if not self.base_url:
```
tatimoreira: Suggestion (non-blocking): this `base_url` check duplicates the one already done in `get_city` before the cache key/loader is built — harmless dead code that could be dropped from `_request_city`.

climate-advisor/service/app/utils/single_flight_cache.py
piotrnowakowski: Agreed—simplified the cleanup to an unconditional `self._inflight.pop(key, None)`.

climate-advisor/service/app/services/citycatalyst_client.py
piotrnowakowski: Agreed—removed the duplicate check from `_request_city`; validation remains in `get_city` before building the cache key and loader.

PR #3114, opened 2026-09-03
fix(settings): account tab polish (brand, password, preferences)

briansebaOEF:
## Summary
- **Brand settings**: removed the "Set up your own brand" copy, clarified the logo format hint (.png), fixed the "Primary color" label casing, removed a duplicate select border, and stopped a full page reload/redirect after saving so the user stays on the tab.
- **Manage password**: fixed spacing above the current-password field, stopped the new-password field from flagging invalid before the user even types, enabled the reset button as soon as all fields are filled (deferring pattern/mismatch/"same as current" checks to submit time), and fixed a stale error re-appearing after a reset.
- **Preferences**: preferred language now defaults to the navbar's active language instead of hardcoded English when unset; numerical format now correctly falls back to comma-and-dot for unset or legacy/unrecognized values (was rendering blank); save button now only enables once something actually changed; button height matches the other account tabs.
- Consistent `minW`/height styling across the account tab's action buttons (Save changes / Reset password).
- Added `new-password-same-as-current` and updated `password-hint` translations across en/de/es/fr/pt.

## Test plan
- [ ] Verify Brand Settings: logo upload preview close icon behavior, save without full-page reload, primary color label/border
- [ ] Verify Manage Password: no error shown before typing, reset button enables once all fields filled, submit-time errors for weak/same/mismatched passwords, error clears correctly after a successful reset
- [ ] Verify Preferences: preferred language defaults to navbar language, numerical format shows "Comma and dot" by default (not blank), save button disabled until a change is made
- [ ] `NUMERICAL_FORMATS` feature flag enabled locally to exercise the numerical formats field


app/src/app/[lng]/settings/account/logo-file-upload.tsx
```diff
@@ -84,7 +84,7 @@ const LogoUploadCard = ({
                 borderRadius="2xl"
               />
             )}
-            {(files.length > 0 || defaultUrl) && (
+            {files.length === 0 && defaultUrl && (
```
Comment (lemilonkh) [inline app/src/app/[lng]/settings/account/logo-file-upload.tsx]: This seems off. Why would we show the delete button only if there are no files to delete?

app/src/app/[lng]/settings/account/BrandSettingsTab.tsx
```diff
@@ -170,16 +167,6 @@ const BrandSettingsTab = ({
...
       </Text>
-      <Text
-        color="content.tertiary"
-        fontWeight="normal"
-        lineHeight="24"
-        fontSize="body.lg"
-        letterSpacing="wide"
-        marginTop="8px"
-      >
-        {t("brand-settings-description")}
```
Comment (lemilonkh) [inline app/src/app/[lng]/settings/account/BrandSettingsTab.tsx:181]: Let's also delete the translation strings in the different languages for this if possible.

lemilonkh (approved):
Amazing, thanks 🙌 ✨

app/src/app/[lng]/settings/account/BrandSettingsTab.tsx
Comment (briansebaOEF) [reply app/src/app/[lng]/settings/account/BrandSettingsTab.tsx:181]: I instructed the AI to do it. But I guess it didn't. I'll check

app/src/app/[lng]/settings/account/logo-file-upload.tsx
Comment (briansebaOEF) [reply app/src/app/[lng]/settings/account/logo-file-upload.tsx]: I'll check

Comment (briansebaOEF) [reply app/src/app/[lng]/settings/account/logo-file-upload.tsx]: Good catch — that read backwards without context. It wasn't actually a functional bug (the file-list row below already has its own remove control for a newly-picked file), but the condition itself was confusing. Renamed it to `canClearSavedLogo` and added a comment explaining the scope: this icon only clears the already-saved org logo, so it's hidden while a new file is staged to avoid a duplicate delete affordance. Pushed in 69d405f7f.

app/src/app/[lng]/settings/account/BrandSettingsTab.tsx
Comment (briansebaOEF) [reply app/src/app/[lng]/settings/account/BrandSettingsTab.tsx:181]: Checked — this was already done. The `brand-settings-description` key (and its translated value) is removed from all 5 locales (en/de/es/fr/pt), in both settings.json and account-settings.json, as part of 9a5f65dfb. Verified it's still absent after the develop merge. Nothing further needed here.

PR #3115, opened 2026-09-03
[1/2] feat(cnb): add reviewed Concept Note chat edits

piotrnowakowski:
Users can request changes to a Concept Note in chat and review the proposed text inline before accepting it. Proposals persist across reloads; acceptance checks the current revision and applies only the selected changes.

The change retains structured missing-information markers and chapter confirmation. Accepting a grounded edit resolves its matching gap in the same transaction. The guided gap interview, its cards, and background chapter-regeneration workflow are excluded. Authorization caching is handled separately in #3113.

The planner independently reviews proposed edits, validates anchors and provenance, and uses the configured CNB model routing. The migration provisions proposal/application records, structured gaps, and exact revision confirmations.

Validation:

- 34 focused Python tests passed; the PostgreSQL migration integration test was skipped because no disposable PostgreSQL test URL was configured.
- 18 Jest tests passed with Node's ESM mode enabled.
- TypeScript check and ESLint passed for the changed frontend components.
- Verified the authorization implementation and its tests match `develop`, and checked for references to removed interview/regeneration APIs and workers.

Browser and live-model behavior were not exercised in this scope-removal pass.

## Stack

1. **#3115 - Climate Advisor reviewed edit workflow** (this PR)
2. #3120 - CityCatalyst inline edit review


climate-advisor/service/app/services/cnb/edit_planner.py
```diff
@@ -0,0 +1,355 @@
...
+        "chapter": {
+            "title": chapter.title,
+            "position": chapter.position,
+            "body_markdown": chapter.body_markdown,
+            "confirmed_body_markdown": chapter.confirmed_body_markdown,
+            "gaps": [
+                {"question": gap.question, "state": gap.state} for gap in chapter.gaps
+            ],
+        },
+        "run_context": {
```
DavidGBbr: `build_planner_input()` sends the raw `run_context` to the planner. This object includes `selected_sources.upload_id` and `sha256`, as well as other internal identifiers. That breaks the CNB boundary that keeps backend identities and fingerprints out of model-facing payloads, and exposes these values to the LLM provider.

Could we send the planner only an allowlisted projection, such as `source_index`, `source_label`, summary, and topics? The backend could then resolve `source_index` when creating `EditSourceSnapshot`. Could we also add a test confirming that planner payloads contain no IDs, hashes, or storage keys?

climate-advisor/service/app/services/agent_service.py
```diff
@@ -381,6 +402,17 @@ async def create_agent(
                         token_ref=self._token_ref,
                     )
                 )
+                if self.concept_note_edit_request is not None:
```
DavidGBbr, line 406: The runtime registers `concept_note_edit_propose`, but the main prompt still says that chat does not persist edits and only documents the source-query tool. For an explicit edit request, the model may respond with unsaved wording instead of calling the edit tool.

Could we add a clear rule to `prompts/cnb/chat.md`: when the user requests a document change and the tool is available, call `concept_note_edit_propose`? The tool creates a reviewable proposal but does not apply the document change. Could we cover this behavior with a tool-selection test or a live model smoke test?

climate-advisor/service/app/services/cnb/edit_validation.py
```diff
@@ -0,0 +1,343 @@
...
+            status_code=422,
+        )
+    return snapshots
+
+
+def index_sources(sources: list[dict[str, Any]]) -> dict[str, dict[str, Any]]:
+    """Index selected sources once by both immutable ID and visible label."""
+    result = {}
+    for source in sources:
+        for key in (source.get("upload_id"), source.get("source_label")):
```
DavidGBbr: `index_sources()` overwrites sources when two entries have the same `source_label`. If the planner returns that label in `source_refs`, validation may record the hash from the wrong source.

Could we treat duplicate labels as ambiguous and require a unique `source_index` or source identity in those cases? Could we add a test with two sources named `plan.pdf`?

DavidGBbr (commented):
Great Job!

climate-advisor/service/app/services/cnb/edit_planner.py
piotrnowakowski: Fixed. The models no longer receive internal IDs or hashes. Added tests for initial requests and follow-up edits.

climate-advisor/service/app/services/agent_service.py
piotrnowakowski: Fixed. The prompt now tells the model to use the edit tool when changes are requested. Verified with live model tests.

climate-advisor/service/app/services/cnb/edit_validation.py
piotrnowakowski: Fixed. Files with the same name are now handled separately. Added a test with duplicate filenames to verify this.

DavidGBbr (approved):

PR #3116, opened 2026-09-04
CC-816: Remove legacy OpenAI Assistants API

cephaschapa:
## Summary

Hard-cut Clima chat to Climate Advisor (`/api/v1/chat/*`) and remove the legacy OpenAI Assistants path that was 404ing in production when `OPENAI_ASSISTANT_ID` pointed at a deleted assistant.

## Changes

- Route `useChat` / `chatService` / SSE streaming through Climate Advisor only
- Delete `/api/v1/assistants/**` routes and swagger tag
- Remove `AssistantThread` / `AssistantMessage` models and add drop migration
- Drop `OPENAI_ASSISTANT_ID` from deploy workflows and `env.example`
- Keep `CA_SERVICE_INTEGRATION` for CNB / stationary-energy agentic gating

## How to test

1. Open Clima chat on an inventory and confirm messages stream via `/api/v1/chat/*` only (no `/assistants` calls in network tab)
2. Confirm thread create still works via `POST /api/v1/chat/threads`
3. After migrate: verify `AssistantThread` / `AssistantMessage` tables are gone
4. Smoke CNB / stationary-energy agentic still gated by `CA_SERVICE_INTEGRATION`

## Ticket

https://linear.app/openearth/issue/CC-816/removing-legacy-assistant-api

## Notes

Prod-only failure mode was dual-backend routing + stale assistant ID; local/dev/test were fine because those environments still used Climate Advisor or a valid assistant. This PR removes the dual path entirely rather than rotating the OpenAI assistant ID.


app/src/hooks/useChat.ts
```diff
@@ -141,12 +88,11 @@ export function useChat({ inventoryId, t }: UseChatProps) {
     try {
       await initializeThread();
 
-      // Use conditional URL based on feature flag
-      const messageUrl = hasFeatureFlag(FeatureFlags.CA_SERVICE_INTEGRATION)
```
lemilonkh, line 145: Let's also remove the feature flag for it?
Both in the env vars as well as the FeatureFlags code file.

app/src/services/api.ts
```diff
@@ -1157,20 +1157,6 @@ export const api = createApi({
           "SectorBreakdown",
         ],
       }),
-      createThreadId: builder.mutation({
```
lemilonkh, line 1160: Was this the only endpoint for the assistants that we still had in the RTK API file?
Since a lot more API route files were removed...

lemilonkh (approved):
Awesome, thanks 🙌✨

PR #3117, opened 2026-09-04
CC-817: Inject public runtime env on each request

cephaschapa:
## Summary

Fix empty `window.__ENV` on deployed builds so client feature flags (and other allowlisted `NEXT_PUBLIC_*` vars) come from the pod at request time instead of a static prerender from the Docker build.

## Changes

- `RuntimeEnvScript` awaits `connection()` before serializing `process.env`
- Document the static-prerender failure mode and how to verify locally with `build` + `start`
- Add a unit test for `getPublicRuntimeEnv`

## How to test

1. Incognito: open `/en/`, run `window.__ENV` — must be non-empty after deploy (not `{}`)
2. `window.qaFlags.debug()` — environment flags match deploy `NEXT_PUBLIC_FEATURE_FLAGS`
3. Local: build without public env, start with flags set, confirm `__ENV` is filled

## Ticket

https://linear.app/openearth/issue/CC-817/fix-feature-flags-not-working-on-dev-and-perhaps-other-envs


lemilonkh (approved):
Amazing work, thanks 👍🏻✨

PR #3120, opened 2026-09-08
[2/2] feat(cnb): add CityCatalyst inline edit review

piotrnowakowski:
## Summary

Adds CityCatalyst inline review for Concept Note chat edits. Users can accept or reject individual changes, apply remaining decisions without overriding earlier choices, and recover from failed draft reloads beside the document controls.

## Changes

- Adds reviewed edit, chapter confirmation, chat reset, source cleanup, and export integration.
- Blocks source deletion throughout active OCR processing and delivery.
- Refreshes edit proposals after chapter confirmation and tests authorization and request validation at the proxy boundary.

## How to test

- All 14 changed Jest suites passed: 159 tests.
- TypeScript and ESLint checks passed locally.
- Final-commit CI passed: production build, OpenAPI validation, API tests, coverage checks, Playwright, auth contract, and CodeQL.
- CI: https://github.com/Open-Earth-Foundation/CityCatalyst/actions/runs/34610163478

## Notes

Stack order: #3115 (merged Climate Advisor reviewed edit workflow), then #3120 (this PR). Updated onto develop after #3115 merged; verified identical file contents before and after GitHub restacked the branch.


app/src/components/ConceptNoteWorkspace/edit-proposal-card.tsx
```diff
@@ -0,0 +1,392 @@
...
+              >
+                {t("edit-reject-all")}
+              </Button>
+              <Button
+                minH="44px"
+                size="sm"
+                data-testid="concept-note-edit-apply-all"
+                disabled={busy || !canApply}
+                loading={busy}
+                onClick={() => void onApply(proposal)}
```
cephaschapa, line 221: This toolbar never sees decisions. Accept all calls onApply(proposal) with no selected_change_ids, so previously rejected hunks still get applied. Reject all has the same problem in reverse. Disable these while any per-change decision exists, or apply only the accepted/undecided IDs.

app/src/backend/ConceptNoteDeletionService.ts
```diff
@@ -0,0 +1,35 @@
...
+    const where = {
+      sourceType: "concept_note_upload",
+      sourceId: { [Op.in]: uploadIds },
+    };
+    const jobs = await db.models.PdfOcrJob.findAll({
+      where,
+      transaction,
+      lock: transaction.LOCK.UPDATE,
+    });
+    if (
```
cephaschapa, line 20: This only blocks running / delivering. PdfOcrJob also has queued and delivery pending. Pending delivery is the real hole: a worker can still be reading S3 while this deletes objects and destroys the row. Treat the pipeline as active when status ∈ {queued, running} or deliveryStatus ∈ {pending, delivering}. Tests today only cover running/delivering.

app/src/components/ConceptNoteWorkspace/chat-panel.tsx
```diff
@@ -372,6 +427,26 @@ export function ConceptNoteChatPanel({
           </Box>
         ))}
 
+        {edits.error && (
```
cephaschapa, line 430: Apply/reject failures only show in the chat transcript. Accept/Reject/Refine live in the document header, so a failed Accept all looks like a no-op. Surface the same alert next to DocumentReviewToolbar.

Related: use-concept-note-edits.ts:174-179 — if apply succeeded but onApplied / refetchDraft() failed, keep a sticky “reload draft” on the document panel. The cache already marks the proposal applied.

app/src/backend/concept-note-edits.ts
```diff
@@ -0,0 +1,65 @@
...
+
+type EditProxyOperation =
+  "list" | "propose" | "read" | "apply" | "reject" | "refine";
+type EditProxyContext = {
+  session: AppSession | null;
+  params: Record<string, string>;
+};
+
+/** Authorize every run-scoped edit operation and preserve the CA status/body. */
+export async function forwardConceptNoteEdit(
```
cephaschapa, line 25: No Jest coverage for this proxy (or confirm / chat reset). Deletion has solid route tests; these routes are the IDOR boundary. Worth 401, UUID 400, city denied, apply/refine body schemas, and forwarding of CA { code, detail } into editErrorCode

app/src/services/api.ts
```diff
@@ -2567,6 +2582,22 @@ export const api = createApi({
...
+      >({
+        query: ({ runId, chapterId, expectedRevision, idempotencyKey }) => ({
+          url: `concept-notes/${runId}/chapters/${chapterId}/confirm`,
+          method: "POST",
+          body: {
+            expected_revision: expectedRevision,
+            idempotency_key: idempotencyKey,
+          },
+        }),
+        invalidatesTags: (_result, _error, { runId }) => [
```
cephaschapa, line 2597: This only invalidates ConceptNoteDraft. Polling for edits only runs while status is processing (use-concept-note-edits.ts:29-34), so the list can stay proposed after confirm. Local proposalMatchesDraft already disables apply, so this is not a silent-apply bug — still invalidate { type: "ConceptNoteEdits", id: runId }.

cephaschapa (approved):
Looking good 👍👍👍, left some comments

app/src/components/ConceptNoteWorkspace/edit-proposal-card.tsx
piotrnowakowski: Fixed. The toolbar now uses “Accept remaining” / “Reject remaining” after individual decisions. Both preserve existing choices and submit only accepted change IDs. Added tests for mixed decisions and failed-request retries.

app/src/backend/ConceptNoteDeletionService.ts
piotrnowakowski: Fixed. Deletion now returns 409 when status is queued/running or delivery status is pending/delivering. Tests cover all four states and verify that neither storage objects nor job rows are deleted.

app/src/components/ConceptNoteWorkspace/chat-panel.tsx
piotrnowakowski: Fixed. Apply/reject errors now appear beside the document toolbar. If applying succeeds but reloading the draft fails, a persistent “Reload draft” action remains visible even after the proposal leaves the toolbar. Tests cover repeated reload failures and successful recovery.

app/src/backend/concept-note-edits.ts
piotrnowakowski: Added route tests covering 401s, invalid UUIDs, denied city access, apply/refine body validation, and confirm/chat reset. Also verified that upstream status, `code`, and `detail` are preserved and correctly interpreted by `editErrorCode`.

app/src/services/api.ts
piotrnowakowski: Fixed. Chapter confirmation now also invalidates `{ type: "ConceptNoteEdits", id: runId }`. Added a test verifying that both draft and proposals refresh without polling, while another run’s cache remains unchanged.

PR #3124, opened 2026-09-08
feat(meed): generate action reports from the results screen

carlosgraffi:
The "Generate report" button on the results screen has been disabled since the screen was built. #3088's `generate-plan` route landed, so this makes it work.

```
POST city/{cityId}/meed/generate-plan   { inventoryId, actionId, languages }
GET  city/{cityId}/meed/generate-plan?inventoryId=&actionId=
```

## Three properties of the route shaped this

**One report per action.** `actionId` is singular upstream, so the multi-action report the screen has always offered is N generations combined at render time. Ten selected actions is ten calls.

**Each is a 10–30 s LLM call.** So the button reports progress per action rather than spinning, and generation runs **sequentially** — hiap-meed caps concurrent reports per pod and returns 429 past the queue wait, so firing ten at once would make some of them fail by design.

**Reports are stored server-side.** Each action is fetched before it is generated, and a 404 there is treated as "not generated yet" rather than an error. Looking at a report twice does not pay for it twice — which is the point of the table Milan added.

## Shape

Split so the parts worth testing do not need a PDF renderer:

| File | |
|---|---|
| `report/reportDocument.ts` | chapters → one language, ordered, empty chapters dropped |
| `report/reportMarkdown.ts` | chapter Markdown → typed blocks |
| `report/meedReportPdf.ts` | jsPDF layout |
| `report/useMeedReport.ts` | orchestration, progress, partial failures |

`reportMarkdown` is a deliberate subset — headings, bullets, paragraphs. The chapters are prose with the occasional list, and a partial renderer that silently mangles anything richer would be worse than one whose limits are stated. Inline emphasis is stripped to its visible text rather than honoured, because jsPDF changes weight per `text()` call.

`meedReportPdf` follows the conventions in `services/PDFExportService.ts` — A4 in millimetres, same margins, `splitTextToSize`, explicit page breaks. That service's helpers are private and shaped around the HIAP action plan, so this is the same approach rather than the same code.

## Failure handling

Actions whose report cannot be produced are **named**, not quietly absent. The run returns them alongside the document, and the screen says which were left out — a short report with no explanation is the worse outcome. If nothing generates at all, no PDF is produced and the screen says so.

## Copy

`generate-report-pending` ("Report generation arrives with the prioritization backend") is deleted along with its key. The caption now always says something true: progress while running, what is missing when nothing is selected, and roughly how long to expect otherwise.

## Testing

11 new unit tests across the two pure modules (55 total in the MEED suites, all passing); `tsc` clean; production build passes; no new lint findings.

Not yet exercised against the live route — the honest check is generating a report on dev after merge. `debugContextOnly` exists on the route for exactly this, and is worth using to verify the wiring without paying for LLM calls; I have not wired it to any UI, since it is a testing affordance rather than a user-facing one.


app/src/app/[lng]/cities/[cityId]/MEED/[inventory]/results/report/useMeedReport.ts
```diff
@@ -0,0 +1,137 @@
...
+
+        let report: MeedPlanRouteReport | null = null;
+        try {
+          report = await fetchPlan({
+            cityId,
+            inventoryId,
+            actionId: target.actionId,
+          }).unwrap();
+        } catch {
+          // 404 — no report stored for this action yet.
```
Comment (cephaschapa) [inline app/src/app/[lng]/cities/[cityId]/MEED/[inventory]/results/report/useMeedReport.ts]: The GET catch treats every failure as “no report yet” and then POSTs. `isFetchBaseQueryError` already exists in `@/util/helpers` — only 404 should generate. A 401/500/network error on GET will fire a 10–30s LLM call per selected action (cost and load), and a 5xx is reported as a generation miss instead of a fetch failure. Rethrow non-404s and wrap the loop in try/finally so `running` cannot stick.

app/src/app/[lng]/cities/[cityId]/MEED/[inventory]/results/page.tsx
```diff
@@ -131,6 +134,41 @@ export default function Page(props: {
...
   }, [ranked, index, t, inventory]);
 
+  const { generate, isRunning, progress } = useMeedReport(cityId, inventoryId);
+  const [reportError, setReportError] = useState<string | null>(null);
+
+  const generateReport = useCallback(async () => {
+    setReportError(null);
+    const targets = selectedIds.map((actionId) => ({
+      actionId,
+      actionName: catalogActionName(index, actionId, t),
```
Comment (cephaschapa) [inline app/src/app/[lng]/cities/[cityId]/MEED/[inventory]/results/page.tsx:149]: `generate()` returns `null` when a run is already in flight, and this treats `!result` as total failure. `isGenerating` updates asynchronously, so a double-click on the CTA sets “No reports could be generated” while the first run continues; a later success does not clear that error, so the banner can sit on top of a downloaded PDF. Ignore `null` (already running) and only show `report-error-none` when `documents.length === 0` after a completed run.

cephaschapa (approved):
Great work diego 👍👍👍

app/src/app/[lng]/cities/[cityId]/MEED/[inventory]/results/report/useMeedReport.ts
Comment (carlosgraffi) [reply app/src/app/[lng]/cities/[cityId]/MEED/[inventory]/results/report/useMeedReport.ts]: Fixed in a268249f9 — both were right.

The GET catch now gates on a literal 404 via `isReportNotFound`, built on the existing `isFetchBaseQueryError`. Non-404s abort the run rather than falling through to a POST: a 401 or 5xx affects every remaining action too, so continuing would have paid for nine more doomed calls. They surface a separate error card — "try again in a moment" is the wrong instruction for an expired session. The loop is wrapped in try/finally so the one-shot guard and `isRunning` reopen on any exit.

On the caller side, `null` (a run already in flight) is now a no-op, and `report-error-none` is only shown for a completed run with zero documents.

Four unit tests cover the 404 discrimination, including the case worth being careful about: `PARSING_ERROR` carries `originalStatus: 404` but is not a missing report.

app/src/app/[lng]/cities/[cityId]/MEED/[inventory]/results/page.tsx
Comment (carlosgraffi) [reply app/src/app/[lng]/cities/[cityId]/MEED/[inventory]/results/page.tsx:149]: Fixed in a268249f9 — both were right.

The GET catch now gates on a literal 404 via `isReportNotFound`, built on the existing `isFetchBaseQueryError`. Non-404s abort the run rather than falling through to a POST: a 401 or 5xx affects every remaining action too, so continuing would have paid for nine more doomed calls. They surface a separate error card — "try again in a moment" is the wrong instruction for an expired session. The loop is wrapped in try/finally so the one-shot guard and `isRunning` reopen on any exit.

On the caller side, `null` (a run already in flight) is now a no-op, and `report-error-none` is only shown for a completed run with zero documents.

Four unit tests cover the 404 discrimination, including the case worth being careful about: `PARSING_ERROR` carries `originalStatus: 404` but is not a missing report.

PR #3126, opened 2026-09-09
fix(CC-770): protect manually-edited modules from seeder overwrite

tatimoreira:
## Summary

Prevents the modules seeder from clobbering module data that's been manually edited through the admin panel: seeded (and re-seeded) POC/OEF modules were unconditionally overwritten on every deploy, discarding any admin edits.

## Changes

- Add `is_manually_edited` boolean column on `Module` (migration + model), default `false`.
- `bulkUpsert` (seeder util) accepts an optional protect-column; when set and the existing row has it `true`, the update is skipped instead of overwriting.
- Modules seeder passes `is_manually_edited` as the protect column for the `Module` table.
- Admin PUT route sets `isManuallyEdited: true` whenever an edit is saved.

## How to test

1. `npx jest tests/seeders-bulk-upsert.jest.ts tests/api/admin-modules-route.jest.ts` — covers the skip behavior and the admin route setting the flag.
2. Manual: run `npx sequelize-cli db:migrate`, edit a POC module via the admin panel (or `PUT /api/v1/admin/modules/[module]`), then re-run `npx sequelize-cli db:seed --seed 20250923103314-modules.cjs` and confirm the edit is preserved. An unedited module should still be overwritten by the seed data.

## Ticket

[CC-770](https://linear.app/openearth/issue/CC-770/adjust-module-seeder-process-to-not-overwrite-manually-edited-data)

## Notes

Scope follows the discussion on the ticket with @lemilonkh: only POC modules are editable via the admin panel today, and those are also created by the seeder, which is where the original conflict came from.


lemilonkh (approved):
Great fix, thanks! 🎉

PR #3129, opened 2026-09-09
fix(meed): make action reports readable — tables, branding, and a clearer block

carlosgraffi:
Two things, both found from a live report on dev. Reviewable commit by commit.

---

## 1. `afb0ec4f7` — the missing rank snapshot

Generating a report on dev failed. From the HAR, both calls 404 but they are different animals:

```
GET  .../meed/generate-plan/?inventoryId=d3844922…&actionId=icare_0024
404  {"error":{"message":"Plan not found"}}

POST .../meed/generate-plan/
404  {"error":{"message":"Rank snapshot not found - run ranking first"}}
```

The GET 404 is the normal "not generated yet" path. The POST is the failure.

`MeedApiService.generatePlan` reads a `MeedRankSnapshot` row — the prioritization request/response pair `reports/output-plan` is written from. That row is created by the ranking POST, and only by the version that landed in #3109 on 2026-09-04. An inventory ranked before that deploy renders a full ranking with nothing behind it.

**No backend change needed** — re-running the ranking writes the snapshot. The frontend has to say so, which it now does: generation 404s abort the run instead of spending one doomed POST per selected action, and the error links back to pre-flight. `MeedReportBlockedError` carries the reason so a run that never started gives the right instruction rather than "try again in a moment".

## 2. `b338ffd9a` — the report looked unformatted

The chapters emit GitHub-flavoured pipe tables (`hiap-meed/.../prompts/city_action_report_*.md`) and `markdownToBlocks` had no table case, so raw `|---|---|` syntax was flowed into the PDF as paragraph text. Six chapters carry a table; all six were unreadable.

**Inline markup is now kept as runs, not stripped.** Bold survives — these chapters lead with `**The ask:**` — and so do links, which is what makes the financing table's "Programme page" cells mean anything: they become real PDF annotations. Wrapping measures across runs, so a line no longer breaks at every bold boundary.

**Tables** get proportional column widths from their content, alignment from the delimiter row, per-cell wrapping, and a header that repeats when a table crosses a page.

**Styling follows `services/PDFExportService.ts`**: the brand palette from `app-theme.ts`, the wide CityCatalyst lockup, an accent rule under the title, brand bars on section headings, tinted table headers and limitation bands, and a footer with page numbers.

Two incidental fixes:

- **`jsPDF` is now a named import.** The ESM build has no default export, so the default form resolves only through a bundler's interop — which is why this renderer could not be unit-tested before.
- **`compress: true`.** The logo was embedded as a raw bitmap, costing ~850 KB on every report. A two-page report goes from **958 KB to 59 KB**.

### Not done

`jspdf-autotable` is already a dependency, imported by `PDFExportService` for side effects and never used. It would handle pagination for free, but it has no model for styled runs inside cells, so bold labels and clickable links would need `didDrawCell` hooks anyway — and mixing its layout with this one's cursor is more coordination than it saves. Worth revisiting if tables get much richer.

## Verification

78 tests across 8 MEED suites (24 new). Five of them render a real PDF from the actual Antofagasta chapter content and assert what was broken: multi-page output, link annotations, file size, one page per action, and an empty document.

Checked by extracting the rendered output: tables lay out in columns, the right-aligned Page column is right-aligned, accents survive, the header repeats on page 2, and the three financing links are real `/URI` annotations.

`tsc` clean, production build passes, no new lint findings.

## Note for the backend

`Inventory not found` and `Rank snapshot not found` are both 404, so the frontend infers intent from a bare status. The second is arguably a **409** — the resource exists, its precondition doesn't. Not changed here.


lemilonkh (approved):
Amazing, thanks 👍 ✨

PR #3133, opened 2026-09-09
CC-761: wait for document context before starting Clima chat

piotrnowakowski:
## Summary

Keep Clima chat blocked until uploaded documents finish processing and their context is ready. Chat, draft setup, and Context now use one shared status so a received PDF is consistently shown as processing, with actionable errors if analysis fails.

## Changes

- Preserve the processing gate across reloads and automatically enable chat when context is ready.
- Enforce complete source-section coverage and retry mismatched partitions in bounded smaller batches.
- Persist safe error reasons and counts, surface messages such as incomplete source coverage with a retry action, and add diagnostic logs.

## How to test

1. Create a concept note with a long PDF. Confirm the shared processing message appears and chat cannot submit, including after reloading.
2. Wait for processing to complete, then ask a document-specific question and check its page citations.
3. Exercise a source-analysis failure and confirm the same actionable error appears across the workspace.

Local validation passed: 34 backend tests, 21 frontend tests, TypeScript, changed-file ESLint, and diff checks. A real 106-page Krakow adaptation-plan upload completed OCR and analysis; chat stayed blocked until ready, then answered correctly with verified page citations.

## Ticket

CC-761

https://github.com/user-attachments/assets/7aed8c23-0ec9-449a-aa06-cf50cabee04d


app/src/components/ConceptNoteWorkspace/context-status.ts
```diff
@@ -0,0 +1,133 @@
...
+  isUploading: boolean;
+  isRetrying: boolean;
+}): ConceptNoteContextState {
+  if (isUploading) return "uploading";
+  if (isRetrying) return "processing";
+
+  // Keep pending persisted uploads authoritative until the run poll catches up.
+  const statuses = uploads.map((upload) => upload.status);
+  if (activeUpload) statuses.push(activeUpload.status);
+  if (statuses.includes("failed") || bundle.status === "failed")
```
DavidGBbr: A failed older upload can block chat indefinitely. getConceptNoteContextState returns failed if any upload has failed, while the workspace restores only the newest upload after a reload. Reproduction: upload A fails, then upload B succeeds and the context bundle becomes ready. After reload, B is selected and shown as ready, but A still makes the aggregate state failed. The chat remains disabled and the UI offers no way to retry A. Could we derive the blocking state from the selected/current upload and bundle instead, or expose retry/recovery for every failed upload? Please add a rendered multi-upload regression test for this scenario.

app/src/components/ConceptNoteWorkspace/context-status.ts
```diff
@@ -0,0 +1,133 @@
...
+    ? "ready"
+    : "none";
+}
+
+export function getConceptNoteContextPresentation(
+  state: ConceptNoteContextState,
+  bundle: ConceptNoteBundleProgress,
+  t: ReturnType<typeof useTranslation>["t"],
+) {
+  const blocked = state !== "ready" && state !== "none";
```
DavidGBbr, line 86: Does this need server-side enforcement too? This only disables the UI based on client-side state. Climate Advisor still creates an agent when the bundle is unavailable, so a client that bypasses the UI could submit a turn before document context is ready. If this is a product invariant rather than only a UX rule, could the backend validate bundle readiness and return a stable error code?

DavidGBbr (approved):
Good job, I loved the video!

app/src/components/ConceptNoteWorkspace/context-status.ts
piotrnowakowski: Addressed. Older failed uploads no longer block a newer successful upload or count toward required ready sources. Pending uploads, a failed current upload, and bundle failures still block chat. Added a rendered regression covering failed A → successful B → reload → enabled composer and successful submission.

app/src/components/ConceptNoteWorkspace/context-status.ts
piotrnowakowski: Agreed—this is now enforced by the backend. CNB turns validate persisted upload and bundle readiness before saving the user message or starting generation. Unready context returns HTTP 409 with `concept_note_context_not_ready`, which the UI handles with recovery guidance. Tests cover pending, failed, missing/stale context, ownership, and ready document/city-only runs.

PR #3134, opened 2026-09-10
CC-749: Fix negative/net emissions display in sector breakdowns

tatimoreira:
## Summary

Fixes how negative/removal values are displayed in the sector and sub-sector emissions breakdown (e.g. AFOLU land use removals), so percentages, amounts, and totals read correctly instead of showing nonsensical negative or >100% values.

## Changes

- Percentages are now computed against gross (emissions-only) totals, so a sector's emitting sub-sectors always sum to 100%; removal sub-sectors get a `null` percentage instead of a skewed one.
- Removal amounts display as a positive number labeled "removed" instead of a raw negative number.
- A sector whose removals outweigh its emissions shows "Net removal" instead of a percentage of total emissions.
- Removal rows in the sub-sector percentage column show `*` (instead of a "Removal" label), with a footnote below the table: "Sub-sectors with removals are not countable as emissions" (per Carlos's follow-up feedback on the ticket).
- Centralized the percentage calculation previously duplicated across `ByScopeView.tsx` and `ReportResults.tsx`.

## How to test

1. Open an inventory with an AFOLU land-use removal (e.g. São Paulo, year 2022) → GHGI → "Emission inventory results" tab → AFOLU sector tab.
2. In the sub-sector table, the "Emissions from land within the city boundary" row should show its amount as "X ktCO2e removed" and `*` in the "% of sector emissions" column.
3. A footnote should appear below the table: "Sub-sectors with removals are not countable as emissions".
4. Other (non-removal) sub-sectors in the same sector should still show a normal `%`, summing to 100%.
5. If the sector's net total is negative, the sector header should read "Net removal" instead of a percentage-of-total-emissions line.
6. `npm run jest -- results.jest results_sector.jest` for the backend percentage calculation tests.

## Ticket

CC-749

## Notes

The ticket's other open item — replacing "No cited source" with "Uploaded" / "Manual entry" based on data provenance — is **not** included here. There's currently no field distinguishing a CSV/XLS import from manual entry, so it needs new backend tracking; flagged as a possible follow-up in a comment on CC-749.


lemilonkh (approved):
Amazing work restructuring the app to support this! Thanks ✨

PR #3140, opened 2026-09-14
feat: CC-674 implement two factor authentication

lemilonkh:
## Summary

Implements two factor authentication (2FA) with TOTP apps, including settings page and login UI

## Changes

- Adds the following API routes in `/v1/auth/2fa`: `setup`, `verify`, `disable`, `check`
- All are POST routes and require authentication except for `check`, which is used on the login screen to determine if the "Security code" field should be shown to the user
- Requires 2FA code during login if it's enabled for the user
- Adds Security sub-tab on Account tab on settings page to allow users to set up 2FA

## How to test

1. Go to settings page, then to the Security sub-tab in the Account tab
2. Enable 2FA
3. Scan QR code using e.g. Google Authenticator or Authy app on your phone
4. Enter 6 digit verification code from authenticator app and submit
5. Log out
6. Once you enter your email, the "Security code" field should show up
7. Enter 6 digit verification code from authenticator app and submit
8. Should be logged in correctly :heavy_check_mark:

## Ticket

CC-674


app/src/app/api/v1/auth/2fa/check/route.ts
```diff
@@ -0,0 +1,51 @@
...
+  email: z.string().email(),
+});
+
+export const GET = apiHandler(async (_req, { searchParams }) => {
+  const { email } = check2FAQuery.parse(searchParams);
+  const user = await db.models.User.findOne({
+    where: { email },
+  });
+  if (!user) {
+    throw new createHttpError.NotFound("User not found");
```
tatimoreira: 🟠 **Security: account enumeration via this endpoint**

This route is unauthenticated (by design, since it needs to run on the login screen) and distinguishes known vs. unknown emails by HTTP status: `404 User not found` here vs. `200 {enabled: ...}` below. That lets anyone probe arbitrary addresses to discover which ones have a CityCatalyst account, before ever attempting a login.

Suggest returning a uniform `200 { enabled: false }` for unknown emails instead of a `404`, so the response shape doesn't leak whether the account exists.

app/src/app/[lng]/auth/login/page.tsx
```diff
@@ -131,7 +165,33 @@ export default function Login(props: { params: Promise<{ lng: string }> }) {
               value.length >= 8 || t("min-length", { length: 8 })
             }
           />
-          <Text color="semantic.danger">{t(error)}</Text>
+          {showSecurityToken && (
+            <Field
```
Comment (cephaschapa) [inline app/src/app/[lng]/auth/login/page.tsx:173]: Seems this field is only visually hidden. register() still marks securityToken as required, so users without 2FA cannot submit?

app/src/app/api/v1/auth/2fa/setup/route.ts
```diff
@@ -0,0 +1,62 @@
...
+ *         description: Internal server error
+ */
+import { generateQRCode, generateTwoFactorSecret } from "@/lib/2fa";
+import { db } from "@/models";
+import { apiHandler } from "@/util/api";
+import createHttpError from "http-errors";
+import { NextResponse } from "next/server";
+
+export const POST = apiHandler(async (req, { session }) => {
+  if (!session) {
```
tatimoreira, line 40: 🟠 **Security: no re-authentication for enabling/disabling 2FA**

This route only checks for a valid session — no current password or current TOTP confirmation is required. Two consequences:

- `twoFactorSecret` is overwritten unconditionally here, even when `twoFactorEnabled` is already `true` for this user. A hijacked/stolen session (XSS, shared device, etc.) can silently swap in a new secret the attacker knows, without disabling 2FA first — the legitimate user's authenticator app then no longer matches what's in the DB.
- The companion `/2fa/disable` route (`app/src/app/api/v1/auth/2fa/disable/route.ts`) has the same gap: session-only, no re-auth, so anyone with the session cookie can strip 2FA protection outright.

Sensitive account-security changes like enabling/rotating/disabling 2FA typically require re-entering the password (or the current TOTP code) in addition to having a session, precisely to guard against this class of hijack.

app/src/app/api/v1/auth/2fa/disable/route.ts
```diff
@@ -0,0 +1,57 @@
...
+ *       500:
+ *         description: Internal server error
+ */
+import { db } from "@/models";
+import { apiHandler } from "@/util/api";
+import createHttpError from "http-errors";
+import { NextResponse } from "next/server";
+
+export const POST = apiHandler(async (_req, { session }) => {
+  if (!session) {
```
tatimoreira, line 53: 🟠 **Security: no re-authentication required to disable 2FA**

Same gap as `/2fa/setup` (see comment there): this route only checks for a valid session, with no password or current TOTP confirmation. Anyone holding the session (stolen cookie, XSS, shared/unlocked device) can disable 2FA on the account with no further proof. Recommend requiring the current password or a valid TOTP code before honoring this request.

app/src/app/[lng]/settings/account/SecurityTab.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+  const [token, setToken] = useState("");
+  const [isSetupMode, setIsSetupMode] = useState(false);
+  const [hasTokenError, setHasTokenError] = useState(false);
+
+  const isEnabled = userInfo?.twoFactorEnabled ?? false;
+
+  const setup2FA = async () => {
+    const result = await setupSecondFactorAuth();
+    console.log("2FA result", result);
```
Comment (cephaschapa) [inline app/src/app/[lng]/settings/account/SecurityTab.tsx]: we can remove the console log.

app/src/app/[lng]/settings/account/SecurityTab.tsx
```diff
@@ -0,0 +1,185 @@
...
+    if (result.data?.success) {
+      setIsSetupMode(true);
+    }
+  };
+
+  const verify2FA = async () => {
+    const result = await verifySecondFactorAuth({ token });
+    console.log("Verify result", result);
```
Comment (cephaschapa) [inline app/src/app/[lng]/settings/account/SecurityTab.tsx]: here as well

app/src/lib/auth.ts
```diff
@@ -92,6 +94,21 @@ export const authOptions: NextAuthOptions = {
           return null;
         }
 
+        if (user.twoFactorEnabled && user.twoFactorSecret) {
```
tatimoreira, line 106: 🟠 **Security: TOTP checked before password, no rate limiting**

The security-token check runs (and can `return null` early) *before* `bcrypt.compare`. Since bcrypt is deliberately slow and TOTP verification is fast, this creates a timing side-channel: a request with a wrong token fails fast, while one with a wrong password (but correct token) pays the bcrypt cost. That lets an attacker brute-force the 6-digit code cheaply, independent of guessing the password correctly.

This is compounded by the fact that this `authorize()` callback isn't wrapped by `apiHandler`, so it gets none of the app's existing rate limiting (`RateLimiter` in `@/util/rate-limiter`, used elsewhere via `apiHandler`) — login/2FA attempts here are currently unthrottled.

Suggested fix: always run the (slow, constant-cost) password check first, and only check the TOTP token once the password is confirmed valid — that removes the fast-fail oracle:

```ts
        const isValid = await bcrypt.compare(
          credentials.password,
          user.passwordHash,
        );
        if (!isValid) {
          logger.error("Invalid password!");
          return null;
        }

        if (user.twoFactorEnabled && user.twoFactorSecret) {
          if (!credentials.securityToken) {
            logger.error("No securityToken passed for user with 2FA enabled");
            return null;
          }
          const isTokenValid = await verifyToken(
            credentials.securityToken,
            user.twoFactorSecret,
          );
          if (!isTokenValid) {
            logger.error("Invalid securityToken for 2FA");
            return null;
          }
        }
```

Separately (can't be expressed as a same-range suggestion since it needs a module-level addition), consider adding a per-email `RateLimiter` instance to this `authorize()` callback — e.g. 5 attempts/minute — so repeated failed logins/2FA guesses get throttled the same way the rest of the API already is.

app/src/i18n/locales/en/settings.json
```diff
@@ -243,10 +243,23 @@
...
+  "two-factor-setup-message": "Click the button below to set up two-factor authentication for additional account security. You will need an authenticator app on your phone.",
+  "two-factor-enabled-message": "Two-factor authentication is enabled",
+  "two-factor-reset-message": "To switch to a different authenticator app or device, disable and set up two factor authentication again.",
+  "two-factor-setup-button": "Setup two-factor authentication",
+  "two-factor-reset-button": "Disable two-factor authentication",
+  "two-factor-verify-button": "Verify two-factor secret",
+  "two-factor-scan-message": "Scan this QR code using an authenticator app on your phone and enter the displayed code below.",
+  "two-factor-token-placeholder": "Enter the 6 digit code from your authenticator app here",
+  "two-factor-setup-success": "Successfully enabled two-factor authentication",
+  "two-factor-setup-failed": "Failed to enabled two-factor authentication. Please try again.",
```
cephaschapa:
```suggestion
  "two-factor-setup-failed": "Failed to enable two-factor authentication. Please try again.",
```

app/src/models/User.ts
```diff
@@ -250,6 +257,16 @@ export class User
           allowNull: true,
           field: "number_format",
         },
+        twoFactorEnabled: {
+          type: DataTypes.BOOLEAN,
+          allowNull: true,
```
cephaschapa, line 262: model sets allow now to true, migration sets it to false

cephaschapa (approved):
Great work 👍👍👍

app/src/lib/auth.ts
lemilonkh: Done, added follow-up ticket CC-875 for the other suggestion. Great, thanks :raised_hands:

app/src/app/[lng]/auth/login/page.tsx
```diff
@@ -79,7 +87,28 @@ export default function Login(props: { params: Promise<{ lng: string }> }) {
...
+  const fullError =
+    error == "invalid-email-or-password" && showSecurityToken
+      ? "invalid-email-password-or-security-code"
+      : error;
+
   const onSubmit: SubmitHandler<LoginInputs> = async (data) => {
+    const status = await run2FAStatusCheck();
+    if (status?.enabled && data.securityToken?.length != 6) {
+      return;
+    }
```
Comment (tatimoreira) [inline app/src/app/[lng]/auth/login/page.tsx:114]: 🔴 **Regression: login can fail silently / throw an unhandled rejection for unknown emails**

`run2FAStatusCheck()` calls `check2FAStatus({ email }).unwrap()` with no `try/catch`. `/api/v1/auth/2fa/check` returns a `404` when no user matches the given email, so `.unwrap()` rejects for that case — and this `await` (line 107) propagates that rejection straight out of the `onSubmit` handler.

Concretely: a user who mistypes their email (or simply doesn't have an account) and clicks "Log in" gets **no error message at all** — not even the existing "invalid email or password" text. The rejection never reaches `clearError()`/`login()`, so the failure is silent to the user (visible only as a console unhandled-rejection). That's a real regression vs. current behavior on `develop`.

Separately, when the check *does* succeed and `status.enabled && securityToken.length != 6` (line 108), the early `return` gives no feedback either — the security-code field pops into view but there's no explicit message telling the user why the submit didn't go through.

Suggest wrapping the `.unwrap()` call in a `try/catch` (treating a failed/absent 2FA-check as "not enabled" rather than aborting the whole submit), and setting an explicit error/toast on the early-return path so users get a visible reason instead of a silent no-op.

app/src/app/[lng]/auth/login/page.tsx
Comment (lemilonkh) [reply app/src/app/[lng]/auth/login/page.tsx:173]: If the field is not part of the markup (when showSecurityToken is false), this is makes the field no longer required. Just tested it.
I previously had it only visually hidden with display: none in case it was false, but that produced the exact error you mentioned.

app/src/app/api/v1/auth/2fa/setup/route.ts
lemilonkh: The setup route no longer allows overwriting the secret now. The issue with the disable route will be handled as part of that comment.

lemilonkh: The setup route no longer allows overwriting the secret now. The issue with the disable route will be handled as part of that comment.

app/src/app/api/v1/auth/2fa/verify/route.ts
```diff
@@ -0,0 +1,79 @@
...
+ *     operationId: verify2FA
+ *     summary: Verify a second factor code from the authenticator app to enable 2FA auth
+ *     description: Verifies the 2FA code in the request against the user's stored 2FA secret and enabled two factor authentication if it's valid
+ *     requestBody:
+ *       required: true
+ *       content:
+ *         application/json:
+ *           schema:
+ *             type: object
+ *             required: [userId]
```
tatimoreira: 🟡 **Swagger doc mismatch: `required: [userId]` but no `userId` field exists**

This schema only declares a `token` property (below), and `userId` doesn't appear anywhere in this request. Looks like a copy-paste leftover from another route. This also means `token` itself isn't marked as required in the generated OpenAPI spec, even though the handler's Zod schema (`z.object({ token: z.string().min(1) })`) does require it.

Suggested fix:
```suggestion
 *             required: [token]
```

tatimoreira:
🟡 **Test coverage**: none of the new 2FA routes, `lib/2fa.ts`, or the `authorize()` changes have test coverage. Given this touches auth, I'd suggest either adding tests before merge, or filing a follow-up ticket (e.g. off CC-674) to add coverage for: the four `/2fa/*` routes (`setup`/`verify`/`disable`/`check`), `generateTwoFactorSecret`/`verifyToken` in `lib/2fa.ts`, and the new 2FA branch in `authorize()`. Flagging so it doesn't get lost — up to you whether it blocks this PR.

app/src/app/[lng]/auth/login/page.tsx
Comment (lemilonkh) [reply app/src/app/[lng]/auth/login/page.tsx:114]: Fixing the request error catching. The silent failure is fine as this will only happen when the user presses enter before the security code input is visible - otherwise there will be a validation error on that field preventing the form submission and showing a proper error message to the user.

app/src/app/api/v1/auth/2fa/disable/route.ts
lemilonkh: Password verification is now required to disable 2FA. Added a modal in SecurityTab for entering and confirming it.

lemilonkh:
Created follow-up ticket CC-877 for adding the tests

PR #3141, opened 2026-09-14
fix(cnb): use canonical CC source cleanup URL; grant CNB module in smoke fixture

carlosgraffi:
Two small fixes found while testing the Concept Note Builder locally on `develop` (post #3115/#3120) with the Kraków and Richfield demo notes from `codex/CNB-fixtures`. Screen recording of the session: https://jam.dev/c/6dde3e57-63a3-4a0b-bb54-86309485bcfe

## Fixes in this PR

**1. Deleting a note that has sources returns 503**
`citycatalyst_client.delete_concept_note_sources` targeted `/api/v1/internal/ca/concept-note-sources` without the trailing slash CityCatalyst enforces. Next answers 308, httpx raises on it, and `concept_note_lifecycle.delete_run` maps that to 503 "Concept Note could not be deleted". The client now calls the canonical `/api/v1/internal/ca/concept-note-sources/` URL directly, avoiding both the 308 and forwarding the `X-CA-Service-Key` through an arbitrary redirect. Any run with ≥1 unshared upload was undeletable; runs with no uploads deleted normally. This affects prod too since it also enforces the trailing slash. One-line fix + a regression test (`35 passed`).

**2. Smoke fixture seeds no module access**
`upsert-ca-smoke-fixture` creates user/org/project/city/inventory but no `ProjectModules` row, so on a clean local DB `useModuleAccess` denies the seeded city and the CNB routes redirect away with a `not-found-description` toast — the note looks nonexistent. Now grants Concept Note Builder to the fixture project (idempotent `findOrCreate`).

## Found but not changed here (for your call)

- **Review before export shows green all-clears when nothing was reviewed.** With every chapter failing validation, the panel said "0 chapters reviewed, 12 could not be checked" and directly below ✅ "No missing required information", ✅ "No evidence warnings", with Continue enabled. A total validation failure reads as a clean bill of health on an export gate.
- **The demo fixtures don't load against `develop` as-is** (both Kraków and Richfield):
  - `ca.concept_note_runs` carries `lifecycle_state` / `duplicated_from_run_id`, and `cnb.concept_note_chapters` carries `regeneration_status` / `regeneration_error` — none exist on develop (CC-730 columns). The seeder fails with `UndefinedColumn`.
  - `funder_templates.chapter_schema` entries have no per-chapter `required_fields`, so `chapter_validation.py` rejects every chapter with 409 `chapter_validation_template_invalid`. Review before export can't run on either note. (The assignments are derivable from the chapter descriptions — 6/6 and 24/24 partition exactly.)
  - Richfield's context bundle `selected_sources[0]` lacks `upload_id` + `sha256`, which `SelectedSource` now requires → `application-context` 400.
  - A `cnb` DB stamped at CC-730's `20260823_120000` blocks alembic on develop.
  Probably easiest to re-snapshot both fixtures from current runs. I patched all of the above in my local DB only to keep testing — nothing in this PR touches fixtures.
- **Clean-setup gaps worth a README line:** source-PDF upload on a new note requires S3 (`AWS_FILE_UPLOAD_S3_BUCKET_ID/_REGION` + creds) and fails as `source_storage_failed` → 503 without it; the CA compose port moved 8080 → 8081; OpenRouter keys minted with an expiry silently die (ours expired 2026-09-11).

## Validation
- `climate-advisor/service/tests/test_citycatalyst_client.py`: 35 passed (incl. new `test_concept_note_source_deletion_uses_canonical_trailing_slash_url`)
- `app`: `tsc --noEmit` clean, ESLint clean on the changed script


piotrnowakowski (approved):
Great job — the two fixes are focused and well tested. I added one small patch: the source-cleanup client now calls the canonical trailing-slash URL directly instead of using `follow_redirects=True`. This keeps the deletion fix while preventing HTTPX from forwarding the `X-CA-Service-Key` and DELETE body through an unexpected redirect. All focused tests and CI checks pass.

PR #3142, opened 2026-09-14
fix(cnb): preserve readiness rejection status

piotrnowakowski:
## Summary

- preserve upstream HTTP error responses from the Climate Advisor before starting the SSE stream
- return the intended `409 concept_note_context_not_ready` and related non-stream errors instead of converting them into HTTP 200 event streams
- pass the Concept Note run ID with workspace chat requests and verify that the run belongs to the supplied thread
- add web and Climate Advisor regression coverage and update the relevant architecture documentation

## Why

The deployed CC-761 readiness test confirmed that an early direct request to `POST /api/v1/chat/messages` was not persisted and did not visibly start a model call, but the web route returned HTTP 200 instead of the required HTTP 409 contract. API clients and monitoring could therefore interpret a rejected request as success.

## Verification

- web Jest suites: 28/28 passed
- Climate Advisor pytest suite: 18/18 passed
- ESLint passed
- TypeScript typecheck passed
- Ruff passed

## Deployment status

This branch is not deployed. The current development environment still runs the original merged `develop` revision, so the live 409 verification remains a post-deployment check.

Follow-up to #3133.


DavidGBbr (approved):
I didn't find any problem to highlight, good job man!

PR #3145, opened 2026-09-15
CC-827: Keep chat SSE responses alive during silent work

piotrnowakowski:
## Summary

Keep chat SSE responses active while model/tool work is silent. Develop failures occurred about 60 seconds after planning started; the configured 60-second ALB idle timeout and missing heartbeats strongly indicate an idle disconnect. ALB termination logs were unavailable.

## Changes

- Send an initial SSE comment and 15-second idle heartbeats on the existing response.
- Preserve byte order, backpressure, tracing context, and producer error propagation.
- Close the source and shield cleanup when the request is cancelled.

## How to test

From `climate-advisor/service`, run `python -m pytest tests/test_sse_heartbeat.py -q`: **6 tests passed** independently. Tests cover silent work, ordering/backpressure, tracing context, failures, cancellation, and the message route. Real ALB verification remains outstanding.

## Ticket

Partial improvement for [CC-827](https://linear.app/openearth/issue/CC-827/cnb-even-tiny-chat-edits-process-all-chapters-and-disconnect-during).

## Notes

Independent branch targeting develop; no stack or dependency on #3148. Four changed files, limited to SSE transport, tests, and README. No demo video.

Generation remains request-bound. Durable reconnect/restart recovery, edit scoping, latency, and invalid-anchor failures remain unresolved. Heartbeats do not prove successful edit processing.

Reasoning streaming and the investigation document live separately in #3148 (CC-907). Original combined work is preserved at `codex/pr-3145-pre-stack-backup-ac01540` (`ac01540aa429b58020aa9fe18e0cf949bd17f893`).


climate-advisor/service/app/utils/sse_heartbeat.py
```diff
@@ -0,0 +1,65 @@
...
+from __future__ import annotations
+
+import asyncio
+from collections.abc import AsyncGenerator
+from contextlib import aclosing
+
+from anyio import CancelScope
+
+SSE_HEARTBEAT_INTERVAL_SECONDS = 15.0
+SSE_HEARTBEAT = b": keep-alive\n\n"
```
DavidGBbr, line 12: The `: keep-alive` comments are forwarded through the proxy to `useSSEStream`. Currently, every non-empty frame is dispatched: `parseSSEEvent` ignores the `:` line and returns an empty event, causing the default handler to log `Unhandled SSE event type`. This produces at least one false warning per chat and another every 15 seconds while the stream is silent. Could we ignore comment-only SSE frames and add a frontend regression test covering both the initial and recurring heartbeats?

DavidGBbr (approved):

climate-advisor/service/app/utils/sse_heartbeat.py
piotrnowakowski: Thanks for spotting this. `useSSEStream` now skips comment-only heartbeat frames before dispatch, so they no longer produce false warnings. I added a frontend regression test covering both the initial and a recurring heartbeat; the focused tests passed. Fixed in 3d4827bd7.

PR #3146, opened 2026-09-15
CC-875: Rate limit login attempts

tatimoreira:
## Summary

Throttles repeated failed login attempts (and 2FA code guesses) per email address, closing the gap where `/api/auth/*` bypassed the API-wide rate limiter entirely.

## Changes

- Add a `RateLimiter` instance in `authorize()` (`src/lib/auth.ts`): 5 attempts / 15 minutes per email, disabled under `PLAYWRIGHT_TEST` like the existing API limiter.
- Throw a distinct `rate-limited` error from `authorize()` so it can be told apart from a wrong password.
- Map that error to a new `too-many-login-attempts` message in `useLogin.ts` / `auth.json`, instead of the generic invalid-credentials text.
- Add `tests/auth-rate-limit.jest.ts` covering the per-email throttling behavior.

## How to test

1. On the login page, submit a wrong password for the same email 5 times — you should see the normal "Invalid email or password" (or 2FA variant) each time.
2. Submit a 6th time with the same email — you should see "Too many login attempts. Please wait 15 minutes and try again."
3. A different email is unaffected.
4. `npx jest tests/auth-rate-limit.jest.ts` passes.

## Ticket

CC-875

## Notes

- Per-email keying only, per the ticket. This means an attacker who knows a victim's email could transiently lock out that victim's real logins by repeatedly guessing their password — combining with IP is a reasonable follow-up, left out of scope here.
- Window/threshold (5 attempts / 15 minutes) chosen to align with common brute-force-throttling guidance rather than the ticket's literal "5/minute" example, since a 1-minute window is trivially bypassed by pacing attempts.


lemilonkh (approved):
Great implementation of this, thanks 👍 ✨

PR #3148, opened 2026-09-15
CC-907: Surface live CNB reasoning summaries and fixing whole document edits

piotrnowakowski:
## Summary

CNB requests now stream available provider reasoning summaries alongside workflow progress. Reasoning opens during generation with consistent italic text at 65% opacity and green Clima accents, and disappears on completion or failure. Summaries remain separate from final answers and persisted chat history.

Document-wide edits previously failed because model-generated offsets did not match repeated text. A bounded document agent now searches a fixed draft snapshot, reads chapter context, and selects server-issued matches. Tools resolve exact offsets and return invalid selections for correction. Independent semantic review checks affected chapters. Rejected candidates return to the editor with indexed objections for up to two repairs, each followed by fresh review; only explicit acceptance writes revisions. Protected markers, locked chapters, and template headings have persisted exclusion notices. Each editor attempt is limited to 12 turns, and all attempts and reviews share one 180-second deadline; an SSE end without a terminal event restores controls through the error path.

## Validation

- Updated recording (18 September): a real semantic rejection recovered after one automatic repair; 38 changes across all 12 chapters were accepted and persisted after reload, with 55 protected-marker matches preserved and the original note unchanged.
- Latest backend validation: 239 CNB/progress tests passed, five opt-in tests skipped. Focused reasoning presentation tests and component lint passed.

- Local real-provider rename: 38 editable occurrences across 12 chapters accepted and persisted; 55 protected-marker occurrences preserved. Repeated acceptance created no extra revisions.
- Kraków audit: Stage IV/KST IV to Stage V/KST V changed 28 occurrences across all 12 chapters after acceptance and reload; 42 protected-marker occurrences were preserved and disclosed.
- Summary-only date correction persisted without changing the other chapters. Rejecting a small wording proposal left the saved draft unchanged. Pending proposals survived reload; offline submission recovered with an error; absent text produced no fabricated edits.
- Focused checks rerun before publication: 47 backend tests and 17 frontend tests passed. Frontend coverage was disabled for this focused subset; this is not a full regression or coverage pass.
- Desktop and phone browser checks passed. Full TypeScript checking was previously blocked by the unchanged `app/src/lib/analytics.ts:146` call to `posthog.identify(undefined, properties)`.

See `climate-advisor/docs/cnb-reasoning-validation.md` for the detailed evidence and limitations. Local recordings and fixtures remain in ignored audit directories.

## Rollout and scope

Apply CNB migration `20260917_120000` with `alembic -c cnb-alembic.ini upgrade head` from `climate-advisor/service` before running the updated service. No new credentials are required. README and architecture documentation describe the edit tools, bounds, and persistence contract.

This remains a draft targeting develop directly. CC-827 transport heartbeats are separate; this PR does not establish durable reconnect/resume, worker restart recovery, or deployed ALB behavior. Review/export feedback and remaining proposal presentation issues are tracked separately in [CC-933](https://linear.app/openearth/issue/CC-933), outside this ticket.

Ticket: [CC-907](https://linear.app/openearth/issue/CC-907).

https://github.com/user-attachments/assets/617979f2-b3a9-4188-aecf-aecf55fefadd


climate-advisor/service/app/utils/cnb_progress.py
```diff
@@ -0,0 +1,216 @@
...
+            if getattr(item, "type", "") == "reasoning":
+                for index, part in enumerate(getattr(item, "summary", []) or []):
+                    await emit_cnb_reasoning(
+                        _summary_done(item.id, index, getattr(part, "text", None)),
+                        stream_id=stream_id,
+                        stage=stage,
+                        chapter_title=chapter_title,
+                    )
+        return
+    is_delta = kind in {
```
DavidGBbr, line 104: [P1] Keep raw reasoning out of the summary stream

The PR contract says this UI exposes provider reasoning summaries, but this branch also forwards `response.reasoning_text.delta` and `.done`. In the installed Agents SDK, Chat Completions `delta.reasoning_content` becomes summary events, while third-party `delta.reasoning` becomes `reasoning_text`; OpenRouter documents that field as internal reasoning steps. Because the editor/reviewer receives private draft and source context, this can expose raw reasoning or internal instructions rather than the bounded summary promised here.

Please restrict this boundary to summary event types (plus completed `item.summary` snapshots) and add a negative regression proving `response.reasoning_text.delta`/`.done` emit no browser SSE while summary reconciliation still works.

climate-advisor/service/cnb_migrations/versions/20260917_120000_add_edit_notices.py
```diff
@@ -0,0 +1,32 @@
...
+
+Revision ID: 20260917_120000
+Revises: 20260909_120000
+"""
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+revision = "20260917_120000"
```
DavidGBbr, line 11: [P1] Keep the migration chain green

Adding this revision correctly advances the Alembic head, but the required Climate Advisor job is now red: `tests/test_cnb_migrations.py` has three PostgreSQL failures because it still expects `20260909_120000`; CI reports 238 passed and 3 failed, then skips image publication/deployment.

Please update the migration-chain expectations to `20260917_120000` and extend the upgrade/downgrade coverage to assert that `concept_note_edit_proposals.notices` is added/removed without losing existing proposal rows or the two-parent merge behavior. The full CNB database/runtime job should be green before merge.

DavidGBbr (changes requested):
Blocking on two correctness/release issues:

1. The SSE adapter currently forwards raw `response.reasoning_text.*` events in addition to provider summaries, which exceeds the PR's summary-only contract and can expose internal reasoning from document-aware workers.
2. The required Climate Advisor CI job is red because the new Alembic head was not reflected in the migration-chain tests (3 failed, 238 passed); publication and deployment are skipped.

Non-blocking documentation warning: the README Configuration section still says CNB chat uses medium reasoning and that CNB chat/source workers retain Chat Completions, while `llm_config.yaml` sets CNB chat to high and the implementation selects Responses for CNB chat/source workers. Please remove the contradictory legacy claims while addressing the blockers.

Focused verification at head `3a26c0040`: 66 backend tests passed, 16 frontend tests passed, targeted ESLint passed, Python compileall passed, and `git diff --check` passed.

DavidGBbr (approved):

climate-advisor/service/app/utils/cnb_progress.py
piotrnowakowski: Thanks, David. I’ve clarified [CC-907’s acceptance criteria](https://linear.app/openearth/issue/CC-907): we intentionally want to display provider-returned reasoning traces, including plaintext reasoning as well as summaries. The previous summary-only wording didn’t reflect that intent.

I’d therefore keep `response.reasoning_text.*` support. This doesn’t establish that provider output is sanitized, but accepting those events alone also doesn’t demonstrate an actual leak. Do you have a concrete example from the current setup that exposes internal prompts or information the user shouldn’t see? We should address any such case directly.

climate-advisor/service/cnb_migrations/versions/20260917_120000_add_edit_notices.py
piotrnowakowski: Fixed in ddce3660f: migration-chain expectations now target 20260917_120000. PostgreSQL coverage verifies that upgrading and downgrading notices preserves every existing proposal field, while the two-parent merge tests remain intact. The Climate Advisor CI test job, including the CNB database/runtime suite, now passes: https://github.com/Open-Earth-Foundation/CityCatalyst/actions/runs/35530531993/job/106130209913.

PR #3159, opened 2026-09-16
CC-866: Add CNB-only manual population entry

piotrnowakowski:
## Summary

CC-866 lets a Concept Note Builder user enter a population amount and year when city population data is unavailable. The value persists on that concept-note run and does not update CityCatalyst city data.

## Changes

- Add an authenticated CNB population update API and a Context tab form to save, edit, or remove the value.
- Pass the user-entered value to CNB chat and chapter drafting with its provenance, and cover the save path with tests.

## How to test

1. Open a concept note for a city without population data, then open **Context**.
2. Enter a population and year, save, and reload. Confirm the value remains filled in.
3. Confirm the city's CityCatalyst population record is still empty.

## Ticket

CC-866

## Notes

The attached browser demo shows `480,000` and `2024` saved and filled after reload on a disposable local run. An unrelated failed edit-proposal warning from that fixture is visible in the recording.

https://github.com/user-attachments/assets/41fcfb87-860b-48cd-8403-0d36e97131de


piotrnowakowski:
@DavidGBbr This PR is ready for your review. I also addressed the population edit race during chapter drafting in commit 654ecbe51.

DavidGBbr (approved):
Reviewed at `654ecbe51`. **Approve**, with one non-blocking note.

The CNB-only population path looks correct: authenticated proxy, run-scoped `context_summary.manual_population`, no write to CityCatalyst city data, `user_entered` provenance in chat/drafting, and `409` + row lock while chapter drafting is running. Save/clear/invalid input coverage is in place and CI is green.

### Non-blocking
The new Context-tab strings (`population-cnb-only` through `population-draft-running`) are English-only. de/es/fr/pt still have the older `population` / `population-unavailable` keys, so those locales fall back to English. Fine to merge as-is; follow-up is `npm run i18n:update` (or hand-add the 14 keys) if we want those languages on this form before users hit it.

PR #3162, opened 2026-09-16
CC-873: Add visual funder and template selection for concept notes

piotrnowakowski:
## Summary

Let Concept Note Builder users browse all database funders, inspect programme and application-template details, and save funding to a run. Adds a compact **Change** button in the funder card’s upper-right corner.

## Changes

- Searchable funder catalogue with programme, eligibility, template chapters, and required-field previews.
- Authorized catalogue and selection APIs, with ownership checks and conflict/busy guards.
- Preserve draft text when changing funding; invalidate previous confirmations, validation, and pending edits for review.
- Frontend/backend regression coverage and architecture documentation.

## How to test

1. Open a concept note → Context → Browse funders or Change.
2. Search by funder/programme; inspect profile and template details.
3. Save a selection and reload; confirm the funder and template persist.
4. On a drafted note, confirm changing funding requires acknowledging re-review.

Validation completed during implementation: 17 Python tests, 10 Jest tests, TypeScript checking, and mocked browser coverage. The attached real-stack demo verifies three local funders, template previews, saving, reload persistence, and button placement. The button-label test selectors were updated afterward; that mocked suite was not rerun for the label-only change.

## Ticket

CC-873 (Linear)

## Notes

The attached 62-second video uses a disposable local Krakow note and real APIs/database. Only the initial 175 seconds of startup were trimmed; no separate takes were spliced. Local development-server memory restarts and startup timeouts occurred before the successful flow; catalogue loading remains visible in the clip.

Local database restoration, researched seed records, credentials, and recording artifacts are not included in the code changes. Seeded programme templates are working outlines, not official submission-form replicas; the referenced NFOŚiGW call is explicitly closed.

https://github.com/user-attachments/assets/abab1d43-7892-4193-8f5f-a9ecd84aba73


app/src/components/ConceptNoteWorkspace/index.tsx
```diff
@@ -462,7 +465,11 @@ export function ConceptNoteWorkspace({
                     <Button
                       size="xs"
                       variant="ghost"
-                      onClick={() => setTab("context")}
+                      onClick={() => {
+                        setTab("context");
+                        if (applicationContext) setFundingOpen(true);
```
cephaschapa: If context isn’t loaded yet, this only refetches and never opens the dialog. After a successful refetch the user has to click again. Prefer open after refetch resolves, or always open and let the dialog show its own loading state.

app/src/components/ConceptNoteWorkspace/funding-selection-dialog.tsx
```diff
@@ -0,0 +1,550 @@
...
+            </Button>
+            <Button
+              size="sm"
+              loading={saveState.isLoading}
+              disabled={
+                !changed ||
+                isLoading ||
+                isError ||
+                busy ||
+                (hasDraft && !acknowledged) ||
```
cephaschapa: This disables Save whenever a draft exists, even if the ack checkbox isn’t shown yet (hasDraft && changed at L464). Harmless today because !changed already disables Save — but it’s easy to break later. Prefer (hasDraft && changed && !acknowledged).

app/src/components/ConceptNoteWorkspace/funding-selection-dialog.tsx
```diff
@@ -0,0 +1,550 @@
...
+  whiteSpace: "normal",
+} as const;
+
+/** Search every stored profile and programme; do not hide records without templates. */
+function matchesFundingSearch(
+  funder: ConceptNoteFunder,
+  query: string,
+): boolean {
+  const terms = query.trim().toLocaleLowerCase().split(/\s+/).filter(Boolean);
+  const text = JSON.stringify(funder).toLocaleLowerCase();
```
cephaschapa: Search matches UUIDs and nested template JSON too. Fine for a first version; if the catalogue grows, search name/region/programme fields only.

app/src/components/ConceptNoteWorkspace/funding-details.tsx
```diff
@@ -0,0 +1,269 @@
...
+export function FundingOpportunityDetails({
+  opportunity,
+  lng,
+}: {
+  opportunity: ConceptNoteFundingOpportunity;
+  lng: string;
+}) {
+  const { t } = useTranslation(lng, "concept-notes");
+  const template = opportunity.template;
+  const amount = (value: string | null) =>
```
cephaschapa: Non-numeric award strings become NaN. Guard with Number.isFinite and fall back to the raw string / not-available.

app/src/services/api.ts
```diff
@@ -2490,6 +2494,32 @@ export const api = createApi({
...
+        providesTags: (_result, _error, runId) => [
+          { type: "ConceptNoteApplicationContext", id: runId },
+        ],
+      }),
+      getConceptNoteFundingCatalogue: builder.query<
+        { funders: ConceptNoteFunder[] },
+        string
+      >({
+        query: (runId) => `concept-notes/${runId}/funding-catalogue/`,
+        providesTags: ["ConceptNoteFundingCatalogue"],
```
cephaschapa, line 2506: Not scoped by runId. Fine if the catalogue is global; if it can differ per run later, tag with id: runId.

cephaschapa (approved):
Great work 👍👍👍

app/src/components/ConceptNoteWorkspace/index.tsx
piotrnowakowski: Fixed. The handler now waits for a successful context refetch and opens the dialog without requiring another click. Added regression coverage for successful recovery, failed retries, and already-loaded context.

app/src/components/ConceptNoteWorkspace/funding-selection-dialog.tsx
piotrnowakowski: Updated to `hasDraft && changed && !acknowledged`, matching the checkbox’s visibility condition.

app/src/components/ConceptNoteWorkspace/funding-selection-dialog.tsx
piotrnowakowski: Updated now. Search uses funder names, countries and regions, plus programme names, regions, sectors and summaries. IDs and template JSON are excluded. Added regression coverage for both matching and excluded fields.

app/src/components/ConceptNoteWorkspace/funding-details.tsx
piotrnowakowski: Added a `Number.isFinite` guard. Valid amounts retain localized formatting, nonnumeric values fall back to the original string, and null or blank values show the translated “not available” label. Added regression coverage.

app/src/services/api.ts
piotrnowakowski: Confirmed that the catalogue is currently global. The route uses `runId` to check ownership and city access, then loads the shared reference catalogue without run-specific filtering. Only the saved funding selection is run-specific, so I’ve kept the shared tag.

PR #3163, opened 2026-09-17
CC-878: Bulk-import inventory files for Chilean municipalities

cephaschapa:
## Summary
- Add admin bulk eCRF/near-eCRF zip import (match, enqueue, auto-approve worker, dry-run, create-missing cities, OpenClimate name fallback).
- Keep signed (negative) total CO2e so AFOLU removals survive import.
- Match Chile MEED files (`inventory-CHL-{ine}-{name}-{year}`) by INE, store UN/LOCODE when MEED has one, and skip OpenClimate city lookup for INE-only comunas.
- Enrich INE-only city population from Global API (best-effort); missing population does not fail the file.

## Test plan
- [ ] Admin bulk zip of CRFFormat / eCRF files: match existing cities, create-missing + OpenClimate unique name, dry-run writes no activities
- [ ] Negative total CO2e row imports as a removal (GHGI results)
- [ ] Chile zip `inventory-CHL-{ine}-…-2022.csv`: La Pintana (`CL13112`) creates without OpenClimate; Santiago maps to `CL SCL`
- [ ] Job UI polls; after enqueue, first 8 files import; remaining files need cron `POST /api/v1/cron/process-bulk-inventory-import` locally
- [ ] Failure report / retry (IMP-010) and Chile playbook (IMP-012) still open


app/src/backend/BulkInventoryImportWorkerService.ts
```diff
@@ -0,0 +1,208 @@
...
+  ): Promise<BulkInventoryImportWorkerResult> {
+    const result: BulkInventoryImportWorkerResult = {
+      jobsTouched: 0,
+      itemsProcessed: 0,
+      itemsCompleted: 0,
+      itemsFailed: 0,
+      itemsSkipped: 0,
+    };
+
+    const items = await db.models.BulkInventoryImportItem.findAll({
```
tatimoreira, line 166: **Two workers can pick up the same file and import it twice.**

In plain terms: this just *reads* the list of "pending" files — it doesn't mark them as taken at the same time. If the cron job (runs every minute) and the "start processing right after upload" call (in the enqueue route's `after()`) both run close together, they can both see the same file as "pending" and both import it before either one gets around to flipping its status to `importing`. That means the same file's data could get written into the inventory twice.

The dedupe check later (looking for an existing `ImportedInventoryFile` with the same content digest) doesn't close this gap either — it's also a read-then-act check, so two workers running at nearly the same time can both pass it before either has finished writing.

**Suggested fix:** make "pick items" and "claim items" one atomic step instead of two, e.g.:
- Wrap the read in `SELECT ... FOR UPDATE SKIP LOCKED` so a second worker skips rows already being read by the first, or
- Do an atomic conditional update instead of a plain read, e.g. `UPDATE ... SET status = 'importing' WHERE status = 'pending' ... RETURNING *`, and only operate on the rows that update actually returns.

Either way, no worker should be able to start on a row that another worker has already selected.

app/src/backend/BulkInventoryImportZip.ts
```diff
@@ -0,0 +1,170 @@
...
+  }
+  if (candidatePaths.length > maxFiles) {
+    throw new BulkInventoryImportZipError("too_many_files");
+  }
+
+  const entries: BulkInventoryImportZipEntry[] = [];
+  for (const path of candidatePaths) {
+    const file = zip.file(path);
+    if (!file) continue;
+    const inner = Buffer.from(await file.async("uint8array"));
```
tatimoreira, line 140: Small heads up, not a blocker: this line fully unpacks each file into memory first, and only checks its size on the next line. So the 20 MiB limit doesn't actually stop a huge file from being unpacked — it just deletes it *after* the damage (the big memory allocation) is already done. A tiny zip file that hides something like a 10 GB file inside (a classic "zip bomb") could crash the server before it ever gets to the size check.

Since this is admin-only, the risk is pretty low. But it'd be an easy fix: JSZip can tell you a file's uncompressed size upfront (via `file._data.uncompressedSize`, or similar) without decompressing it, so you could check the size first and skip decompressing anything too big.

app/src/backend/BulkInventoryImportEnqueueService.ts
```diff
@@ -0,0 +1,489 @@
...
+      const { entry, result, openClimate } = row;
+      const superseded =
+        result.parsed.exportDate != null &&
+        !latestKeys.has(
+          exportKey(result.originalFileName, result.parsed.exportDate),
+        );
+
+      let s3Key: string | null = null;
+      try {
+        s3Key = await InventoryFileStorageService.uploadBulkImportFile(
```
tatimoreira, line 328: Scaling note, not blocking: this whole loop runs one file at a time — upload here, plus `findOrCreateCityAndInventory`/`enrichCityBestEffort` further down for files needing a new city — all inside the single request that has to finish within `maxDuration = 120`. Fine for today's Chile batch, but worth knowing where the time goes if batches grow.

A few ways to speed it up, roughly smallest → biggest change:
- The full zip is already stored as `archive.zip` a few lines up, so this per-file upload is mostly redundant — the worker could extract each file from the stored zip on demand instead of it being re-uploaded individually here.
- `findOrCreateCityAndInventory`/enrichment (the only network-bound part of this loop) could move into the worker's per-item step, same as the actual import already does — keeping enqueue to just matching + row inserts.
- Or, smallest change: wrap the independent uploads in a bounded-concurrency `Promise.all` instead of a sequential loop.

Not asking for this now, just flagging in case Chile's batch size grows.

app/src/backend/FormatAdapterService.ts
```diff
@@ -305,14 +306,13 @@ export default class FormatAdapterService {
...
-  /** Adapter D: has GPC ref + emissions + notation columns. */
+  /** Adapter D: GPC reference + total emissions. Notation is optional. */
   private static isNearECRF(headersLower: string[]): boolean {
     const hasGpcRef = headersLower.some((h) => /gpc.*(ref|reference)/i.test(h));
     const hasEmissions = headersLower.some((h) =>
       /total.*emission|total.*co2e|ghg.*emission/i.test(h),
     );
-    const hasNotation = headersLower.some((h) => /notation/i.test(h));
-    return hasGpcRef && hasEmissions && hasNotation;
+    return hasGpcRef && hasEmissions;
```
tatimoreira, line 315: This one's worth a closer look since it's a behavior change to shared, existing detection logic — not new bulk-import-only code. `isNearECRF` used to require three things to match (GPC ref column + an emissions column + a notation-key column) before treating a file as Adapter D. This PR drops the notation requirement to two, because Chile's MEED CSVs don't have a notation column.

That makes the match looser for *every* file going through this detector, not just Chile files — any spreadsheet (from the regular single-file admin upload too, not just bulk import) that happens to have a column matching `/gpc.*(ref|reference)/` and one matching `/total.*emission|total.*co2e|ghg.*emission/` will now be routed down the Adapter D / near-ecrf path, even if it isn't actually that format. The new tests only cover real Chile/CRFFormat files being correctly detected — there's no test for a file that has those two headers by coincidence but is a different format, so I can't tell from the diff whether that's already ruled out elsewhere (e.g. by `isCIRIS`/`isBIOMATEC`/`isMultiCity` running first) or genuinely more permissive now.

Could this be scoped down instead — e.g. only relax the notation requirement when the file also matches the Chile-specific header set, rather than for every file that reaches this detector?

PR #3164, opened 2026-09-17
fix(CC-901): reset stale logo file state after save

tatimoreira:
## Summary

Fixes the Save Changes button staying permanently disabled after removing a saved org logo in Brand Settings, caused by stale internal file-upload state that wasn't reset after a successful save.

## Changes

- `BrandSettingsTab.tsx`: key `FileUploadRoot` to `organization?.logoUrl` so it remounts (dropping any stale accepted file) whenever the saved logo actually changes.
- `branding/route.ts`: pass `AWS_SESSION_TOKEN` through to the S3 client so temporary/SSO AWS credentials work for local logo-upload testing.

## How to test

1. Go to Settings → Account → Brand Settings.
2. Upload a logo image and click **Save Changes**; confirm it saves successfully.
3. Click the logo's remove/reset control.
4. Click **Save Changes** again — it should be enabled and the save should succeed (previously it stayed disabled).

## Ticket

CC-901


lemilonkh (approved):
Awesome, thanks 👍 ✨

PR #3165, opened 2026-09-18
feat: CC-898 implement recovery codes

lemilonkh:
- Implements generation of recovery codes in route /api/v1/auth/2fa/verify
- Shows recovery codes once on 2FA setup success page (on settings > security tab) with descriptive message and copy to clipboard button
- Allows using longer recovery codes on login page and in authorize method


app/src/app/[lng]/auth/login/page.tsx
```diff
@@ -24,7 +24,8 @@ export type LoginInputs = {
   securityToken: string;
 };
 
-const securityCodePattern = /^[0-9]{6}$/;
+// allow either 6-digit numerical codes or 10-digit hex codes separated by an optional - in the middle
+const securityCodePattern = /^[0-9]{6}|[0-9a-f]{5}-?[0-9a-f]{5}$/;
```
Comment (tatimoreira) [inline app/src/app/[lng]/auth/login/page.tsx:28]: **Broken regex — accepts malformed codes**

`|` has lower precedence than the anchors, so this parses as `(^[0-9]{6})` OR `([0-9a-f]{5}-?[0-9a-f]{5}$)`, not "6 digits or a hex-dash-hex code, in full." Verified with node:

```
/^[0-9]{6}|[0-9a-f]{5}-?[0-9a-f]{5}$/.test("1234567")    // true — extra trailing digit
/^[0-9]{6}|[0-9a-f]{5}-?[0-9a-f]{5}$/.test("123456abc") // true — extra trailing junk
```

Needs the alternation grouped so both anchors apply to the whole pattern:

```suggestion
const securityCodePattern = /^([0-9]{6}|[0-9a-f]{5}-?[0-9a-f]{5})$/;
```

app/src/lib/2fa.ts
```diff
@@ -35,3 +39,45 @@ export const verifyToken = async (token: string, secret: string) => {
...
+  return { recoveryCodes, hashedCodes };
+};
+
+// Check hash of recovery code against remaining hashes in user record and remove if successful
+export const verifyRecoveryCode = async (user: User, recoveryCode: string) => {
+  // re-add middle dash in case it was not entered by user to make the hash still match
+  if (!recoveryCode.includes("-") && recoveryCode.length === 10) {
+    recoveryCode = `${recoveryCode.slice(0, 5)}-${recoveryCode.slice(5)}`;
+  }
+  const hash = crypto.createHash("sha256").update(recoveryCode).digest("hex");
```
tatimoreira, line 68: **Case sensitivity: recovery code matching will silently fail on uppercase input**

Generated codes are lowercase hex (`crypto.randomBytes(5).toString("hex")`), but nothing normalizes `recoveryCode` here before hashing/comparing. If a password manager or the user's input method uppercases characters on paste/autofill, a valid code will fail to match — with no indication to the user why (it just looks like an "invalid code" auth failure).

Suggest lowercasing the input before the dash-reinsertion/hash step:

```suggestion
  recoveryCode = recoveryCode.toLowerCase();
  // re-add middle dash in case it was not entered by user to make the hash still match
  if (!recoveryCode.includes("-") && recoveryCode.length === 10) {
    recoveryCode = `${recoveryCode.slice(0, 5)}-${recoveryCode.slice(5)}`;
  }
  const hash = crypto.createHash("sha256").update(recoveryCode).digest("hex");
```

app/src/lib/auth.ts
```diff
@@ -129,10 +133,16 @@ export const authOptions: NextAuthOptions = {
...
             return null;
           }
-          const isValid = await verifyToken(
-            credentials.securityToken,
-            user.twoFactorSecret,
-          );
+          let isValid = false;
+          if (credentials.securityToken.length >= recoveryTokenMinLength) {
+            // allow using a single-use recovery code and delete it from user record if successful
+            isValid = await verifyRecoveryCode(user, credentials.securityToken);
```
tatimoreira, line 139: Nice — this reuses the existing login-path rate limiter (5 attempts/15 min per email, see the `RateLimiter` comment above in this file), so recovery-code brute forcing is already throttled without any extra work here.

app/src/lib/2fa.ts
```diff
@@ -35,3 +39,45 @@ export const verifyToken = async (token: string, secret: string) => {
...
+  const userHashes = user.twoFactorRecoveryHashes ?? [];
+  const codeIndex = userHashes.indexOf(hash);
+
+  if (codeIndex !== -1) {
```
cephaschapa: Small race: two concurrent logins with the same code can both succeed before either save. Unlikely, but a single DB update that removes the hash atomically would be safer.

app/src/app/[lng]/auth/login/page.tsx
```diff
@@ -24,7 +24,8 @@ export type LoginInputs = {
   securityToken: string;
 };
 
-const securityCodePattern = /^[0-9]{6}$/;
+// allow either 6-digit numerical codes or 10-digit hex codes separated by an optional - in the middle
```
Comment (tatimoreira) [inline app/src/app/[lng]/auth/login/page.tsx:27]: Nice, this addresses the case-sensitivity fix in `verifyRecoveryCode` correctly. But there's a gap one layer up: this client-side pattern still only allows lowercase `a-f`, and it's wired into react-hook-form's `pattern` validation on the security-token input below. If a user pastes an uppercased recovery code (autocapitalize, a password manager, etc.), this regex rejects it client-side with "security-code-invalid" before the request ever reaches the server — so the server-side lowercasing fix never gets exercised for that case.

Suggest either adding the `i` flag here:
```suggestion
const securityCodePattern = /^([0-9]{6}|[0-9a-f]{5}-?[0-9a-f]{5})$/i;
```
or normalizing the field value to lowercase on change/blur so both layers agree.

tatimoreira (approved):

PR #3170, opened 2026-09-18
fix(invites): don't fail invite creation when the email can't be sent

briansebaOEF:
## Summary
`POST /api/v1/user/invites` returns **500** whenever the invitation email fails to send, even though the invites were created.

The route is meant to treat email as best-effort: it returns the invite URLs so the UI can copy them to the clipboard, and it has an `if (!sendInvite) logger.warn(...)` fallback. But `sendEmail` ([`lib/email.ts`](app/src/lib/email.ts)) never returns a falsy value. It **rethrows** every error. So when delivery fails:

1. The error skips the warning branch and is caught by the outer per-email `catch`
2. The invite is pushed to `failedInvites`, so the route throws `500 Something went wrong`
3. The `CityInvite` rows already exist, but the client never gets `inviteUrls`, and the UI shows "Failed to send invitations"

Found during onboarding, where the local Ethereal SMTP credentials were invalid (`535 Authentication failed`). It would also happen in production on any mail provider hiccup.

## Fix
Wrap `sendEmail` in its own `try/catch` in the route and log a warning (now with the error), matching the intended best-effort behaviour. `sendEmail` itself is unchanged, since other callers may depend on it throwing.

## Tests
New case in `tests/api/user_invites.jest.ts`: mocks `@/lib/email` to throw and asserts a 200, `success: true`, an invite URL for the email, and a persisted `CityInvite` row.
- Fails on `develop` (500), passes with the fix
- Uses `jest.unstable_mockModule` plus a dynamic import of the route, since the suite runs with `--experimental-vm-modules`

## Test plan
- [x] `NODE_OPTIONS=--experimental-vm-modules NODE_ENV=test npx jest tests/api/user_invites.jest.ts`: 2 passed
- [ ] With broken SMTP locally, inviting from onboarding succeeds and the invite link is copied to the clipboard


lemilonkh (approved):
Awesome, thanks 👍

PR #3171, opened 2026-09-21
Draft: Generic file ingestion architecture

cephaschapa:
## Summary
- Draft architecture for a **generic file upload → ingest → commit** spine shared across CityCatalyst modules
- Modules plug in their own formats, taxonomies, and SoT writers (GHGI reference, CCRA greenfield, CNB Path B)
- Aligns with NativeInputCatalog, AgenticModuleScope, and GHGI bulk (CC-878) lessons

## Doc
`docs/GenericFileIngestionArchitecture.md`

## Test plan
- [ ] Stakeholder review of ownership table and pipeline stages
- [ ] Confirm open decisions D1–D6
- [ ] No runtime code in this PR


docs/GenericFileIngestionArchitecture.md
```diff
@@ -0,0 +1,414 @@
...
+  end
+
+  subgraph Plugin["Module Ingest Plugin"]
+    Detect["Detect format"]
+    Parse["Parse"]
+    Adapt["Adapt → canonical rows"]
+    Validate["Validate vs taxonomy"]
+    Policy["Approval policy"]
+    Write["Commit writer → module SoT"]
+    Catalog["Register NativeInputCatalog"]
```
mircorudolph, line 106: Is this only happening here at the end?

If this only happens after module specific parsing, we might lose the actual document content.

E.g. in my head the workflow would be:

1. City uploads a CAP -> ingest workflow (auth, upload, store, parse into markdown)
This content should already be registered into native input catalog in my opinion. Because the "raw" markdown might be useful for later Clima interactions, CNB or whatever

2. Module specfic adapter parses specific info from that CAP e.g. emissions into data model of GHGI
-> this then gets ALSO registered into native input catalog as extracted GHGI values.

This way NativeInputCatalog has 2 entries. The original files (as parsed markdown) for general extraction of data downstream AND the module specific GHGI values.


In this mermaid schema if this is one sequential flow, it might be ok to do both at the end.
Just thinking if there are use-cases where 1 upload will be used for different modules (e.g. CAP for GHGI and CNB) so not 1 single adapter is being used.

Uploaded files should belong to a document pool.

On that note: We were often discussing about a clearer hierachy/ owning/ scope of documents.
E.g.: Is a file uploaded from a user in CNB belonging to that user? Or to that specific CNB draft? Or to that city? Or to the entire org?
Sometimes we scope documents to the user, however to me it does not make sense in a context of CC, where it is not about user data but about city or org data and multiple users work togehter on 1 inventory, HIAP etc.
@joaquinOEF I am not sure if we have full clarity about such things but that has been important e.g. in how we scope access within Clima and NativeInuputCatalog @DavidGBbr @piotrnowakowski

piotrnowakowski: Agree that is a very good oportunity to create some hierarchy document that we can refrence as one source of thruth whenver I want to enable or disable and org or user having access to something

docs/GenericFileIngestionArchitecture.md
```diff
@@ -0,0 +1,414 @@
...
+
+1. **CC Core** owns the shared ingestion lifecycle (bytes, jobs, stages, errors, auth/tenancy hooks).
+2. **Modules** own format detectors, adapters, target taxonomies, and writers into their **system of record (SoT)**.
+3. Output of a successful ingest is always data that **matches the module’s CityCatalyst schema/taxonomy** — never raw vendor spreadsheets as product state.
+4. The same spine serves **UI upload**, **admin/bulk zip**, and **agentic** flows (draft → confirm → commit).
+5. After commit, modules **register** discoverable pointers in `NativeInputCatalog` (existing contract).
+
+### Non-goals
+
+- One mega-table that stores every module’s domain rows.
```
piotrnowakowski: So this feels like native input catalogue or I am missing somehting. In case they are same it should be named here explicitly. If they differ we should clarify how

docs/GenericFileIngestionArchitecture.md
```diff
@@ -0,0 +1,414 @@
...
+  Agent --> Enqueue
+  Enqueue --> Store --> Job --> Runner
+  Runner --> Detect --> Parse --> Adapt --> Validate --> Policy
+  Policy -->|approve / auto| Write --> SoT
+  Write --> Catalog --> NIC
+  Policy -->|needs review| Review["Human or agent review UI"]
+  Review -->|confirm| Write
+```
+
+---
```
piotrnowakowski: I am missing layer of persistance for the user decisions as they might not be done in the same flow. Do we allow it and we will store state ? it might make sense for long ghgi uploads where there is a lot to check for example. That storage should be idealy then shared across the modules\plugins

docs/GenericFileIngestionArchitecture.md
```diff
@@ -0,0 +1,414 @@
...
+| Auth, org freeze, module access, city/project scope | Core + existing Permission/Module services | Plugin declares required scope shape |
+| Format detection / adapters | **Module plugin** | e.g. GHGI eCRF vs CCRA hazard CSV |
+| Canonical intermediate type | **Module plugin** | Typed per module; not one global row type |
+| Taxonomy / schema validation | **Module plugin** | Against CC form-schema / CCRA models / CNB contracts |
+| Approval policy | **Module plugin** | Auto vs review vs agent-draft-only |
+| Commit / replace / versioning | **Module plugin** | Writes module SoT only |
+| NativeInputCatalog registration | **Module plugin** (via existing catalog services) | After durable commit |
+| UI copy, wizards, mapping screens | Module product UI | May share generic progress components |
+
+**Hard rule:** Core never interprets domain columns. If Core needs a “preview,” it asks the plugin for a sanitized summary DTO.
```
piotrnowakowski: For what Core needs preview if the human in the loop/approval is owned by modules ?

docs/GenericFileIngestionArchitecture.md
```diff
@@ -0,0 +1,414 @@
...
+## Pipeline stages
+
+Every item runs the same ordered stages. Stages may no-op when a plugin says so.
+
+| Stage | Core responsibility | Plugin responsibility |
+| --- | --- | --- |
+| 1. **Receive** | Auth, scope, size/ext allowlist from plugin metadata | Declare allowed extensions / max size |
+| 2. **Store** | Persist archive + per-file objects; content hash | — |
+| 3. **Match** (batch only) | Invoke plugin matcher; record unmatched | Filename / manifest → entity ids (city, assessment, …) |
+| 4. **Detect** | Call plugin | Return format id (`ecrf`, `near-ecrf`, `ccra-hazard-v1`, `pdf`, `unknown`) |
```
piotrnowakowski: I think detection should be owned by the core at least for the extension like pdf vs word and passed to the plugin with the file to enable plugin already running the correct extraction piece but I am open to change my mind

docs/GenericFileIngestionArchitecture.md
```diff
@@ -0,0 +1,414 @@
...
+
+---
+
+## Open decisions (product / platform)
+
+| # | Question | Options | Lean |
+| --- | --- | --- | --- |
+| D1 | New generic tables vs `moduleId` on bulk GHGI tables? | New `Ingest*` · extend bulk · both for a while | Extend or dual-write first; avoid big-bang rename |
+| D2 | Where does LLM column-mapping live? | Plugin-only · shared “mapping stage” service | Shared stage invoked only when plugin requests it |
+| D3 | Cross-module zip (GHGI + CCRA in one archive)? | Reject · multi-plugin dispatch | Reject in v1 |
```
piotrnowakowski: So what if that is actually needed across different plugins like CNB will reuse some of the data from CAP upload ? will that be blocked or how we will link it back for CNB to use if CAP owns the conversion ? link file to a city ?

docs/GenericFileIngestionArchitecture.md
```diff
@@ -0,0 +1,414 @@
...
+| GHGI commit | `InventoryImportService` |
+| GHGI bulk | `BulkInventoryImport*` services + models |
+| Auto-import façade | `InventoryFileAutoImportService` |
+| Native catalog | `NativeInputCatalog` + `app/src/backend/agentic/native-input-catalog/` |
+| Agentic registries | `app/src/backend/agentic/**/registry.ts` |
+| CCRA target SoT (proposal) | [CCRAModuleArchitecture](./CCRAModuleArchitecture.md) |
+
+---
+
+## Document history
```
piotrnowakowski: I think we need to also cover the path of deleting the file and at the same time getting read of any mentions of it from native input catalogue any changes or adjustments also. 
Do we want to think about versioning or replacing older documents ?

PR #3173, opened 2026-09-21
Fix MEED ranking storage when actions are legally excluded

carlosgraffi:
## Summary

Real MEED rankings that contained legally removed actions failed to save: `POST /api/v1/city/{id}/meed/rank` returned 409 with `MeedActionRemoved_ranking_id_fkey` and the user saw "Ranking could not be generated". The removed-actions rows were inserted outside the transaction that creates the `MeedRanking` row, so the foreign key pointed at an uncommitted row.

## Changes

- `MeedApiService.ts`: `{ transaction }` was passed as the second argument of `.map()` (its `thisArg`, silently ignored) instead of `bulkCreate()`. Moved it to `bulkCreate`, matching the ranked-actions insert five lines above. Row fields are unchanged.

## How to test

1. Run a real ranking (hiap-meed on `HIAP_MEED_API_URL`) for a city whose legal screening removes at least one action — e.g. Antofagasta `CL ANF`, which has Global API coverage.
2. Before: the rank route logs `ForeignKeyConstraintError … MeedActionRemoved_ranking_id_fkey` and returns 409. After: the run lands on `/results`, and `MeedActionRanked`, `MeedActionRemoved` and `MeedRankSnapshot` all have rows for the inventory.

## Notes

The mock ranking (`MEED_MOCK_RANKING`) never stores anything, which is why this didn't show up locally before. The file already fails `prettier --check` on `develop`; this PR deliberately touches only the affected hunk.


app/src/backend/MeedApiService.ts
```diff
@@ -312,26 +312,28 @@ export default class MeedApiService {
         })),
         { transaction },
       );
+      // `{ transaction }` is bulkCreate's second argument, not map's: passed
+      // to map it is ignored (map's second argument is `thisArg`), the rows
+      // were inserted outside the transaction, and the FK to the uncommitted
+      // MeedRanking row failed on every ranking that had removed actions.
```
lemilonkh: Good explanation, but I think we can remove these comments from the codebase since it's fixed now 👍

lemilonkh (approved):
Great spot, thanks 🙏

app/src/backend/MeedApiService.ts
carlosgraffi: Removed in 891ab05 — the explanation lives in the commit message now.

PR #3178, opened 2026-09-21
fix(CC-926): remove blank space in subsector notation key dropdowns

tatimoreira:
## Summary
- Fixes a CSS layout bug in the notation key dropdowns on the Notation Keys page where a blank gap appeared before the option description (most visible on the "NE" option).
- Root cause: `SelectItem`'s flex row (`justify-content: space-between`) treated the label and description as separate flex children, spreading them apart instead of letting them flow together. Wrapping both in a single `<Text as="span">` fixes it.

Fixes CC-926

## Test plan
- [x] Verified in browser: opened an inventory's Notation Keys page, opened a subsector's Notation key dropdown, confirmed all four options ("NE", "NO", "CI", "IE") render with label + description flowing together and no blank gap.


lemilonkh (approved):
Thanks!

PR #3179, opened 2026-09-21
CC-872: Close concept-note deletion dialogs and handle stale notes

piotrnowakowski:
## Summary
Close the concept-note deletion dialog before showing success. When another tab has already deleted the note, verify its absence through a fresh authorized run list, close the stale dialog, and show an informational notice instead of an endless “Please try again” error. Unverified absence and genuine failures keep the error dialog open.

## How to test
1. Delete a concept note and confirm that the dialog closes automatically and the card disappears.
2. Open Delete for the same note in two tabs. Confirm in one, then the other; the second dialog should close with “This concept note is no longer available.” Reload and confirm the note stays absent.

Validation: 7 Playwright browser integration tests passed against the local app in Chromium; scoped ESLint and `git diff --check` passed. The isolated component Jest test has been removed. Browser coverage includes Cancel, successful deletion and reload, a stale confirmation across two tabs, 404 with a still-present note or failed authorized list, and 403/409/503 failures followed by a successful retry. API responses are controlled; this suite does not verify backend source or artifact cleanup. The existing recordings below separately show the local CC/CA flows.

To run the browser suite, start the app with `CA_SERVICE_INTEGRATION,CONCEPT_NOTE_BUILDER` enabled, set `CNB_TEST_URL` to its URL and `CNB_AUTH_STATE` to an authenticated Playwright storage-state file, then run from `app`:

```sh
npx playwright test --config e2e/deletion.playwright.config.ts
```

## Ticket
[CC-872](https://linear.app/openearth/issue/CC-872)

## Notes
Fixed: deleting a concept note closes the confirmation window automatically. If the note has already been deleted in another tab, the stale confirmation also closes and the misleading "Please try again" message no longer appears or reappears. Both behaviors were verified in the attached videos.

https://github.com/user-attachments/assets/05c19d62-947c-4450-ad41-bd3b9e6a37cb

https://github.com/user-attachments/assets/04a3ed75-bb4e-409f-9fb9-b742572ab0c3


app/tests/concept-note-lifecycle-dialog.jest.tsx
```diff
@@ -0,0 +1,157 @@
...
+  globalThis.IS_REACT_ACT_ENVIRONMENT = true;
+  globalThis.structuredClone = (value) =>
+    value === undefined ? value : JSON.parse(JSON.stringify(value));
+  globalThis.ResizeObserver = class {
+    observe() {}
+    unobserve() {}
+    disconnect() {}
+  };
+  ({ ConceptNoteLifecycleDialog: Dialog } =
+    await import("@/components/ConceptNoteDashboard/lifecycle-dialog"));
```
lemilonkh: We rarely test frontend components in isolation in unit tests like this (usually we do integration tests with Playwright for these as it's easier to verify the functionality works as expected on the finished page).

But fine to have nonetheless.

lemilonkh (approved):
Great, thanks 👍 ✨

app/tests/concept-note-lifecycle-dialog.jest.tsx
piotrnowakowski: Replaced the isolated component unit test with Playwright integration tests on the concept-note dashboard and removed the previous Jest test. The seven browser tests cover successful deletion, Cancel, stale confirmations across two tabs, and error/retry handling. All seven pass, and lint is clean.

PR #3180, opened 2026-09-21
fix(CC-901): keep remove-logo button within visible bounds

tatimoreira:
## Summary

Follow-up to #3164: the remove-logo icon button was positioned with `left: 100%` inside an `overflow: hidden` container, clipping it completely out of view. Users could see and save a logo but had no visible way to remove one, matching the regression reported in [CC-901](https://linear.app/openearth/issue/CC-901/cannot-save-after-removing-logo).

## Changes

- Reposition the remove-logo icon button to `top`/`right` offsets so it renders inside the visible bounds of the logo preview instead of being clipped.

## How to test

1. Run `npm run dev`, log in, go to Settings > Account > Brand Settings.
2. Upload a logo and click **Save changes**.
3. Confirm the remove ("×") icon is now visible in the top-right corner of the saved logo preview.
4. Click it and confirm **Save changes** becomes enabled, then save to restore the default branding.

## Ticket

CC-901


lemilonkh (approved):
Thanks!

PR #3181, opened 2026-09-21
fix(meed): show finance opportunities/projects and follow endpoint values

carlosgraffi:
## Problem
In Actions & Plans v2 (MEED), an action's funding opportunities and matched projects never appeared — the expanded finance row always said "No matched funds/projects yet", and the results action drawer had no finance at all.

**Root cause:** the `modules/meed/finance/follow` proxy (zod guard + `MeedGlobalApiService.followLink`) only allowed links starting with `/api/v1/cities/`, but the Global API feasibility rows link to `/api/v1/climate-finance/opportunities|projects`. Every call returned 400 and, with no error branch, rendered as "no data". Nothing was missing upstream.

While verifying, several places turned out to contradict what the endpoints return; those are fixed here too.

## Changes
**Read finance through hiap-meed (per review)**
- Feasibility, opportunities and projects now use the existing `meed/finance/*` routes (hiap-meed), so the finance screens see the same screening/filters the ranking used. The Global API `finance/follow` + `modules/meed/finance/feasibility` proxies, `resolveFinanceLink`, `followLink` and their hooks/test are removed.
- hiap-meed answers 404 when an action has no matched projects; `MeedApiService.getFinanceProjects` maps that to an empty list instead of a 400.
- Opportunities render `current` then `monitor` (≤5 each); projects are capped at 5 by hiap-meed.

**Finance in the drawer**
- `FinanceMatches` (extracted from `RowDetail`) is shared by the finance row and a new `ActionFinanceSection` in the results `DetailPanel` (route, reason, top funds/projects, link to Finance). Adds an explicit load-error state.

**Follow the endpoint's values**
- Routes map one-to-one to the five API routes (previously `self-deliverable` + `own-budget feasible` were merged, and `needs technical assistance` showed as "Unclassified"). Summaries/insights updated accordingly.
- City profile uses the upstream archetypes (Self-sufficient / Delivery-ready / Well-resourced / Support-ready); null profile claims nothing. Meters only say Higher/Lower, which is all the banding tells us.
- Fund count label follows `inputs.finance.fund_access` (direct / competitive).
- Project cost scales by `amount_unit` (`CLP_thousands` was formatted as millions → 1000× too large).
- Opportunity statuses `periodic` / `emerging` / `in_rollout` are shown instead of dropped; unknown routes/statuses render the raw value rather than a guessed label.

**Product decisions**
- Finance score uses the standard scale: High ≥0.7 (green), Medium ≥0.55 (yellow), Low ≥0.4 (orange), Very low (red). Adds a `caution` (orange) `MeedTone` from existing tokens. Route tag tones unchanged.
- Removed the "shapes N% of the ranking / feasibility score" notes from the emissions, context, policy, regulations, finance and preferences pages (and the matching tooltip sentences).

## Testing
- `npx tsc --noEmit` (no source errors), ESLint clean on touched files, `npm run jest -- tests/meed-` → 13 suites / 115 tests (`meed-finance-facts` rewritten around the real route strings).
- Browser against local hiap-meed, Chile demo cities: Antofagasta finance row `c40_0037` → `meed/finance/opportunities` + `projects` 200 (9 funds / 5 projects); `c40_0010` (no projects) → 200 with an empty list; results drawer `c40_0016` → Financing section with 6 funds / 5 projects; route chips match the API exactly (Antofagasta 25/77, Temuco 25/43/34, Canela 10/15/4/73).

## Follow-ups (not in this PR)
- Finance is on hiap-meed now; actions, city attributes and policy scores still read the Global API via `modules/meed/*`. Switching those needs shape work (`action_id` vs `src_action_id`, flattened indicators) and a consistent error contract in `MeedApiService.makeRequest`.
- Upstream data question on FPA `amount_committed` units.

EN locale only, per convention.


app/src/app/api/v1/city/[city]/modules/meed/finance/follow/route.ts
Comment (lemilonkh) [inline app/src/app/api/v1/city/[city]/modules/meed/finance/follow/route.ts:1]: Ideally this route's usages are replaced with the direct API routes that query the data via the hiap-meed service. Then this file can be removed.

app/src/backend/meed/financeLinks.ts
```diff
@@ -0,0 +1,29 @@
...
+ * opportunities and projects under `/api/v1/climate-finance/`.
+ */
+const ALLOWED_PREFIXES = ["/api/v1/cities/", "/api/v1/climate-finance/"];
+
+/**
+ * Normalise a relative link and check it against the allow-list. Returns the
+ * encoded `pathname + search` to request, or `null` when the link is absolute,
+ * escapes the allowed prefixes (`..`), or points anywhere else.
+ */
+export function resolveFinanceLink(link: string): string | null {
```
lemilonkh: We can remove this function and file once we use the direct API routes that query the hiap-meed service rather than using the redirect proxy for the global API.

app/src/backend/meed/MeedGlobalApiService.ts
```diff
@@ -71,15 +72,18 @@ export class MeedGlobalApiService {
   /**
    * Follow a relative link returned inside a Global API response
    * (`links.projects` / `links.opportunities` on finance feasibility rows).
-   * Only Global-API city paths are allowed.
+   * Only the paths allowed by `resolveFinanceLink` are followed.
    */
-  public static async followLink(relativePath: string): Promise<unknown | null> {
-    if (!relativePath.startsWith("/api/v1/cities/")) {
+  public static async followLink(
```
lemilonkh: Function can be removed and replaced by using the direct API routes instead of the global API link follow proxy.

app/tests/meed-finance-links.jest.ts
lemilonkh, line 1: Can be removed once we use the direct API routes to query the hiap-meed service rather than using the global API link proxy.

app/tests/meed-finance-facts.jest.ts
lemilonkh, line 1: Can be removed once we use the direct API routes to query the hiap-meed service rather than using the global API link proxy.

lemilonkh (approved):
Thanks! Let's make sure to start using the new API routes for querying the hiap-meed service to ensure the same context and filters are used for both the ranking and the input data.

app/src/app/api/v1/city/[city]/modules/meed/finance/follow/route.ts
Comment (carlosgraffi) [reply app/src/app/api/v1/city/[city]/modules/meed/finance/follow/route.ts:1]: Done in 29b488f7f — `FinanceMatches` and the feasibility reads now use the hiap-meed-backed `meed/finance/{feasibility,opportunities,projects}` routes; this file and the `modules/meed/finance/feasibility` proxy are removed.

app/src/backend/meed/financeLinks.ts
carlosgraffi: Removed in 29b488f7f (no more link following).

app/src/backend/meed/MeedGlobalApiService.ts
carlosgraffi: Removed in 29b488f7f, together with `fetchFinanceFeasibility`.

app/tests/meed-finance-links.jest.ts
carlosgraffi: Deleted in 29b488f7f along with `resolveFinanceLink`.

app/tests/meed-finance-facts.jest.ts
carlosgraffi: Kept this one: it doesn't touch the proxy — it pins the route/profile/fund-access/amount-unit label mapping to the five real route strings hiap-meed returns, which is still what the finance screens render.

PR #3184, opened 2026-09-21
fix(CC-932): derive NavigationBar locale flag from request-scoped lng

tatimoreira:
## Summary
- `NavigationBar` read the global `i18next.language` singleton (`src/i18n/client.ts`) directly to render its locale flag/label, instead of the request-scoped `activeLng` it already derives from route params.
- Because that i18next instance is initialized once per Node.js server process, concurrent SSR requests resolving different locales could leak state into each other, producing a server/client hydration mismatch (flag rendered ES on the server, EN on the client, hydration bails and the page hangs on the loading spinner).
- Fix: swap both `i18next.language` reads in the flag/label render for the existing `activeLng` variable, which is derived purely from that render's own route params — no shared mutable state left in the path.

Fixes [CC-932](https://linear.app/openearth/issue/CC-932/fix-react-hydration-error-in-navigationbar-locale-detection).

Note: the same `i18next.language`-in-render pattern exists in a handful of other files (signup page, PreferencesTab, methodologies page, HIAP tab/page, PlanDetailsBox, ModalPublish, PublishedView). Left out of scope for this fix and logged as a comment on CC-932 for a possible follow-up.

## Test plan
- [x] `npx eslint src/components/navigation-bar.tsx` — clean
- [x] `npx tsc --noEmit` — no errors in this file
- [x] Local dev server: opened `/en/...` and `/es/...` dashboard tabs concurrently against the same running server, confirmed each renders its own correct flag/label (`us.svg`/EN vs `es.svg`/ES) with zero hydration errors in console
- [x] Fired 15 rounds of parallel (`Promise.all`) SSR fetches to `/en/...` and `/es/...` against the live dev server to stress the concurrency path


tatimoreira:
Follow-up for the other files with the same `i18next.language`-in-render pattern (noted in the PR description) is tracked as [CC-939](https://linear.app/openearth/issue/CC-939/audit-remaining-i18nextlanguage-singleton-reads-outside-navigationbar).

lemilonkh (approved):
Great, thanks 👍

PR #3187, opened 2026-09-22
CC-806: Reduce CNB polling and support longer document edits

piotrnowakowski:
## Summary

Reduce repeated CNB workspace requests without changing API rate limits. Replace recurring browser resource polling with authenticated SSE observation and share active observations, while allowing document edit planning up to five minutes and 100 searches per proposal.

## Changes

- Share adaptive upstream observation by user/run/resource within each server process; back off from 10 to 20/30 seconds for unchanged active work and honor Retry-After.
- Share matching observations across browser tabs using Web Locks/BroadcastChannel, with independent heartbeats and bounded upstream reads.
- Recover failed initial resource reads through SSE cache hydration, seed draft-start responses, recover missing proposal collections, and refresh the run after drafting completes.
- Preserve upstream Retry-After during SSE authorization and configure the draft-search budget centrally.
- Raise the edit planner deadline from 180 to 300 seconds and the search cap from 30 to 100; update configuration docs and include CNB operation architecture notes.

## How to test

1. Open the same CNB workspace in multiple tabs and inspect requests while idle and during draft/upload/edit processing. Stable state should not trigger recurring resource polling.
2. Request a full-document rewrite of a 12-chapter note, review and accept the proposal, then reload to confirm persistence.

Validation: 71 focused frontend tests and 43 backend edit-session/review tests passed after review fixes; scoped ESLint, formatting checks, and git diff --check passed. Regression coverage includes failed initial reads, proposal cache recovery, draft-start hydration, terminal draft reconciliation, pre-stream Retry-After through the real API wrapper, and configurable search limits. The recorded local real-provider retry completed its stream in 248 seconds, proposed 12 changes across 12 chapters, and persisted accepted edits after reload.

## Ticket

[CC-806](https://linear.app/openearth/issue/CC-806)

## Notes

The recording uses a real authenticated local stack, cloned databases and real LLM calls. It is local evidence, not deployment verification. Upstream polling remains adaptive and process-local; this does not introduce durable jobs or cross-replica deduplication. Full TypeScript checking remains blocked by the unchanged app/src/lib/analytics.ts:146 error. The successful rewrite is one run, not a guarantee for every long document.

Recording: 00:41 submit; 04:59 accept; 05:20 reload persistence verified.

https://github.com/user-attachments/assets/0fc37b6b-08ff-4298-a3d6-aff280247d2d


app/src/util/types.ts
```diff
@@ -1334,6 +1334,14 @@ export interface ConceptNoteUploadStatusRequest {
   uploadId: string;
 }
 
+export interface ConceptNoteWorkspaceSnapshot {
+  edits?: import("@/util/concept-note-edit-types").EditProposal[];
```
lemilonkh: I think you can probably import this as a type import at the top of the file instead of inline?

lemilonkh (approved):
Amazing to support server-sent events, I'm sure this could be used in other places in the future if we refactor it as reusable functionality! Thanks 😃 🎉 ✨

app/src/util/types.ts
piotrnowakowski: Good call, thanks! Moved `EditProposal` to a top-level `import type` in b529daf.

PR #3189, opened 2026-09-22
fix(meed): balance the context grid and drop hardcoded weight displays

carlosgraffi:
## Changes
**Context card grid** — five cards on a three-column grid gave a 2:1 top row over a 1:1:1 bottom row, so the emissions card looked misaligned. Now a six-column grid at `lg`: the first two cards span three, the last three span two.

**No hardcoded weights on screen** — every displayed weight now comes from the ranking's own `metadata.weights`:
- `readRankingWeights` returns `null` instead of falling back to the mirrored `scoring_config.py` constants.
- Score legend, composition formula (`score × weight = …`), the "drives the ranking" insight and the "Scoring weights" config row render only when the ranking reported its weights; otherwise pillar scores show without a formula.
- Pre-flight sliders drop "Default: 55%" / "Default weights follow the standard methodology: …"; the slider starting position is the one remaining mirror of hiap-meed's defaults (it exposes no endpoint for them).
- `scoringWeights.ts` shrinks to that slider seed; step-contribution tables and helpers removed.

## Testing
`tsc` clean, MEED jest 13 suites / 116 tests (adapter test covers the no-weights case), pre-existing ESLint `set-state-in-effect` errors untouched. Checked on Antofagasta: grid balanced at 1400px; results page still shows the formula (from `metadata.weights`); pre-flight no longer prints default percentages.


lemilonkh (approved):
Thanks!

PR #3190, opened 2026-09-22
CC-826: Store structured document artifacts and full unverified visual annotations

DavidGBbr:
## Summary
- New PDF OCR jobs persist `combined_markdown.md` and `document.structured.json` from one validated Mistral response, with annotation mode stored on the job.
- CNB PDF uploads request `visual_context`. GHGI and native Markdown stay `none`. A requested annotation that is missing or invalid retries with the existing OCR policy and fails the upload explicitly after exhaustion.
- Climate Advisor delivers the **complete** unverified image-annotation envelope (`source: image_annotation`, `quantitative_reliability: unverified`, unchanged `provider_annotation`) through projection, context bundle, source-query tool, and runtime chat context. It does **not** redact numbers or filter to a closed vocabulary.
- Visual annotations remain untrusted descriptive context only: never instructions, citations, source excerpts, evidence, or trusted quantitative inputs. Exact excerpts and citations come only from source Markdown.
- `VISUAL_CONTEXT_CONTRACT_VERSION` invalidates legacy filtered cache rows; an authorised refresh rehydrates the full envelope from the structured artifact without re-running Markdown LLM analysis. Ready bundles with a stale visual contract are no longer treated as `already_current`.

## Test plan
- [x] Focused app Jest: structured document, Mistral OCR, PDF OCR queue/delivery/migrations, structured read route (full-payload equality).
- [x] Climate Advisor pytest: visual context, context bundle service/persistence (including `begin_build` stale visual), source tools, prompt projection.
- [x] `npx tsc --noEmit` and `git diff --check`.
- [ ] Independent re-review of ready-bundle invalidation + payload equality after this head.
- [ ] Piotr acknowledgement of Clima capability/guardrails before rollout.


DavidGBbr:
@piotrnowakowski Could you review the Clima/Climate Advisor consumer contract in this PR? Mirco asked us to align with you on the new annotation capability and the guardrails for current and future Clima prompts.

This PR stores the structured Mistral OCR artifact alongside Markdown. CNB opts into image/chart annotation; the GHGI extraction path remains non-annotated. The full annotation, including numeric text, reaches downstream consumers with `source: image_annotation` and `quantitative_reliability: unverified`.

Could you confirm the following from the Clima side?

1. Prompts and tools treat visual annotations as unverified descriptive context. They must not follow instructions found in an annotation or use its values for calculations, quantitative analysis, exact claims, evidence, or citations without independent validation.
2. For GHGI inventory questions, Clima selects accepted CityCatalyst inventory values as the factual source. Uploaded Markdown and visual annotations can provide supplementary context but must not override those accepted values. Please identify any Clima route or prompt that could currently promote annotation content to a trusted fact.
3. If broader source routing across modules is still missing, should that be tracked as separate follow-up work, or is there a concrete blocker for enabling this PR's capability?

The synthetic municipal PDF check illustrates the risk: text-table OCR matched 9/9 fields, while the chart annotation matched only 4/10 series points. For example, it assigned the 2022 target value (121,000 tCO2e) to the actual series and the actual value (115,000 tCO2e) to the target series. It also said the 2024 target was missed although the chart shows actual emissions below the target pathway. The application preserved the provider annotation unchanged and retained the `unverified` metadata, so this is a provider interpretation error that downstream consumers must handle safely.

Is the Clima capability and guardrail contract acceptable to you? Please call out any missing prompt/tool/source-routing path or rollout blocker.

PR #3193, opened 2026-09-22
CC-940: Align CNB file readiness and deletion guidance

piotrnowakowski:
## Problem and behavior

In CNB, a PDF could show Ready in Context while Clima still displayed "File received — processing" and had not assembled its chat context. The Context subtitle and badge now remain Processing until that context is ready, and show a failed state when context preparation fails.

Deletion also surfaced a generic error when source delivery was still active. Climate Advisor now preserves the upstream 409 so the UI tells users to wait and retry. S3 batch-delete failures log the failed object error codes for diagnosis; they still stop deletion so cleanup references are preserved. The separate dev permission issue is tracked in CC-940.

## Verification

- 39 scoped web Jest tests passed; scoped ESLint passed.
- 43 Climate Advisor pytest tests passed.
- After the IAM policy was updated outside this PR, the previously failing disposable dev note `test` deleted with HTTP 204 and remained absent after browser reload. This confirms the live permission/deletion path, not deployment of this PR.
- Full browser checks for the new status and 409 messaging remain for post-deploy validation, tracked in CC-940.


tatimoreira:
Found one issue in `context-tab.tsx`: when a file is still being converted (raw status `"processing"`), the label now shows "Processing" instead of "Converting". This is because the code reuses the plain word `"processing"` for two different things — the real "still converting" state, and the new "converted, but chat context not ready yet" state — and then can't tell them apart when picking the label. Before this PR, "still converting" correctly showed "Converting"; now it always shows "Processing".

Not a crash, just a small step backward in wording — worth a quick fix (e.g. a separate flag instead of reusing the string `"processing"` for both cases) before merge.

tatimoreira (approved):
Solid, well-tested fix for the CNB status/deletion mismatch. Left one comment on a small label-collision issue in context-tab.tsx (Converting vs Processing) — not blocking, worth a follow-up.

PR #3197, opened 2026-09-23
fix(CC-942): convert fuel-sales tonnes to m3 via density

cephaschapa:
## Summary

- Fixes CC-942: fuel-sales activity data entered in tonnes produced 0 emissions because `units-tonnes` was missing from the mass-to-volume density path, so conversion returned NaN and the formula silently collapsed to 0.
- Includes tonnes in density-based conversion, throws on unsupported/null conversion factors, and fails loudly when volume-based activity amounts are non-finite.
- Adds regression tests covering tonnes/kg to m3 conversion and matching emissions for the same gasoline quantity in m3 vs tonnes.

## Test plan

- [ ] `npx jest --testPathPattern=unit_conversion_fuel_sales`
- [ ] Enter fuel-sales activity in tonnes (e.g. gasoline) and confirm emissions match the equivalent cubic-meters entry
- [ ] Confirm invalid/unsupported unit conversion returns a clear 400 instead of saving 0 emissions

Closes CC-942


lemilonkh (approved):
Amazing solve, thank you! ✨ 🎉

PR #3198, opened 2026-09-23
CC-863: Model-proposed next questions in concept note chat

piotrnowakowski:
## Summary

The concept note chat now shows two short, model-proposed next questions directly above the **Ask Clima** input. Selecting one fills and focuses the composer. Suggestions never send messages or edit the document, and they refresh after each reply, tab change, or draft revision.

## Changes

- **Climate Advisor:** new `POST /v1/concept-notes/{run_id}/chat/suggestions` endpoint. It uses the `cnb_chat_suggestions` model (`openai/gpt-5.6-luna` via OpenRouter, medium reasoning) with the prompt in `prompts/cnb/chat_suggestions.md`. It returns exactly two questions, each up to 80 characters.
- **Bounded context:** the model sees the first 20k `o200k_base` tokens of the created document, up to six recent messages (2k tokens each), and compact workspace metadata. Uploaded source bodies and internal IDs are excluded. The call has a 20 s deadline, no retries, and a 4,096-token completion cap. Provider failures and invalid output return an empty list.
- **CityCatalyst:** new `/api/v1/concept-notes/[runId]/chat/suggestions` proxy route. The new `ChatSuggestions` component waits briefly before requesting, caches results per workspace state, cancels stale requests, and falls back to two translated deterministic questions.

## How to test

1. Open an existing concept note with a created document and start a fresh chat.
2. Confirm that two short questions appear above the Ask Clima input, outside the scrolling history.
3. Click one: the composer fills and gets focus. Send it, and confirm that two new questions appear after the answer.
4. Switch tabs (Draft / Structure / Context) and confirm that the suggestions refresh.

## Browser recording

https://github.com/user-attachments/assets/9a2da0cd-ce71-4ce7-a1b7-d8773ea16975

The 129-second recording used real local authentication, the backend, and OpenRouter calls, with no mocks. At 00:29 two short questions appear above Ask Clima. At 00:35, 01:11 and 01:34 a suggested question is selected, and each answer is followed by two refreshed questions. Four HTTP 200 suggestion responses each returned exactly two questions (longest: 54 characters). The recorder reported no findings and no console or network issues.

## Ticket

[CC-863](https://linear.app/openearth/issue/CC-863)

## Notes

- Deploy Climate Advisor before the frontend. The frontend falls back to deterministic questions if the endpoint is unavailable. No database migration.
- Not covered by the recording: production, new document generation, edit acceptance, export, mobile layout, and persistence after reload.
- Kept as a draft for CI and review.


DavidGBbr:
Thanks — one blocking issue I found:

`ChatSuggestions` aborts its browser fetch when the workspace changes, but the new proxy route does not forward `req.signal` to `callAuthorizedConceptNoteApi`, even though that helper already supports forwarding a signal upstream. As a result, once the debounce has fired, a tab/revision/chat change only hides the stale result; Climate Advisor/OpenRouter still runs for up to 20 seconds and may consume the full completion budget.

Could we propagate cancellation through the proxy and add a regression test that aborts an in-flight request and verifies the upstream/model request is cancelled?

DavidGBbr (approved):

PR #3199, opened 2026-09-23
fix(hiap-demo): move the Brazil demo under /public/hiap/ to avoid the [inventory] route

carlosgraffi:
## Why

On the deployed develop build `/en/public/hiap-brazil-demo` returned the 404 page: the segment sits next to the dynamic `/[lng]/public/[inventory]` route and was resolved as an inventory id (thanks @lemilonkh for spotting it). Locally the static segment won, which is why it went unnoticed.

## What

- The whole demo tree moves from `app/src/app/[lng]/public/hiap-brazil-demo/` to `app/src/app/[lng]/public/hiap/brazil-demo/` (git rename, no content changes).
- `DEMO_ROOT` in `_lib/hrefs.ts` and the test imports follow the new path.
- New URL once deployed: `https://citycatalyst.openearth.dev/en/public/hiap/brazil-demo/` (PT: `/pt/…`).

Feature flag (`HIAP_BR_DEMO`, dev only) and the `/[lng]/public/` auth bypass are unchanged.

## Checks

`tsc` clean, demo tree lints with 0 errors, fixture tests pass (24).


lemilonkh (approved):
Thanks!

PR #3201, opened 2026-09-23
feat(hiap-demo): in-app reviewer feedback form with screenshots

carlosgraffi:
Follows #3199 (the route move, merged).

## What

A feedback tool for the Brazil Phase 3 review that needs no new infrastructure and no account for reviewers:

- **`/{lng}/public/hiap/brazil-demo/feedback`** — public page (flag-gated with the demo): name and organisation, the ten reviewer questions, and any number of remarks with screen ID, type, priority, section, comment, suggestion and screenshots (file picker or paste into the comment box). EN and PT. Linked from the demo banner as "Feedback form".
- **`POST /api/v1/hiap-demo/feedback/`** — public, exists only where the flag is on; validates with zod, stores one `HiapDemoFeedback` row per remark plus one for the answers, uploads screenshots (PNG/JPG/WebP, ≤8 MB, ≤6 per submission) to the existing upload bucket through `S3FileStorageProvider`, and answers with an explicit 502 if the upload fails.
- **`/feedback/responses`** and `GET` on the same route — admin session only, newest first, screenshots inline.
- Migration `20260923140000-hiap-demo-feedback` creates the table.

## Checks

`tsc` clean, lint clean on the new files, migration applied locally. On the local production build: text-only submission stored (one `answers` row, one `comment` row), empty submission rejected with 400, unauthenticated `GET` refused with 403. The screenshot path reaches S3 and fails locally only because the local AWS key is a placeholder; dev has the real credentials the branding upload already uses.


carlosgraffi:
Closing: feedback collection will live outside CityCatalyst to keep the app lean. Branch kept in case we revisit.

PR #3202, opened 2026-09-23
CC-945: Show a drafting overview as Clima's first chat message

piotrnowakowski:
## Summary

When a concept note finishes drafting, Clima now posts an overview as the first chat message: what it processed, how each input was used, what is still missing, and suggested next requests. The prompt that triggers it is hidden from the user. This PR also fills in the city profile in the context bundle, which was declared but never populated.

## Changes

- **Hidden overview turn.** When the draft finishes (or fails partway), the chat panel sends `options.concept_note_turn = "draft_overview"`. Climate Advisor replaces the client text with a server-owned trigger, saves no user message, and claims the turn once per drafting build. A second request gets `409 concept_note_draft_overview_unavailable`, which the UI ignores. A failed turn releases the claim so it can be retried.
- **Grounding.** The turn gets the usual context bundle plus a summary of each chapter (title, whether drafted, start of text, open questions), with IDs removed.
- **New prompt.** `prompts/cnb/draft_overview.md` follows the role/task/input/tools/output structure. It is added to the CNB chat instructions for this turn only.
- **Draft state.** The draft response now includes `overview_pending`.
- **Label.** The chat shows "Summarising the new draft" during the overview turn only, in all five locales. Other turns keep their existing status line and reasoning traces.
- **City profile.** `cc_context.city` is now filled from the CityCatalyst city page: name, LOCODE, country, region, area and the latest population. The boundary geometry is left out. A failed lookup warns and keeps the last stored profile.

## How to test

1. Create a new concept note with a document, choose a funder with a template, and click **Start drafting**.
2. When drafting completes, Clima shows "Thinking… Summarising the new draft" and then streams the overview. No user message appears.
3. Reload the page. The overview is still the first message and no new chat request is sent.
4. On the **Context** tab, the city is shown as included in the run.

Automated tests: `pytest tests/cnb/test_draft_overview.py tests/cnb/test_city_profile.py tests/cnb/test_context_bundle_service.py` and `jest tests/concept-note-chat-progress.jest.tsx`.

## Ticket

[CC-945](https://linear.app/openearth/issue/CC-945/cnb-show-an-overview-of-what-changed-after-concept-note-generation)

## Notes

- The demo below was recorded against the local stack with real OpenRouter calls and a new note. Nothing was mocked.
- "How I used it" is inferred from chapter text: chapters do not yet record which source they used. Clickable source links need per-chapter source tracking, which is a follow-up.
- `cc_context.project` and `cc_context.ccra` are still never populated.

https://github.com/user-attachments/assets/5549039d-3acf-4b29-8010-2faad85c5cae


climate-advisor/service/app/routes/messages.py
```diff
@@ -104,6 +109,17 @@ async def post_message(
             context=payload.context,
             options=payload.options,
         )
+
+        # The hidden drafting-overview turn uses server-owned trigger text and
+        # is claimed once per drafting build before any streaming starts.
+        draft_overview_claim = None
+        if is_draft_overview_turn(payload.options):
+            draft_overview_claim = await claim_draft_overview(
```
DavidGBbr: One edge case to consider: the overview claim is committed before bearer validation. If Core rejects the token, the request returns before `StreamingHandler` is created, so its cleanup cannot release the claim. A retry then receives 409 and the overview is skipped for this build.

```py
draft_overview_claim = await claim_draft_overview(...)
...
if exc.status_code in _CC_AUTH_REJECTION_STATUSES and token_from_payload:
    raise HTTPException(status_code=401, ...)
```

Could you consider validating the bearer before claiming the overview, or releasing the claim on these early failures? I’m approving the PR and leaving this as a follow-up observation.

DavidGBbr (approved):

climate-advisor/service/app/routes/messages.py
piotrnowakowski: thanks a lot man implemented both of those 🙏

- **Claim after auth:** the overview claim now runs after bearer validation and message persistence, right before the `StreamingHandler` is built. A rejected token returns 401 without touching the claim.
- **Release on early failure:** if handler setup throws after the claim, the route calls `release_draft_overview` before re-raising, so the overview stays retryable.

Covered by two new tests in `tests/cnb/test_draft_overview.py`: a rejected bearer leaves the overview pending and a valid retry claims it, and a handler setup failure releases the claim. Fixed in 617c19b22.

PR #3203, opened 2026-09-23
CC-946: Allow changing funder while documents are processing

piotrnowakowski:
## Summary
- While uploaded documents were processing, the Funder dialog disabled **Save** and Climate Advisor rejected the change with 409. This PR allows the change during document processing.
- Funding selection only writes `funder_context`/`similar_projects`, and `complete_build` keeps unrelated bundle sections, so the two don't conflict. The bundle row is now locked (`FOR UPDATE`) during the funding write so a build that finishes at the same time can't overwrite it.
- Draft generation and in-flight edits still block funding changes.

## Changes
- `ConceptNoteWorkspace/index.tsx`: the funding dialog's `busy` flag no longer includes `contextStatus.busy`.
- `funding_selection.py`: removed the `context_bundle == building` guard and added the row lock.
- Tests: the guard test now covers only draft generation, and a new test confirms funding saves while the build is running.

## Testing
- `pytest tests/cnb/test_funding_selection.py`: 14 passed

Linear: [CC-946](https://linear.app/openearth/issue/CC-946/cnb-allow-changing-funder-while-documents-are-processing)


piotrnowakowski:
## Demo: assigning a funder while a document is processing

https://github.com/user-attachments/assets/6152ce53-ea4f-4fda-8b63-459dd39bedb9

Recorded against the real local stack from this branch (Next.js + Climate Advisor + Postgres). Nothing was mocked.

| t | Step | Result |
|---|------|--------|
| 0–6s | Open the Kraków concept note (no funder) and upload a PDF | Processing starts; `POST …/uploads/` → 202 |
| 6–13s | **Browse funders** → European Investment Bank → EIB Private Sector Loans | No "wait for the current operation" warning; **Save selection** enabled |
| ~14s | Save while the file still shows *Converting* | `PATCH …/application-context/` → **200** |
| 17–20s | Dialog closes | Funder profile: **Connected · European Investment Bank** |
| 20–42s | Processing finishes | File **Ready**; funder still selected |

**DB check after the run:** funder set on the run, `context_bundle.status = ready`, `funder_context` = EIB, `selected_sources` = 1. The funding write and the build completion were both kept.

Not covered by the video: the draft-running block and the exact-moment write race. The pytest and the bundle row lock cover those.

DavidGBbr (approved):

PR #3205, opened 2026-09-23
fix(CC-819): filter out null-email entries in findUsersInProject

tatimoreira:
## Summary

Fixes a crash on Account Settings > Teams > Collaborators pagination in development: some `CityInvite` rows have `email = NULL` (legacy invites created before the `email` column existed), and `UserService.findUsersInProject` unsafely cast these to `string`, producing a `null` that the frontend later crashed on (`.email.split(...)`).

## Changes

- `findUsersInProject` now filters out entries with a falsy `email` from all four sources (`OrganizationInvite`, `OrganizationAdmin`→`User`, `CityUser`→`User`, `CityInvite`) instead of casting `email as string`, so a null-email row is silently excluded rather than reaching the API response.

## How to test

1. In development, open Account Settings > Teams for a project with 50+ collaborators/invites.
2. Page through the Collaborators table past page 1 — no crash, and no phantom blank-email row.
3. `npm run lint` / `npx tsc --noEmit` pass on `app/src/backend/UserService.ts`.

## Ticket

CC-819

## Notes

Investigated directly against the dev DB: found 50 `CityInvite` rows with `NULL` email, all created before 2025-01-15 (when the `email` column was added via a bare `ALTER TABLE` with no backfill). They collapsed into a single crashing entry due to `uniqBy(users, "email")` deduping on the shared `undefined` key. Confirmed with Milan that this code-level fix is sufficient — no dev DB cleanup needed, since null-email rows can no longer surface or crash the endpoint regardless of what's in the database.


lemilonkh (approved):
Awesome, thanks 👍🏻✨

PR #3207, opened 2026-09-23
Fix CNB dashboard showing upload retry while sources upload

piotrnowakowski:
## Summary

While a new concept note uploads its sources, the dashboard card behind the create dialog showed "Upload incomplete" and "Try upload again", even though nothing had failed. The card now shows "Uploading sources…" during the upload. The retry state only appears when the upload actually fails.

## Changes

- `new-concept-note-dialog.tsx`: once the run is created, the dialog reports it via `onUploadingRunChange`. It clears this on failure and keeps it set on success, so the card can't flash the retry state while the page navigates to the workspace.
- `index.tsx`: the dashboard tracks the uploading run for both the create and retry dialogs and passes `uploading` to its card.
- `run-card.tsx`: while `uploading` is on, the card shows the info-tone "Uploading sources…" badge and its normal actions instead of the retry state. It reuses the existing i18n key.

## How to test

1. Open `/en/cities/<cityId>/concept-notes/`, click **+ New concept note**, attach a PDF of a few MiB and click **Create & open workspace**. Throttling upload in DevTools makes the window easier to see.
2. While the button says "Uploading sources…", the new card behind the dialog shows the blue "Uploading sources…" badge. It never shows "Try upload again".
3. Force an upload failure (e.g. stop Climate Advisor mid-upload). The card falls back to "Upload incomplete" / "Try upload again" as before.

## Notes

Demo below: real stack (Next + Climate Advisor from this branch), upload throttled to 90 KiB/s and no mocks. During the ~50 s upload the card was sampled 141 times: "Uploading sources…" 141×, "Try upload again" 0×. The failure path is unchanged and covered by `e2e/concept-note-upload-recovery.spec.ts`. A page reload mid-upload still shows the retry state, because the client can't tell the upload is still in flight.

https://github.com/user-attachments/assets/fb3d4c65-b8d3-4929-92c1-315a9e179afd


lemilonkh (approved):
Great, thanks 👍

